PR #117, opened 2014-01-08
enable debugging in rstudio when testing a file

jefferis:
- not sure if the try is necessary 
  (I was thinking about older Rstudio versions - don't know how to test version)
- this works for me for breakpoints in functions
  but not for top level breakpoints in tests themselves

### DIFF
--- a/R/test-files.r
+++ b/R/test-files.r
@@ -95,6 +95,9 @@ test_file <- function(path, reporter = "summary",
   old <- setwd(dirname(path))
   on.exit(setwd(old))
   
+  # set debug mode if we are using RStudio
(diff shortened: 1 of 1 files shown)


kenahoo:
`rstudioapi` is on CRAN now, could this move forward?

hadley:
Sure, I'll take a look when I'm next working on testthat

kenahoo:
Excellent - I had assumed this was an issue in RStudio, not in testthat, so this is good news. =)

jucor:
Friendly ping on this, please :)

kenahoo:
Is @jefferis 's approach here the best way forward?  If so, could we get a ruling and/or a merge?  Would be GREAT to not have to edit my tests just because I want to debug.

PR #184, opened 2014-09-26
Fix colour test cases, they switched on colors, always

gaborcsardi:
Well, almost always, when these options were not set before running the tests. It was my ignorance of course, I did not read the `options()` docs carefully enough. :(

### DIFF
--- a/tests/testthat/test-colour.r
+++ b/tests/testthat/test-colour.r
@@ -3,7 +3,7 @@ context("Colours")
 
 test_that("We have colours if we want to", {
 
-  op <- options()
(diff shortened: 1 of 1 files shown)


hadley:
See my fix for an even better way of doing this

gaborcsardi:
Don't you need `on.exit()`? The point of separating the query and the setting was that I wanted to put `on.exit()` between them.

hadley:
No, because I reset with `options(op)`

gaborcsardi:
Yes, sorry, did not see that in the patch, as it didn't change. Btw. if there is an error, then it won't reset, but this is nitpicking....

PR #186, opened 2014-09-27
functions is_not_a() and expect_is_not() added

Mullefa:
Found theses functions to be useful and a natural addition to `is_a()` and `expect_is()` respectively.


hadley:
Is there a reason you can just use `not()`?  Asserting that something isn't an object of a class doesn't seem that useful to me (as a general operation)

Mullefa:
Fair enough

PR #188, opened 2014-09-30
Allow Perl-style regex filters for tests

leeper:
Perl-style regular expressions are currently not allowed in the `filter` argument to `test_dir`, `test_package`, and `test_check`. This pull request adds a `perl` argument (logical, default `FALSE`) that enables Perl-style regular expressions, which is passed down to `grepl`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # testthat 0.9.1.9000
 
+* `test_dir()`, `test_package()`, and `test_check()` have an added `...` argument
+  that allows filtering of test files using, e.g., Perl-style regular expressions,
+  or `fixed` character filtering. Arguments in `...` are passed to `grepl()` (@leeper).
+
 * `test_check()` uses a new reporter specifically designed for `R CMD check`.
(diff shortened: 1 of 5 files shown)


hadley:
Can you please rebase?

leeper:
Done.

hadley:
The diff doesn't look right to me - can you please double check?

leeper:
Oh, yea, that was weird. How does this look?

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # testthat 0.9.1.9000
 
+* `test_dir`, `test_package`, and `test_check` have an added `...` argument
```
hadley: Doesn't look like you actually modified `test_check()` (which is right) but can you remove from here?

Also house style is to follow function names with `()`, e.g.` test_dir()`.

leeper:
Meant to modify `test_check()`. This has been fixed, so NEWS is now correct.

Added parentheses to function names in NEWS.

PR #207, opened 2015-01-05
fix evaluation environments for mock definition

### LINKED ISSUE #193: Cannot access variables declared outside the with_mock function
R version: 3.1.0
testthat: 0.9.1

Small example of the problem

```
test_that( "test",
          {
            value <- 1.5
            with_mock(
              expect_equal(2 * 3, 4),
              all.equal = function(x, y, ...) {value}
            )
          }
)
```

If I am executing the above code snippet through the “Test package” button in RStudio, then I get the error:

```
“1. Error: test -----------------------------------------------------------------
object 'value' not found”
```

Currently it seems it is not possible to reference from the mocked function variables which are declared outside the scope of the with_mock function

krlmlr:
With test. Fixes #193.


hadley:
Could you please rebase/merge?

krlmlr:
Done.

PR #208, opened 2015-01-07
reset failed flag in Reporter base class

### LINKED ISSUE #206: failed <<- TRUE should be done by the Reporter base class
### Reason

The member is declared and initialized in the base class, and hence should be modified (only) there.

### Implementation

All other reporters will need to `callSuper()` in `add_result()`, and probably also in `start_reporter()`.

### Use case

Different test files use different reporters, like this:

```
with_reporter(MultiReporter$new(reporters = list(get_reporter(), DetailedReporter$new())), {
...
})
```

The "main" file (e.g., `tests/testthat.R`) uses the "basic" reporter only to collect global success status:

```
test_check("tikzDevice", reporter = Reporter$new())
```

Currently I need to implement a subclass to work around this.

krlmlr:
Fixes #206.

I'm getting different kinds of random errors for the "watcher works correctly" test, with and without this pull request. Ubuntu 14.10, R 3.1.2.

### DIFF
--- a/R/reporter-minimal.r
+++ b/R/reporter-minimal.r
@@ -16,12 +16,12 @@ NULL
 MinimalReporter <- setRefClass("MinimalReporter", contains = "Reporter",
   methods = list(
     add_result = function(result) {
+      callSuper(result)
       if (result$skipped) {
         cat(colourise("S", "skipped"))
       } else if (result$passed) {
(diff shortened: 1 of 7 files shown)


hadley:
FWIW I hope to move to R6 one day

PR #210, opened 2015-01-07
Parallel ready (updated)

kforner:
Just merged the latest testthat master, and get rid of my branch

As discussed in my previous pull requests, I removed the parallel code and made the code parallel-ready, i.e. it is now easy to implement the parallel testing
in another package.

I merged your master, fixed some issues (R CMD check was not working at all), fixed some tests, integrated the new "skip" feature into the "testthat_results", removed the parallel code
and exported the functions I needed.

I also contacted Simon Urbanek to see if he's interested by the little hacked I made to add progress support to clusterApplyLB.


DESCRIPTION
```diff
@@ -6,6 +6,8 @@ Description: A testing package specifically tailored for R
 Authors@R: as.person(c(
     "Hadley Wickham <hadley@rstudio.com> [aut, cre]",
     "RStudio [cph]"))
+Author: Hadley Wickham <hadley@rstudio.com>
+Maintainer: Hadley Wickham <hadley@rstudio.com>
```
hadley: Can you please remove?

NEWS.md
```diff
@@ -1,8 +1,12 @@
 # testthat 0.9.1.9000
 
+* test (test_dir, test_file, test_package, test_check) functions now return a
```
hadley: I think `test_check()` has a different interface, since it's only ever used in R CMD check.

R/colour-text.r
```diff
@@ -1,8 +1,9 @@
-
+# fix the check NOTE:Namespace in Imports field not imported from: ‘crayon’
```
hadley: It's not really necessary (I'd argue that the warning is spurious because it doesn't capture build time dependencies) but getting rid of a NOTE is usually worth it.

R/test-files.r
```diff
@@ -23,34 +23,78 @@ test_env <- function() {
...
+    reporter = current_reporter, start_end_reporter = FALSE)
+  current_reporter$end_reporter()
 
+  results <- unlist(results, recursive = FALSE)
+
+  invisible(structure(results, class = 'testthat_results'))
+}
+
+
+#' source the helper scripts if any
```
hadley: Do these functions need to be exported?

R/test-files.r
```diff
@@ -23,34 +23,78 @@ test_env <- function() {
...
+#' @param path path to tests
+#' @param filter cf \code{\link{test_dir}}
+#' @return the test file paths
+#' @export
+find_test_scripts <- function(path, filter = NULL) {
   files <- dir(path, "^test.*\\.[rR]$")
   if (!is.null(filter)) {
     test_names <- basename(files)
-    test_names <- gsub("test-?", "", test_names)
+    test_names <- gsub("^test-?", "", test_names)
```
hadley: Thanks for catching this!

hadley:
Overall, it looks good. I've added a few minor comments, and then will do a thorough proof read and style check once you've processed them.

NEWS.md
kforner: Yes, but it still returns invisibly the test results. I guess there's no harm in that ?

R/colour-text.r
kforner: I do not understand why it is spurious: it complains about nothing being imported from a package listed in Imports.

hadley: Because things _are_ imported, with `::`

R/test-files.r
kforner: Yes I need all of them

R/test-files.r
kforner: You're welcome. I probably was bitten at one point

R/test-files.r
hadley: Ok, maybe flag as `@keywords internal` so the don't show up in the main doc index?

hadley:
I think long-term we should probably move to a system where the reporters do no data collection, and instead just respond to events. That would make this code a bit cleaner, and would simplify the reporters.

R/colour-text.r
kforner: I would say they are accessed, not imported. i.e the symbols are not copied in the imports env, aren't they ?

R/test-files.r
kforner: ok. I kept test_file and test_dir public which can be useful for everybody.

R/colour-text.r
hadley: Yeah, but that's not what the check is for - if that same call was inside a function, `R CMD check` wouldn't complain about it.

kforner: I just checked: the NOTE is absolutely not about the code. It is due to the fact that crayon is listed in the Imports, but nothing was imported. To fix the note, I located in the code where crayon was used, imported the appropriate symbols, and used them directly just to make the code look nicer.

hadley: There should be no need to manually add `green`, `yellow` and `red` to the namespace - and in fact you could argue that crayon doesn't even need to be in the `Imports` field at all because it's not a run-time dependency (you only need it at build time).

gaborcsardi: Hadley is of course right that `testthat` does not import `crayon`; more precisely, it imports it at build time. But I am actually not sure that the current solution is optimal, considering how R packages work. 

What is a bit unintuitive is that essentially all of `crayon` will be imbedded into the installed `testthat` this way. This can cause surprises. E.g. if there is a bug in `crayon`, and a new version is released to fix it, then you cannot simply fix the bug by installing a new version of `crayon`, since `testthat` does not import `crayon`. You need to rebuild `testthat`.

This of course all comes down to R's lack of build time dependencies. More precisely, this is not even a build time dependency, but it is an install time dependency. A build time dependency is what you need to build a source R package, e.g. roxygen2 to create the manual pages. `crayon` is needed when you install the source package.

Considering this, I think it is simpler to just import `green`, `yellow` and `red`. Or, we can still use `::`, but create `testthat_colours` at load time.

kforner: Sorry for being slow, I had never thought about a build time stuff like that. Of course the code is executed during the install. 
Should we keep we keep the imported color functions as Gabor suggested ?

hadley: Yes, definitely. Your fix is correct, I was just being nit picky about the R CMD check complaint.

NEWS.md
hadley: Sure. Can you please also use the standard style, e.g. `` `test_dir()` ``?

NEWS.md
```diff
@@ -1,8 +1,12 @@
...
+  **testthat_results** object that contains all results, and can be printed or
```
hadley: I'd use `` ` `` here, not `**`

R/test-files.r
```diff
@@ -23,34 +23,80 @@ test_env <- function() {
...
+    reporter = current_reporter, start_end_reporter = FALSE)
+  current_reporter$end_reporter()
+
+  results <- unlist(results, recursive = FALSE)
+
+  invisible(structure(results, class = 'testthat_results'))
+}
+
 
+#' source the helper scripts if any
```
hadley: Title should start with uppercase letter and end with .

R/test-files.r
```diff
@@ -23,34 +23,80 @@ test_env <- function() {
...
+    test_names <- basename(files)
+    test_names <- gsub("^test-?", "", test_names)
+    test_names <- gsub("\\.[rR]", "", test_names)
+
+    files <- files[grepl(filter, test_names)]
   }
-  with_reporter(reporter, lapply(files, .custom_test_file))
 
-  invisible(lister$get_summary())
+  file.path(path, files)
```
hadley: I think this could be replaced by `dir(, full.names = TRUE)` above

R/test-files.r
```diff
@@ -77,26 +123,42 @@ source_dir <- function(path, pattern = "\\.[rR]$", env = test_env(),
 #' @param path path to file
 #' @param reporter reporter to use
 #' @param env environment in which to execute the tests
-#' @return a data frame of the summary of test results
+#' @return the results as a "testthat_results" (list)
 #' @export
 test_file <- function(path, reporter = "summary", env = test_env()) {
-  current_reporter <- find_reporter(reporter)
+  reporter <- if (is.null(reporter)) NULL else find_reporter(reporter)
```
hadley: Why not make that the behaviour for `find_reporter()`?

R/test-files.r
```diff
@@ -77,26 +123,42 @@ source_dir <- function(path, pattern = "\\.[rR]$", env = test_env(),
...
+  invisible(.test_file(path, env, reporter, TRUE))
+}
+
+.test_file <- function(path, parent_env, reporter, start_end_reporter) {
+  if (is.null(parent_env)) parent_env <- globalenv()
```
hadley: Should be default argument?

R/test-files.r
```diff
@@ -77,26 +123,42 @@ source_dir <- function(path, pattern = "\\.[rR]$", env = test_env(),
...
   lister <- ListReporter$new()
-  reporter <- MultiReporter$new(reporters = list(current_reporter, lister))
-  lister$start_file(basename(path))
-  with_reporter(reporter, .test_file(path, env))
+  reporter <- if (!is.null(reporter))
```
hadley: I really don't like this style of assignment. Can you please repeat `reporter <- ` inside the branches, and multiline branches always need `{`

R/test-files.r
```diff
@@ -77,26 +123,42 @@ source_dir <- function(path, pattern = "\\.[rR]$", env = test_env(),
...
+    MultiReporter$new(reporters = list(reporter, lister))
+  else
+    lister
 
-  invisible(lister$get_summary())
-}
+  old_reporter <- set_reporter(reporter)
+  old_dir <- setwd(dirname(path))
+  on.exit({
```
hadley: Double indented, and `on.exit()` should always use `add = TRUE` for safety

R/test-files.r
```diff
@@ -77,26 +123,42 @@ source_dir <- function(path, pattern = "\\.[rR]$", env = test_env(),
...
-  sys.source2(basename(path), new.env(parent = parent_env))
+  fname <- basename(path)
+  lister$start_file(fname)
+
+  sys.source2(fname, new.env(parent = parent_env))
   end_context()
+
+  if (start_end_reporter) reporter$end_reporter()
+
+  invisible(structure(lister$results, class = 'testthat_results'))
```
hadley: It would be slightly better to have 

```R 
testthat_results  <- function(results) { 
  invisible(results, class = "testthat_results")
}
```

since you can then document that class specification in that function.

R/test-package.r
```diff
@@ -49,19 +49,27 @@ test_package <- function(package, filter = NULL, reporter = "summary") {
...
   if (file.exists(test_path2)) test_path <- test_path2
 
-  reporter <- find_reporter(reporter)
+  res <- .run_tests(package, test_path, filter, reporter)
+
+  invisible(res)
+}
 
+
+.run_tests <- function(package, test_path, filter, reporter)
```
hadley: I prefer not to use `.` for internal functions

R/test-package.r
```diff
@@ -49,19 +49,27 @@ test_package <- function(package, filter = NULL, reporter = "summary") {
...
+  reporter <- find_reporter(reporter)
   env <- test_pkg_env(package)
-  with_top_env(env, {
-    df <- test_dir(test_path, reporter = reporter, env = env, filter = filter)
+  res <- with_top_env(env, {
+    test_dir(test_path, reporter = reporter, env = env, filter = filter)
   })
 
-  if (reporter$failed) {
+  if (!all_passed.testthat_results(res)) {
```
hadley: Can you please either make `all_passed` a proper S3 generic, or just rename `all_passed.testthat_results` to `all_passed`

R/test-package.r
```diff
@@ -49,19 +49,27 @@ test_package <- function(package, filter = NULL, reporter = "summary") {
...
     stop("Test failures", call. = FALSE)
   }
-  invisible(df)
+  res
```
hadley: This should be `invisible(res)`

R/test-package.r
```diff
@@ -72,16 +80,9 @@ test_check <- function(package, filter = NULL, reporter = "summary") {
     stop("No tests found for ", package, call. = FALSE)
   }
 
-  reporter <- find_reporter(reporter)
-  env <- test_pkg_env(package)
-  with_top_env(env, {
-    df <- test_dir(test_path, reporter = reporter, env = env, filter = filter)
-  })
+  res <- .run_tests(package, test_path, filter, reporter)
```
hadley: Then this does need an intermediate assignment (and the same in `test_package()`)

R/test-results.r
```diff
@@ -0,0 +1,65 @@
...
+}
+
+
+# return if all tests are successful w/o error
+all_passed.testthat_results <- function(res) {
+  df <- as.data.frame.testthat_results(res)
+  sum(df$failed) == 0 && all(!df$error)
+}
+
+#' convert testthat_results to data frame
```
hadley: This doesn't need to be documented, just exported

R/test-results.r
```diff
@@ -0,0 +1,65 @@
...
+as.data.frame.testthat_results <- function(x, ...) {
+  summarize_results(x)
+}
+
+
+#' print testthat_results as a data frame
+#' @param x   the testthat results
+#' @param ... ignored
+#' @export
+print.testthat_results <- function(x, ...) {
```
hadley: Same here

hadley:
Final minor nitpicks. Once fixed, can you please rebase/merge, and resubmit? (Just so we get fresh conversation state)

R/test-files.r
kforner: because it's your function, and I did not want to potentially break existing code.

R/test-files.r
kforner: for .test_file() ? When I have no clear idea about the defaults, I prefer leave them explicit, especially for a internal function...

R/test-files.r
kforner: ? I thought single assignment was good practice ? Define once and only once...

R/test-files.r
hadley: The existing `find_reporter()` assumes that the input is a character, so changing it shouldn't affect any existing code.

R/test-files.r
hadley: It's just a bit weird to use this form of setting a default value, and makes it harder for me to figure out after the fact how it's supposed to be used.

R/test-files.r
hadley: I think it makes code harder to reader because the value and the name end up widely separated

R/test-package.r
kforner: why, since it is not a user function ?

R/test-results.r
kforner: ok I got it

PR #211, opened 2015-01-07
export colourise

robertzk:
Per @krlmlr 's [comment](https://github.com/hadley/testthat/commit/7261c33ebe21024831d47f5c5d7afca578932135#commitcomment-9189667)


R/colour-text.r
```diff
@@ -5,6 +5,7 @@ testthat_colours <- list(
   error = crayon::red
 )
 
+#' @export
 colourise <- function(text, as = c("passed", "skipped", "error")) {
```
krlmlr: I think the parameters `fg` and `bg` should remain and trigger a warning when used.

robertzk: How's this?

On Wed, Jan 7, 2015 at 4:26 PM, Kirill Müller <notifications@github.com>
wrote:

> In R/colour-text.r
> <https://github.com/hadley/testthat/pull/211#discussion-diff-22621607>:
>
> > @@ -5,6 +5,7 @@ testthat_colours <- list(
> >    error = crayon::red
> >  )
> >
> > +#' @export
> >  colourise <- function(text, as = c("passed", "skipped", "error")) {
>
> I think the parameters fg and bg should remain and trigger a warning when
> used.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hadley/testthat/pull/211/files#r22621607>.
>



-- 
-Robert Krzyzanowski

R/colour-text.r
```diff
@@ -5,9 +5,13 @@ testthat_colours <- list(
   error = crayon::red
 )
 
-colourise <- function(text, as = c("passed", "skipped", "error")) {
+#' @export
+colourise <- function(text, as = c("passed", "skipped", "error"), fg, bg) {
```
krlmlr: This doesn't work for a call like `colourise("text", "green")`. OTOH, we really want to have `as` in the second position. @hadley: Any idea?

robertzk: Just compare values. One second.

On Wed, Jan 7, 2015 at 4:38 PM, Kirill Müller <notifications@github.com>
wrote:

> In R/colour-text.r
> <https://github.com/hadley/testthat/pull/211#discussion-diff-22622471>:
>
> > @@ -5,9 +5,13 @@ testthat_colours <- list(
> >    error = crayon::red
> >  )
> >
> > -colourise <- function(text, as = c("passed", "skipped", "error")) {
> > +#' @export
> > +colourise <- function(text, as = c("passed", "skipped", "error"), fg, bg) {
>
> This doesn't work for a call like colourise("text", "green"). OTOH, we
> really want to have as in the second position. @hadley
> <https://github.com/hadley>: Any idea?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hadley/testthat/pull/211/files#r22622471>.
>



-- 
-Robert Krzyzanowski

robertzk: This will do it.

On Wed, Jan 7, 2015 at 4:41 PM, Robert Krzyzanowski <technoguyrob@gmail.com>
wrote:

> Just compare values. One second.
>
> On Wed, Jan 7, 2015 at 4:38 PM, Kirill Müller <notifications@github.com>
> wrote:
>
>> In R/colour-text.r
>> <https://github.com/hadley/testthat/pull/211#discussion-diff-22622471>:
>>
>> > @@ -5,9 +5,13 @@ testthat_colours <- list(
>> >    error = crayon::red
>> >  )
>> >
>> > -colourise <- function(text, as = c("passed", "skipped", "error")) {
>> > +#' @export
>> > +colourise <- function(text, as = c("passed", "skipped", "error"), fg, bg) {
>>
>> This doesn't work for a call like colourise("text", "green"). OTOH, we
>> really want to have as in the second position. @hadley
>> <https://github.com/hadley>: Any idea?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/hadley/testthat/pull/211/files#r22622471>.
>>
>
>
>
> --
> -Robert Krzyzanowski
>



-- 
-Robert Krzyzanowski

robertzk: Actually, this breaks on 

```r
colourise(bg = 'blue', fg = 'red', text = 'blah')
```

since any call to `colourise` that uses a named reference to `text` not in the first argument will have to

R/colour-text.r
```diff
@@ -5,9 +8,19 @@ testthat_colours <- list(
...
+  fg_warn <- function() warning("The fg parameter to testthat::colourise is deprecated.", call. = FALSE)
```
krlmlr: How about one function `param_warn(name)`?

R/colour-text.r
```diff
@@ -5,9 +8,19 @@ testthat_colours <- list(
...
+  if (!missing(fg)) fg_warn()
+  if (any(is.element(as, allowed_colors))) {
```
krlmlr: Does capitalization matter here?

R/colour-text.r
```diff
@@ -5,9 +8,19 @@ testthat_colours <- list(
...
+    fg_warn()
+    return(eval.parent(sys.call()[1:2]))
+  }
+  if (!missing(bg)) warning("The bg parameter to testthat::colourise is deprecated.", call. = FALSE)
```
krlmlr: I think we should add a test for the compatibility behavior. We could mock `getOption` or just set the corresponding option to `FALSE` for the test.

R/colour-text.r
robertzk: The only case this still breaks on is

```r
colourise('red', 'blue', text = 'blah')
```

R/colour-text.r
robertzk: Yes, might as well!

R/colour-text.r
robertzk: Should work for all cases now.

R/colour-text.r
robertzk: It seems so, the old version gives colored results for capital versions of colors.

R/colour-text.r
robertzk: Ok, working on it.

R/colour-text.r
hadley: I think this is now too complicated - it would be better to have two functions. One function should be exported and use the old interface and wrap the equivalent crayon function with a deprecation message. The other function should be internal and be the same as the current `colourise()` but with a different name.

robertzk:
I'm going to write documentation and tests and re-open later.

PR #214, opened 2015-01-09
Parallel ready: 2nd update

kforner:
I implemented hadley's review suggestions, merged and requested a new pull.


NEWS.md
```diff
@@ -27,16 +32,20 @@
...
 
-* New `succeed()` exepctation always succeeds.
+* New `succeed()` expectation always succeeds.
 
 * `skip_on_travis()` allows you to skip tests when run on Travis CI.
   (Thanks to @mllg)
 
 * `colourise()` was removed. (Colour is still supported, via the `crayon` package.)
 
+<<<<<<< HEAD
```
hadley: Looks like a merge failure here

hadley:
Thanks!

krlmlr:
@kforner: If I understand correctly, this PR was to support parallel testing with external code. Do you have an example or a package that implements this? Is there overlap with #501? Thanks!

kforner:
@krlmlr: What do you mean by "with external code" ? 
The parallel testing that I use is basically: mclapply(test_files, test_file), and then to aggregate the individual test results. So the granularity is the test source file. There is no public example of such a package since testthat does not support that. I have a custom extension built on top of testthat, thanks to the modifications of this PR.

krlmlr:
I guess the custom extension you were referring to is what I meant by "external code". #501 looks very similar to this idea, I had missed your PR.

How do you aggregate the individual test results? Which reporter do you use?

kforner:
> How do you aggregate the individual test results? Which reporter do you use?

in parallel, I disable the reporter (NULL) and I get the results (from my PR: the results as a "testthat_results" (list)) for each test file. 

Then I convert the individual results to data frame, and aggregate the results by context and file. 
Finally I can report all the results by test, or only a summary by file.

byapparov:
@kforner It would be great if you could share some code that would enable us to run tests behind this in parallel:

```r
test_dir(
        path = paste0("my/test"),
        reporter = "junit"
)
```

right now, if I put `test_that` calls inside of `mclapply`, I cannot get correct Junit output.

PR #222, opened 2015-01-17
fixes #221

### LINKED ISSUE #221: Feature request: conditional adjustment of working directory for "stand-alone" testing
Might just be me, but I often put together my unit tests "piece by piece" in the sense that I actually execute explicit calls to `expect_*()` or `test_that()` in a "stand-alone" manner to test things out on an atomic level before, for example, using something like `test()`.

Whenever the actual test code relies on things contained in the `~/tests/testthat` directory, e.g. `tests/testthat/rtools-2.15` in your `devtools` package, running tests that way involves temporarily adjusting the working directory in a manual manner. IMO, this step could/should be automated.

I've put together a function that would handle this (see upcoming pull request)

rappster:
I've added `test_that_suggested()` to demonstrate how calls to `test_that()` could make use of `withConditionalWorkingDirectory()`. Of course, the same `with` approach could also be used on a more atomic level, i.e. for calls to `expect_*()` functions.

Let me know what you think.

### DIFF
--- a/R/withConditionalWorkingDirectory.r
+++ b/R/withConditionalWorkingDirectory.r
@@ -43,8 +43,6 @@ withConditionalWorkingDirectory <- function(code) {
   wd <- getwd()
   if (!length(grep("/tests/testthat$", wd))) {
     setwd("tests/testthat")
-  } else {
-    getwd()
   }
   on.exit(setwd(wd))
   force(code)
--- a/R/withConditionalWorkingDirectory.r
+++ b/R/withConditionalWorkingDirectory.r
@@ -1,0 +1,62 @@
+#' @title
+#' Evaluate test code in appropriate working directory.
+#'
+#' @description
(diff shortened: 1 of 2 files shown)


R/withConditionalWorkingDirectory.r
```diff
@@ -0,0 +1,60 @@
...
+#    expect_true(file.exists("test_dir"))
+#  )
+#' # Clean up:
+#' if (grepl(basename(getwd()), old)) {
+#'   unlink("tests", recursive = TRUE)
+#' }
+#' setwd(old)
+#' @export
+withConditionalWorkingDirectory <- function(code) {
+  wd <- getwd()
```
hadley: `file.exists()` would be a better test, and I think it should error if the directory doesn't exist

hadley:
Hmmmm, I understand the problem but this doesn't feel like quite the right fix to me. Lets discuss more in the issue.

PR #223, opened 2015-01-18
add 'Catch' for cpp unit testing

kevinushey:
This PR adds the [Catch](https://github.com/philsquared/Catch) unit testing library to the `include/tests` dir, and adds some wrappers to allow one to have a `testthat`-like interface for testing C++ code.

It's a very quick proof of concept (needs more work to properly respect all compilation arguments) but might be useful.

The basic idea is:

1. The user puts C++ test files in `tests/testthat/cpp/`, prefixed with `^test-` by default,
2. The test runner checks to see if there are any files in that folder; if so, it attempts to compile and run an executable that prints out the test results.

I haven't completed the wiring (because I'm not aware how the reporters work yet) but let me know if this is worth finishing up.

### DIFF
--- a/R/test-compiled-code.r
+++ b/R/test-compiled-code.r
@@ -1,3 +1,100 @@
+trim_whitespace <- function(string) {
+  vapply(string, FUN.VALUE = character(1), USE.NAMES = FALSE, function(x) {
+    gsub("^[\\s\\n]*|[\\s\\n]*$", "", x, perl = TRUE)
+  })
+}
+
+read_makevars <- function(pkg_path) {
+
+  # Look for 'makevars' files in a number of locations, and parse them.
+  is_windows <- Sys.info()[["sysname"]] == "Windows"
+
+  # This won't be perfect for e.g. scripts that declare
+  # variables in a configure script, but it's good for 99%
+  # of users. We list the files in descending order of
+  # priority.
+  makevars_paths <- c(
+
+    # Makevars.site in etc
+    file.path(R.home(), "etc", "Makevars.site"),
+
+    # ~/.R/Makevars (note: same on windows)
+    file.path(Sys.getenv("HOME"), ".R", "Makevars"),
+
+    # src/Makevars
+    if (is_windows)
+      file.path(pkg_path, "src", "Makevars")
+    else
+      file.path(pkg_path, "src", "Makevars.win")
+
+  )
+
+  flags <- lapply(makevars_paths, function(path) {
+    if (!file.exists(path)) return(character())
+
+    content <- paste(readLines(path), collapse = "\n")
+
+    # Replace '\\n' with plain old space
+    content <- gsub("\\\\\\s*\\n", " ", content, perl = TRUE)
+
+    # Now split back up again -- now, each unique definition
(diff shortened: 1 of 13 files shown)


R/test-compiled-code.r
```diff
@@ -0,0 +1,99 @@
...
+  )
+
+  if (file.exists(bin_path))
+    stop("Directory '", bin_path, "' already exists; collision with other running tests?")
+
+  dir.create(bin_path, recursive = TRUE)
+  on.exit(unlink(bin_path, recursive = TRUE), add = TRUE)
+
+  # TODO: choose appropriate compiler?
+  compiler <- if (Sys.info()[["sysname"]] == "Darwin")
```
hadley: Maybe get this from the appropriate env var, if set?

hadley:
I like the idea! Especially since that coveralls now supports coverage stats on C/C++ code

R/test-compiled-code.r
kevinushey: I went the full 9 yards and wrote a simple parser for the various `Makevars` that could be used to set this as well :)

R/test-compiled-code.r
```diff
@@ -0,0 +1,224 @@
...
+  splat <- strsplit(LinkingTo, "\\s*[,\\n]\\s*", perl = TRUE)[[1]]
+
+  # Remove any version information
+  splat <- gsub("\\s*\\(.*", "", splat)
+
+  splat
+
+}
+
+test_compiled_code <- function(package, test_path, filter, verbose = FALSE) {
```
jjallaire: I think you can replace 100% of the test_complied_code implementation by just calling R CMD SHLIB with the test_path as an argument, etc. (it will read Makevars, build the command line, etc.). You *might* still need to do the DESCRIPTION parsing but command line flags will all be gotten right via R CMD SHLIB. Note that this in the technique we use for libclang in RStudio to discover all of the command line flags to feed to CRAN.

kevinushey: Right! I had forgotten about that. I'll replace this with `R CMD SHLIB --dry-run`.

hadley:
I think it's worth finishing up if you're still interested

kevinushey:
Sure, let me try to clean it up.

kevinushey:
@hadley, I think this should be good to go. I just tested this out on a Windows VM and all works well there too.

I updated the PR to use `R CMD SHLIB --dry-run` as per @jjallaire's suggestion and I think everything is much cleaner now. Some other todos / thoughts:

- Should the test executable be placed in a package folder, e.g. in the `bin` folder?
- Do we need to do anything special to accommodate `src/Makevars` files in packages?
- Should we attempt to parse the test output from Catch and use the currently selected testthat reporter instead?

kevinushey:
Hmm, it looks like `R CMD SHLIB` basically asks if there is a `Makevars` or `Makevars.win` file in the same directory where the file is being compiled -- perhaps on compilation of test files we should just copy those to the directory where we're compiling the test executable.

kevinushey:
Arghh, I pulled / rebased from a stale branch. Will have to fix up from home later :/

PR #226, opened 2015-01-30
correct function name in equivalence documentation

paulstaab:
I'm a bit confused about the current documentation of the equivalence checks. My best guess is that the second `expect_equal` should be `expect_equivalent` or am I mistaken here?


hadley:
Yes, thanks!

PR #227, opened 2015-02-14
add usage advice for expect_equal_to_reference

jennybc:
If you __only__ run tests via R CMD check, I don't think tests using `expect_equal_to_reference()` are actually in force. The reference RDS is created anew à la Groundhog Day, in some temporary directory, then discarded.

Even in you agree w/ my point, who knows if you like the way I've worded the cautionary advice? But it felt more constructive to make a PR than just open an issue.

Since I had to re-roxygenize, I seem to have also caught an `Rd` update related to another recent commit (30c876630dcb765b89f0eb9062b58b7b96784d12).


hadley:
Looks good - thanks!

PR #235, opened 2015-03-04
The highest compliment of all

Ironholds:
Based on a twitter conversation with @jennybc - we don't want the compliments to get stale!

### DIFF
--- a/R/reporter-summary.r
+++ b/R/reporter-summary.r
@@ -101,6 +101,9 @@ SummaryReporter <- setRefClass("SummaryReporter", contains = "Reporter",
   ":)",
   "Woot!",
   "Way to go!",
+  "Hadley has nothing on you!",
+  "Your code is a wonderland",
+  "Good Job! You've made the RCats purr",
   "Nice code."
 )
 labels <- c(1:9, letters, LETTERS)


hadley:
Can you please also add a bullet to NEWS?   You might also want to wait a couple of days and integrate with https://gist.github.com/rmflight/01d76949952a54568506

gaborcsardi:
How about having a longish list of superb words, for endless fun and surprises? E.g. something like https://github.com/sindresorhus/superb/blob/master/words.json ?

hadley:
I like the idea of some madlib style compliments :)

gaborcsardi:
Yeah, something like 
`(|Wow! )`
`(You are|Your code is|Your package is|This code is|This code is just)`
`(|extra|very)`
`(glorious|superb|slick|...)`
`(|today|, again|...)`
`!`

Ironholds:
I like this plan! I'll revisit the original patch this evening and add the things mentioned above (Gabor, you want to handle the madlib stylings? Vectors and non-determinism)

gaborcsardi:
Actually, I wanted to have something else in the https://github.com/metacran/mason package:

```
Looks like this will be a slick package, I am building it right now!
```

It is crazy to create a `praise` package for this?

jennybc:
If there's a separate package, I propose it be named and designed so it could potentially dispense both praise and solace. Not entirely kidding.

gaborcsardi:
Well, that would be three packages: `praise`, `solace` and the third one with the common machinery. I **am** kidding. 

But maybe one package is not crazy. It does look somewhat unprofessional, though, to write and use these kind of packages. Should we care?

Ironholds:
I wrote a package thats' primary contribution to the R sphere is setting pi to precisely 3. I'm sort of okay with it ;p

It's not unprofessional to have feelings, or to be surrounded by people with "Dr." before their name and worry you're letting the group down. It's not unprofessional to be smart enough to know you're not the smartest person in the room, or to have code that doesn't run first time, or to get impostor syndrome. It shouldn't be unprofessional to write a package that pokes at this and hopefully makes people smile when they read the README and maybe, just maybe, helps one or two of them, even in a way that they know is ridiculous.

The thing I love about the R community is precisely that it is not about ranking people on what they have done to make tidy data, or what they have done to provide APIs, or what they have done to make URLs first-class citizens. It's about what they have done /for the community/. And sometimes that's Serious Business packages and sometimes it's making people smile.

gaborcsardi:
All right, here we go: https://github.com/gaborcsardi/praise

gaborcsardi:
So, praise is on CRAN: http://www.r-pkg.org/pkg/praise
Maybe we can update this PR to use it?

Ironholds:
Go nuts! Or submit your own and I'll close this one; this week looks...hectic. I may not get to write recreational code (such an imposition).

hadley:
@Ironholds @gaborcsardi I'm happy to add more praise with the praise package. Does either of you want to do a PR?

gaborcsardi:
Sure, I can, but I am still travelling with very minimal internet access. Will try to catch up next week.

Issue #244, opened 2015-04-09
warnings not converted to errors with warn=2 in tests


kforner:
create an empty package with this test file:
```
test_that('warning bug', expect_error({options(warn = 2); warning('toto')}) )
```
with the current HEAD version of testthat, you get:
```
1. Failure(@test-testhat_bug.R#1): warning bug ---------------------------------
{
    options(warn = 2)
    warning("toto")
} code did not generate an error
```

I tracked it down to commit 5ffe62f9441b3e03c801032908479945d4e651df , which makes sense.

I tried to fix it, but it does not seem trivial. 
But I wonder: is it really a good idea to silence the messages and warnings in the tests ? 
I mean, there are too cases: 
 * expected warnings, that the test author is aware of. In that case he can explicitly suppress them using suppressMessages and suppressWarnings around the expect_xxx() calls.

* non expected warnings. If you silence them, you could hide bugs in the tests code. 

What do you think ?

hadley:
Lets discuss in #254

HyukjinKwon:
Hi @hadley So, if I understood correctly, ignoring messages is the right thing to do. However, do you mind if I ask whether it is possible to turn the warnings into errors in any way please?

kforner:
I think I have the same kind of debugging problems. When there's a warning, it's not always easy to track down the exact source. Before I used to insert browser() call, set options(warn = 2) and options(error = recover) but that no longer works for me. Is there a proper way to do it ?
Thanks.

PR #245, opened 2015-04-09
fix issue #244 using suppressWarnings and suppressMessages to silence th...

kforner:
...e tests


kforner:
does not fix the bug.

PR #248, opened 2015-04-15
skip test if devtools is not installed

### LINKED ISSUE #247: Test that fails when devtools isn't installed
from @s-u

```
1. Error: can mock if package is not loaded ------------------------------------
there is no package called 'devtools'
1: withCallingHandlers(eval(code, new_test_environment), error = capture_calls)
2: eval(code, new_test_environment)
3: eval(expr, envir, enclos)
4: with_mock(`devtools::add_path` = identity, expect_identical(devtools::add_path, identity)) at test-mock.r:120
5: extract_mocks(new_values = new_values[!code_pos], .env = .env)
6: lapply(setNames(nm = mock_qual_names), function(qual_name) {
       pkg_name <- gsub(pkg_and_name_rx, "\\1", qual_name)
       name <- gsub(pkg_and_name_rx, "\\2", qual_name)
       if (pkg_name == "")
           pkg_name <- .env
       env <- asNamespace(pkg_name)
       if (!exists(name, envir = env, mode = "function"))
           stop("Function ", name, " not found in environment ", environmentName(env),
               ".")
       mock(name = name, env = env, new = eval(new_values[[qual_name]]))
   })
7: FUN(X[[i]], ...)
8: asNamespace(pkg_name)
9: getNamespace(ns)
10: tryCatch(loadNamespace(name), error = function(e) stop(e))
11: tryCatchList(expr, classes, parentenv, handlers)
12: tryCatchOne(expr, names, parentenv, handlers[[1L]])
13: value[[3L]](cond)

Error: Test failures
Execution halted
```

@krlmlr can you please take a look?

krlmlr:
Fixes #247.

### DIFF
--- a/tests/testthat/test-mock.r
+++ b/tests/testthat/test-mock.r
@@ -118,6 +118,7 @@ test_that("can mock hidden functions", {
 test_that("can mock if package is not loaded", {
   if ("package:devtools" %in% search())
     skip('devtools is loaded')
+  skip_if_not_installed("devtools")
   with_mock(`devtools::add_path` = identity, expect_identical(devtools::add_path, identity))
 })
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
 * Added `skip_on_os()`, to skip tests on specified operating systems
   (@kevinushey).
 
+* Skip test that depends on `devtools` if it is not installed (#247, @krlmlr)
+
 # testthat 0.10.0
 
 * Failure locations are now formated as R error locations.


krlmlr:
I think the coverage report is unrelated.

hadley:
How about using `skip_if_not_installed()`? (or better, use a recommended package that's more likely to be installed)

tests/testthat/test-mock.r
```diff
@@ -116,9 +116,11 @@ test_that("can mock hidden functions", {
 })
 
 test_that("can mock if package is not loaded", {
-  if ("package:devtools" %in% search())
-    skip('devtools is loaded')
-  with_mock(`devtools::add_path` = identity, expect_identical(devtools::add_path, identity))
+  if ("package:boot" %in% search())
+    skip('boot is loaded')
+  if (!length(find.package("boot", quiet=TRUE)))
```
hadley: Can you please use `requireNamespace("boot", quietly = TRUE)`?

But is it even necessary to test? I think only broken installs will not have boot. But won't you also have to add boot to suggests in order to avoid R CMD check warnings?

krlmlr: Would be much better to create a [disposable](https://cran.r-project.org/web/packages/disposables/index.html) package for this. Do you mind if I add this to the "Suggests"?

hadley: But then you'll have to get that on CRAN?  Why can't we go back to using devtools which is already in suggests?

krlmlr: I guess we can skip that test if `devtools` is not installed (the original case in #247). It will be tested with `R CMD check`, because it's in "Suggests".

The `disposables` package is already on CRAN, it can create one-off packages for testing.

PR #249, opened 2015-04-20
added expect_in and is_in

BayesyBrad:
expect_in and is_in are tests with expectations that an object is
within a vector of expectations, using match()

I use this kind of construct to check that an object, usually from some kind of user input or scraped text, is in a vector of acceptable values.


hadley:
Can you please start the conversation off by filing an issue that discusses why you want this?

PR #262, opened 2015-05-26
Handle skipped tests in TAP reporter 

nealrichardson:
Because `skip`ped tests did not have `result$passed`, they were marked as "not ok" in the TAP output. Following http://testanything.org/tap-specification.html, I added a skip directive to the output in case of `result$skipped`. 

I'd happily write a test for this change if you could indicate where that should be. I didn't find any tests for the reporter output, only for the existence of reporters. I did confirm that the output with the skip directives is correctly parsed, at least by the Jenkins TAP parser:

screen shot 2015-05-26 at 11 35 27 am


nealrichardson:
Ok, tests for the TAP reporter output added, including a test for the skip directive. This should increase test coverage, though the coveralls job in https://travis-ci.org/hadley/testthat/builds/64178838 timed out (which appears to be why the Travis build was marked as failing; the tests passed).

tests/testthat/test-reporter-tap.r
```diff
@@ -0,0 +1,14 @@
+context("TAP reporter")
+
+test_that("TAP reporter handles context and pass/fail/skip", {
+  filename <- tempfile(fileext="tap")
+  sink(filename)
+    test_dir("test_dir", reporter = "tap")
```
hadley: I think you could simplify this with `capture.output()`

tests/testthat/test_dir/test-skip.r
```diff
@@ -0,0 +1,6 @@
+context("Skip")
```
hadley: Could you please use a lower case skip to be consistent with the other files?

tests/testthat/test-reporter-tap.r
nealrichardson: Yes, thanks, much nicer.

tests/testthat/test_dir/test-skip.r
nealrichardson: Done, though I noticed that "Bare" and "Basic" in test_dir are also capitalized.

hadley:
Could you please also add a bullet point to NEWS?

hadley:
Thanks! (For future reference, I don't get a notification when you make changes, just when you make comments, so once you're done, it's helpful to add a comment like "PTAL")

nealrichardson:
Ah ok, thanks for pointing that out, will do.

nealrichardson:
Looks like the NEWS bullet I added was clobbered in a merge: https://github.com/nealrichardson/testthat/compare/master...hadley:master#diff-8312ad0561ef661716b48d09478362f3L10

I could make a new PR with that if you want, but maybe it's easier for you to just add it back.

PR #268, opened 2015-06-16
changed the testthat_results data frame column "skipped" from numeric…

kforner:
… to logical


hadley:
Why? Couldn't there be multiple skipped tests?

kforner:
unless mistaken, there's a row per test, and a test can only be skipped once.

kforner:
Why do you mean by "multiple skipped tests" ? I thought that a skip() call ended the current test execution, like a return().

kforner:
seems like the travis check failed because of the coverage. The R CMD check was successful...

hadley:
Thanks!

PR #269, opened 2015-06-16
Add argument to turn off suppression of warnings and messages

### LINKED ISSUE #244: warnings not converted to errors with warn=2 in tests

jimhester:
Closes #244

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat 0.10.0.9000
 
+* `test_that()` gains a `suppress` option to turn off automatic suppression of
+  warnings and errors. (@jimhester #269, @kforner #244)
+
 # testthat 0.10.0
 
 * Failure locations are now formated as R error locations.
--- a/R/test-that.r
+++ b/R/test-that.r
@@ -36,7 +36,7 @@
 #'   warning("a warning!")
 #'   message("a message!")
 #' }, suppress = NULL)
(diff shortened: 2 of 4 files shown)


jimhester:
Thinking about this a little more this isn't exactly the API I would prefer.  While it does allow you turn turn off suppression for each `test_that` block it does not allow you to turn it off globally.

I may change the default value `suppress = getOption("testthat.suppress", c("message", "warning"))` which would let you change them globally.

hadley:
I think a global option is the wrong way to do this.

PR #270, opened 2015-06-18
Fix extraneous NAs from compare.character's message.

kbroman:
- Consider this example:

        x <- y <- letters[1:5]
        y[3] <- "x"
        expect_equal(x,y)

- The output looked like:

        Error: x not equal to y
        1 string mismatches:
        x[3]: "x"
        y[3]: "c"

        x[NA]: NA
        y[NA]: NA

        x[NA]: NA
        y[NA]: NA

        x[NA]: NA
        y[NA]: NA

        x[NA]: NA
        y[NA]: NA

- This fix eliminates those extraneous lines with `NA`.

### DIFF
--- a/R/compare.r
+++ b/R/compare.r
@@ -78,8 +78,9 @@ compare.character <- function(x, y, ..., max_diffs = 5, max_lines = 5,
   }
 
   width <- width - 6 # allocate space for labels
-  n_show <- seq_len(min(length(diff), max_diffs))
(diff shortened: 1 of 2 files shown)


hadley:
Duplicate of #274 - that also fixes a few other minor problems.  Thanks for spotting and fixing this though! It's been bothering me for a while but I never had time to look into it

PR #278, opened 2015-07-09
TAP reporter should use SKIP directive for skipped tests

skonzem:
This change produces the following:
```
ok 194 # SKIP watcher works correctly
```
Compare to examples from the [spec](https://testanything.org/tap-specification.html):
```
ok 2 - # SKIP no /sys directory
ok 23 # skip Insufficient flogiston pressure.
1..0 # Skipped: WWW::Mechanize not installed
```


skonzem:
Actually #262 is the better PR.

PR #284, opened 2015-08-02
sqlite Reporter additon

osunderdog:
Added an sqllite reporter for tracking test over time.
Depends on RSQLite
Needed a way to keep track of test results from data sets over long periods of time.
This is used for testing datasets from manufacturing environment to find cases where the data sets to not conform to expected conditions. 
Adds a start and end datetime attribute to mark the timeline of a test result.
For example if a test runs every 5 minutes and it is validating data conforms to existing requirements, but for a period of time it does not, then this will give us a history to review and identify the issue in the dataset.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,6 +40,7 @@ Collate:
     'reporter-multi.r'
     'reporter-rstudio.R'
     'reporter-silent.r'
+    'reporter-sqlite.R'
     'reporter-stop.r'
     'reporter-summary.r'
     'reporter-tap.r'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -12,6 +12,7 @@ BugReports: https://github.com/hadley/testthat/issues
 Depends:
     R (>= 3.1.0),
     methods
+    RSQLite
 Imports:
     digest,
     crayon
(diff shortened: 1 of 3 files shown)


osunderdog:
This is my first time committing back to a package.  Hope i'm on the right track here.

osunderdog:
Fixing Depends with comma

hadley:
I don't think this is a bad idea, but I don't think testthat is the right place for it - it should go in its own package.

If you want to keep working on this, I think it would be a valuable contribution to the community. And with only a bit more effort you should be able to generalise to work with any database, using the functions in the DBI package to abstract over the backend.

osunderdog:
Yeah, I was wondering if the addition of a dependency on RSQLite would be an issue.  Yes I can probably abstract this to callback functions... for example 'setup' call back for first time action and 'end test' call back for action on each.  I'll give it a try and re-submit.

Thanks.

hadley:
Sorry, I think you misunderstood - I think a reporter that records results in a database is most suitable for its own package, not inclusion in testthat.

hadley:
And there should be no need for callback functions. You should just provide a DBI database connection, and then use standard DBI generics on that object.

osunderdog:
Got it.  I've used DBI in perl.  I'll give that a try.

osunderdog:
I have a first pass of a separate package for DBI.  https://github.com/osunderdog/DBIReporter-testthat.git <https://github.com/osunderdog/DBIReporter-testthat.git>

I was thinking of posting on google+ ’Statistics and R’ community, but I’m not sure I want that much visibility.

PR #299, opened 2015-09-25
Fix watcher issues

krlmlr:
- use explicit shell call in test
- use `tryCatch` for checking if digest succeeds
    - the previous approach is still subject to race conditions -- not likely, but possible

CC @eddelbuettel: Would you support an argument to `digest()` that makes it return NA instead of throwing an error?

### DIFF
--- a/tests/testthat/test-watcher.r
+++ b/tests/testthat/test-watcher.r
@@ -47,7 +47,7 @@ test_that("watcher works correctly", {
   dir.create(test_path)
 
   delayed.bash.cmd <- function(command) {
-    system(paste0("sleep 1;", command), wait=FALSE)
+    system(paste0("bash -c 'sleep 1;", command, "'"), wait=FALSE)
   }
 
   add.code.file <- function(file.name) {
--- a/tests/testthat/test-watcher.r
+++ b/tests/testthat/test-watcher.r
@@ -35,7 +35,7 @@ test_that("compare state works correctly", {
  })
 
 test_that("watcher works correctly", {
-  skip_on_cran() && skip_on_os("windows")
+  skip_on_cran()
 
   loc <- tempfile("watcher", tmpdir = "/tmp")
   dir.create(loc)
--- a/tests/testthat/test-watcher.r
+++ b/tests/testthat/test-watcher.r
@@ -36,6 +36,12 @@ test_that("compare state works correctly", {
 
 test_that("watcher works correctly", {
   skip_on_cran()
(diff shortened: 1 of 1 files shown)


R/watcher.r
```diff
@@ -47,11 +47,10 @@ watch <- function(path, callback, pattern = NULL, hash = TRUE) {
...
 safe_digest <- function(path) {
-  if (!file.exists(path)) return(NA_character_)
-  if (is_directory(path)) return(NA_character_)
-  if (!is_readable(path)) return(NA_character_)
-
-  digest::digest(path, file = TRUE)
+  tryCatch(
+    digest::digest(path, file = TRUE),
+    error = function(e) NA_character_
+  )
```
eddelbuettel: Not my project but as @krlmlr pinged me here:  I find the now-removed lines clearer and easier to read than what you suggest as a replacement.

eddelbuettel:
@krlmlr You we could talk over at littler about how to return NA rather than calling error.

R/watcher.r
krlmlr: Thanks for the feedback. Unfortunately, the previous approach isn't perfectly robust -- someone could modify `path` between the `is_readable()` and `digest()` calls. I agree that this is unlikely, still it looks like only the `tryCatch` approach is really error-prone. That's why I pinged you -- if `digest()` returned `NA` everything would be much easier here.

hadley: testthat used to use the `tryCatch()` strategy but the problem is that it silently ignored other types of errors apart from file existence.

tests/testthat/test-watcher.r
```diff
@@ -47,7 +47,7 @@ test_that("watcher works correctly", {
   dir.create(test_path)
 
   delayed.bash.cmd <- function(command) {
-    system(paste0("sleep 1;", command), wait=FALSE)
+    system(paste0("bash -c 'sleep 1;", command, "'"), wait=FALSE)
```
hadley: Does this make it work on windows?

R/watcher.r
krlmlr: I'll remove this patch for now. Still, short of trying, there seems to be no way to safely tell if `digest()` will succeed.

tests/testthat/test-watcher.r
krlmlr: If `bash` is on the PATH, then probably yes. Updated the test.

hadley:
Thanks!

PR #300, opened 2015-09-26
use Catch to enable C++ unit testing

kevinushey:


kevinushey:
The error looks like the build failed while attempting to ascertain code coverage. @hadley, can you review?

R/test-compiled-code.r
```diff
@@ -0,0 +1,121 @@
+# Geenrates a (semi-random) directory wherein compilation
+# can take place, within tempdir().
+cpp_compilation_tempdir <- function(package) {
+
+  # Generate a path wherein we perform the executable compilation
+  random_string <- paste(
```
hadley: Why don't you just use `tempfile()`?

NEWS.md
```diff
@@ -11,6 +11,14 @@
 
 * Failure locations are now formated as R error locations.
 
+* The [Catch](https://github.com/philsquared/Catch) unit
+  testing library can now be used for unit testing of C++
+  files. You can place C++ files within the `tests/testthat/cpp`
```
hadley: Why not just put them in `tests/thatthat/`?

NEWS.md
```diff
@@ -11,6 +11,14 @@
...
+  folder named as `test-*.cpp`, and `test_dir()` and friends
+  will automatically compile and run tested examples. See
+  `system.file(package = "testthat", "resources", "test-example.cpp")`
```
hadley: For the sake of testing (and code coverage!) it would be useful to include an example

hadley: Oh, you have, oops!

R/test-package.r
```diff
@@ -53,12 +53,17 @@ test_package <- function(package, filter = NULL, reporter = "summary", ...) {
...
-run_tests <- function(package, test_path, filter, reporter, ...)
-{
+run_tests <- function(package, test_path, filter, reporter, ...) {
   reporter <- find_reporter(reporter)
   env <- test_pkg_env(package)
   res <- with_top_env(env, {
-    test_dir(test_path, reporter = reporter, env = env, filter = filter, ...)
+    results <- test_dir(test_path, reporter = reporter, env = env, filter = filter, ...)
+
+    # TODO: Parse output and append success / failure to results
```
hadley: This seems like a fairly major todo - what happens if the tests fail?

R/test-compiled-code.r
kevinushey: I wasn't aware of the `pattern` argument to `tempfile()` -- I'll use that instead.

NEWS.md
kevinushey: I was worried about potential collisions with other filters that e.g. just look for `test-.*` -- can I expect that all pre-existing filters / commands are properly scoped (and so only pick up `R` test files in that folder)?

R/test-package.r
kevinushey: Nothing :( You get test output, but no reported failure or anything of that form.

If necessary, we can either write a parser for Catch test output that can then translate results as needed, but I wasn't sure exactly what would need to be done here.

NEWS.md
kevinushey: It looks like `find_test_scripts` does the right thing: `files <- dir(path, "^test.*\\.[rR]$", full.names = TRUE)`. I'll follow its cue.

kevinushey:
@hadley, it turns out including Makevars is a bit of a headache for a couple reasons:

1) `R CMD check` runs its tests in a context where the tests have been separated from the package, and
2) Launching a new R process with a different working directory during `R CMD check` will fail due to the absence of `startup.Rs`, which appears to be some funky mechanism R CMD check uses to inject some options.

Let me know if you have any better ideas for how we could properly respect `Makevars` for test compilation.

kevinushey:
Closing in lieu of updated PR.

PR #302, opened 2015-09-29
use Catch for unit testing of compiled code

kevinushey:
NOTE: not quite ready yet, but this is getting close now.

This PR uses Catch to run package tests. The general idea is this:

1) Unit tests are included in C++ files, using a testthat-like syntax with basic verbs `context`, `test_that`, `expect_true` and `expect_false`.

2) Packages which want to include these unit tests can simply copy the `test-runner.cpp` file from the `testthat` package into their `src/` directory (will add some helpers to make that easy),

3) `testthat::test_compiled_code()` will attempt to look for a routine called `run_testthat_tests()` as an exported routine from the package currently being tested; if it exists it will be run and unit test results will be collected.


hadley:
Like the overall flow!

kevinushey:
With the current flow, client packages could enable C++ unit testing by:

- Adding `LinkingTo: testthat` to their package's `DESCRIPTION`,

- Including a file called `test-runner.cpp` with the contents:

```
#define TESTTHAT_TEST_RUNNER
#include <tests/testthat.h>
```

- Using regular testthat-like tests throughout their other C++ files in the `src/` directory.

Note that the test runner is also included in the testthat package as `system.file(package = "testthat", "resources/test-runner.cpp")`.

kevinushey:
Looks like things are failing due to a problem with the `TAP` reporter right now; otherwise I think this should be ready for review.

kevinushey:
Oh, the tap reporter breaks because the output is captured by `capture.output` in the report tests. I'll look into it.

kevinushey:
@hadley can you give this another round of review? You might want to look at `use_catch()` which is the main function client packages can use to bootstrap themselves into unit testing -- let me know what you think. (Note that most of the ugliness in that package is in ensuring that `LinkingTo: testthat` exists in the `DESCRIPTION`; I'm not sure if `devtools` or otherwise has a more full-fledged `DESCRIPTION` reader / writer.

kevinushey:
Looks like AppVeyor failed for unrelated reasons. @hadley, let me know if you have time to review once more.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # testthat 0.10.0.9000
 
+* Added [Catch](https://github.com/philsquared/Catch) for unit testing of C++ code.
```
hadley: This'll now need to be moved

R/test-compiled-code.r
```diff
@@ -0,0 +1,148 @@
...
+#' for running these tests will be compiled into the \R package,
+#' with the entry point \code{run_testthat_tests()}. \code{testthat}
+#' will use that entry point to run your unit tests when detected.
+#'
+#' @param dir The directory containing an \R package.
+#'
+#' @export
+#' @seealso \href{https://github.com/philsquared/Catch}{Catch}, the
+#'   library used to enable C++ unit testing.
+use_catch <- function(dir = getwd()) {
```
hadley: I think it would be more R like to put this function at the top of the file

R/test-files.r
```diff
@@ -41,12 +41,16 @@ test_files <- function(paths, reporter = "summary",
...
   current_reporter$start_reporter()
+
   results <- lapply(paths, test_file, env = env,
     reporter = current_reporter, start_end_reporter = FALSE)
+
   current_reporter$end_reporter()
 
-  results <- unlist(results, recursive = FALSE)
+  # TODO: Integrate results.
+  compiled_results <- test_compiled_code(reporter = reporter)
```
hadley: It seems weird to run this here and not have a `path` argument

R/utils.r
```diff
@@ -103,3 +103,5 @@ skip_on_appveyor <- function() {
 
   skip("On Appveyor")
 }
+
+trim_whitespace <- function(text) gsub("^\\s*|\\s*$", "", text, perl = TRUE)
```
hadley: Why `perl = TRUE`?

R/test-compiled-code.r
```diff
@@ -0,0 +1,148 @@
...
+
+  # Copy the test runner.
+  success <- file.copy(
+    system.file(package = "testthat", "resources", "test-runner.cpp"),
+    test_runner_path,
+    overwrite = TRUE
+  )
+
+  if (!success)
+    stop("failed to copy 'test-runner.cpp' to '", src_dir, "'")
```
hadley: Please use `call. = FALSE`

R/test-compiled-code.r
```diff
@@ -0,0 +1,148 @@
...
+#' @seealso \href{https://github.com/philsquared/Catch}{Catch}, the
+#'   library used to enable C++ unit testing.
+use_catch <- function(dir = getwd()) {
+
+  desc_path <- file.path(dir, "DESCRIPTION")
+  if (!file.exists(desc_path))
+    stop("no DESCRIPTION file at path '", desc_path, "'")
+
+  src_dir <- file.path(dir, "src")
+  if (!file.exists(src_dir))
```
hadley: `if (!file.exists(src_dir)) && !dir.create(src_dir)` ?

R/test-compiled-code.r
```diff
@@ -0,0 +1,148 @@
...
+    output <- utils::capture.output(status <- .Call(routine)),
+    error = function(e) {
+      warning(sprintf("failed to call test entrypoint '%s'", routine))
+    }
+  )
+
+  if (reporter_type(reporter) != "silent" && !identical(output, ""))
+    cat("\nC++ unit test results:", output[-1], sep = "\n")
+
+  status == 0
```
hadley: I wonder if this would be better if you did `expect_equal(status, 0)`? Then you could eliminate the `cat()` above - you might a custom expectation to give a more informative error message for failure.

hadley:
I wonder if a better strategy would be to add `tests/testthat/test-cpp.R` containing something like:

```R
context("cpp")
test_compiled_code("pkname")
```

That would require a little more work when calling `use_catch()` but I think it would considerably simplify the interface with the rest of testthat.

NEWS.md
kevinushey: Maybe we should leave NEWS commits until after merge? I've had to amend this many times now because of NEWS updated from various pull requests. :/

R/test-files.r
kevinushey: Because the tests are a property of the package itself (and so perhaps this isn't really the most natural place to fit it, but I couldn't think of a better place)

R/utils.r
kevinushey: No real reason beyond it being a bit faster.

kevinushey:
Great idea re: adding the `test-cpp.R` file and returning custom expectation. Now everything fits a lot more naturally (and I cleaned up some of the now unneeded modifications to `test-files.r` and friends).

kevinushey:
Do you have any strong feelings about including the tests header with `<tests/testthat.h>` vs. `<testthat.h>`?

R/test-compiled-code.r
```diff
@@ -0,0 +1,131 @@
...
+    output <- utils::capture.output(tests_passed <- .Call(routine)),
+    error = function(e) {
+      warning(sprintf("failed to call test entrypoint '%s'", routine))
+    }
+  )
+
+  # Drop first line of output (it's jut a '####' delimiter)
+  info <- paste(output[-1], collapse = "\n")
+
+  expect_that(tests_passed, function(result) {
```
hadley: Can you please look at `expect_silent()` to see how a modern expectation should look?

R/test-compiled-code.r
```diff
@@ -0,0 +1,131 @@
+#' Test Compiled Code in a Package
+#'
+#' Test compiled code in the package \code{package}. See
+#' \code{\link{use_catch}()} for more details.
+#'
+#' @param package The name of the package to test.
+#'
+#' @export
+test_compiled_code <- function(package) {
```
hadley: And maybe this should now be `expect_cpp_correct()` or similar?

R/test-compiled-code.r
hadley: Post refactoring this is now at the right place

hadley:
This is looking great :)  `#include <testthat.h>` seems most natural to me.

kevinushey:
Okay, I think this is good to go!

hadley:
Thanks!!

PR #305, opened 2015-10-03
A few inverted expectiations

### LINKED ISSUE #303: Negation of tests
For "old-style expectations", there is `not()` -- e.g., `expect_that(object, not(is_null(...)))`: http://stackoverflow.com/a/30037495/946850. I'm not aware of a way to achieve this with the "new style". How about the following syntax:

```
n(expect_null)(object)
```

The package could also define

```
nexpect_null <- n(expect_null)
```

(same for all other functions).

At the core is the function `nexpect_that()` which is `expect_that()` with a negated condition. The function `n()` would be implemented by substituting `expect_that` with `nexpect_that` in the body.

krlmlr:
- ~~`expect_success()` -- if failure should not abort the test~~
- `expect_less_than_or_equal()` and `expect_more_than_or_equal()`

With tests. Modeled as "new-style" expectations, the arguments of the `_or_equal` variants differ.

Also adds `expect_silent()` to the "expectations" family.

`expect_unnamed()` seems unnecessary, as `expect_named()` can already test for absence of names.

@hadley: Do you want to export `expect()` at some point?

Fixes #303.


R/expectations-success.R
```diff
@@ -0,0 +1,25 @@
...
+#' @export
+#' @family expectations
+#' @examples
+#' expect_success(print("Hi!"))
+#' \dontrun{
+#' expect_success(stop("What?"))
+#' }
+expect_success <- function(expr) {
+  label <- find_expr("expr")
+  out <- evaluate_promise(expr)
```
hadley: I think you accidentally included this line

krlmlr: Good catch. I was wondering if this function necessary at all, we already have `expect_error(..., NA)`

hadley: Yeah, that's a good point.

krlmlr:
Removed expect_success().

Another option would be to support an optional argument: `expect_less_than(..., allow_equal = FALSE)`. This would be more in spirit of the other functions: expect_named(..., NULL), expect_error(..., NA), expect_warning(..., NA), ...

hadley:
I'd also feel better if they were named `expect_lt`, `expect_lte` etc. The current names are a bit verbose.

krlmlr:
Added shorter versions: lt, lte, mt, mte

hadley:
I think gt (greater than) is the more common abbreviation

krlmlr:
Now using gt.

hadley:
Thanks!

PR #307, opened 2015-10-04
Support shorthand syntax reporter1+reporter2+...

krlmlr:
for creating a multi-reporter that reports to reporter1, reporter2 and the others.

Now `find_reporter("")` throws an error.

Use case: `devtools::test(reporter = "summary+stop")` in a Makefile.

@hadley: I'm not sure which of the two below is the better syntax:

```
devtools::test(reporter = "summary+stop")
devtools::test(reporter = c("summary", "stop"))
```

I could also support both. I still need to add documentation and NEWS, please let me know what you think before I proceed.

PR #308, opened 2015-10-04
new FailReporter

krlmlr:
single purpose: fail at end if any errors or failures were detected

Difference to StopReporter: The latter stops immediately, the new reporter proceeds until end of tests.


krlmlr:
This works best with #307 (`reporter = "summary+fail"`). Am I missing a better way?

krlmlr:
Probably this should be default behavior for all reporters. Think about implementing that logic in the code that calls end_reporter().

krlmlr:
Merged with master. Should the StopReporter stop after the first event (error/failure/warning), and not wait until the end of the test?

PR #309, opened 2015-10-11
Fix comparison of integers

krlmlr:
Revert to default comparison also if x is integer.

Current behavior:

```
> testthat::expect_equal(0L, as.Date("2015-10-10Z"))
Error in `-.Date`(x, y) : can only subtract from "Date" objects
9: stop("can only subtract from \"Date\" objects")
8: `-.Date`(x, y)
7: format(x - y, digits = 3)
6: paste0(format(x, digits = 3), " - ", format(y, digits = 3), " == ",
       format(x - y, digits = 3))
5: compare.numeric(actual, expected, ...)
4: compare(actual, expected, ...)
3: condition(object)
2: expect_that(object, equals(expected, label = expected.label,
       ...), info = info, label = label)
1: testthat::expect_equal(0L, as.Date("2015-10-10Z"))

> testthat::expect_equal(0L, "0")
Error in x - y : non-numeric argument to binary operator
7: format(x - y, digits = 3)
6: paste0(format(x, digits = 3), " - ", format(y, digits = 3), " == ",
       format(x - y, digits = 3))
5: compare.numeric(actual, expected, ...)
4: compare(actual, expected, ...)
3: condition(object)
2: expect_that(object, equals(expected, label = expected.label,
       ...), info = info, label = label)
1: testthat::expect_equal(0L, "0")
```


R/compare.r
```diff
@@ -145,7 +145,7 @@ compare.numeric <- function(x, y, ..., max_diffs = 10) {
 
   # If they're not the same type or length, fallback to default method
   equal <- paste0(equal, collapse = "\n")
-  if (!is.integer(x) && !is.numeric(y)) return(comparison(FALSE, equal))
```
hadley: Maybe a better check would be `!identical(class(x), class(y))` ?

krlmlr: Currently:

```
> testthat::expect_equal(1L, 0)
Error: 1L not equal to 0
1 - 0 == 1
```

I guess this case (and others) will be handled differently if comparing classes of x and y.

hadley: I think that's probably reasonable - it doesn't really matter what the values are if the classes are different.  But a bigger problem is `expect_equal(1, 1L)` which for convenience shouldn't fail. Maybe we should special case integer vs. numeric comparison?

This current behaviour is also bad:

```R
expect_equal(1L, factor(1))
#  Error: 1L not equal to factor(1)
# 1 - 1 == NA In addition: Warning message:
# In Ops.factor(x, y) : ‘-’ not meaningful for factors
```

hadley: And this gives an uninformative message:

```R
x <- structure(1L, class = "myclass")
expect_equal(1L, x)
#  Error: 1L not equal to x
# 1 - 1 == 0 
```

krlmlr:
Updated: Added support for complex, fallback to string comparison if length different, otherwise use numeric comparison only for "true" numbers.

krlmlr:
This probably conflicts with #313.

krlmlr:
Merged with master.

PR #313, opened 2015-10-23
compare.numeric message fix for named/unnamed objects

perryjer1:
When calling `compare.numeric` to compare numerics that were equal except for names, the message from `all.equal` (via `attr.all.equal`) was being dropped. See the test_compare.r additions for examples. The solution I put in compare.r was to just add in the original message so the user can at least see it, but I don't know if this is the best way to handle it.


hadley:
I fixed it by totally rewriting `compare.numeric` - I think it should be a lot more robust now

PR #319, opened 2015-11-10
More informative output in expect_identical()

krlmlr:
- using compare()


NEWS.md
```diff
@@ -3,6 +3,9 @@
 * Added [Catch](https://github.com/philsquared/Catch) for unit testing of C++ code.
   See `?use_catch()` for more details. (@kevinushey)
 
+* `expect_identical()` and `is_identical_to()` now use `compare()` for more
```
hadley: FWIW I think it's safer to always put new NEWS items at the top, so they're anchored at a fixed position.

PR #317, opened 2015-11-07
throws_error(NA) has unexpected behavior

krlmlr:
- with regexp=NA and an actual error thrown
- failure message with successful code and expected error

Also affects expect_error(..., NA).

Solved by adapting implementation to that of gives_warning().

PR #320, opened 2015-11-13
new expectation: code runs without errors

famuvie:
I often use the following structure for testing that something **does not**
throw errors:

```R
res <- try(expr, silent = TRUE)
expect_false(inherits(res, 'try-error'))
```

which, in the case of failure, gives a message like:
> Error: inherits(res, "try-error") isn't false

I propose instead:
```R
expect_try(expr)
```

which I find much cleaner.
Furthermore, in the case of failure, the message is also cleaner:
> Error: stop() threw an error

### DIFF
--- a/R/expectations.r
+++ b/R/expectations.r
@@ -219,6 +219,42 @@ normalise_names <- function(x, ignore.order = FALSE, ignore.case = FALSE) {
   x
 }
 
+#' Expectation: does object executes without error?
+#'
+#' You can either check for the presence of names (leaving \code{expected}
+#' blank), specific names (by suppling a vector of names), or absence of
+#' names (with \code{NULL}).
+#'
+#' @inheritParams expect_that
+#' @family expectations
+#' @export
+#' @examples
+#' expect_try(1)
+#'
+#' \dontrun{
+#' expect_try(stop('fail'))
+#' }
+expect_try <- function (object, info = NULL, label = NULL) {
+  if (is.null(label)) {
+    label <- find_expr("object")
+  }
+  expect_that(eval(substitute(object)), isnot_error(), info, label)
+}
+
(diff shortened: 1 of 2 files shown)


famuvie:
I admit the function names are not great.
Suggestions welcome :)

hadley:
You can just do `expect_error(..., NA)`

famuvie:
Ok! I was not aware of this.
Nonetheless, the error message is not very informative:
```R
expect_error(stop(), NA)
```
> Error: is.character(regexp) is not TRUE

krlmlr:
@famuvie: This is handled in #317, could you please check and comment there if it works for you?

famuvie:
For the record, to test the outcome of some function, I am using the following structure from now on:

```R
  good_function <- function(x) cat('I expect this to work fine'); return(TRUE)
  res.try <- expect_error(
    res <- good_function(),
    NA
  )
  
  if (res.try$passed) {
    # ...
    # further expectations about res
  }
```
In this way, if the function fails, I get only one error.
I don't want to see errors for all the expectations on the results, since all are consequences of the first.

krlmlr:
I'm wondering why you're using expect_error() at all. An unchecked error (outside of try() or expect_error) already aborts the test. To clean up resources, use on.exit(), [withr](https://github.com/jimhester/withr) or [later](https://github.com/kevinushey/later).

famuvie:
Mmm... true. I think it was a reminiscence of previous practice.
Sometimes I need to run `good_function()` **outside** a specific test, in order to use its outcome within several different tests.
In that context, I need to run the function within `try()`, otherwise a failure would break all the testing process.
Then I kept using `try()` within tests, without realizing it was needless.
Thank you.

PR #334, opened 2015-12-10
Add error handling for test_file

cfhammill:
I was running in to problems with a test suite that relied on running test_dir, errors in the test files that exist outside of any particular test caused the whole suite to stop. The proposed change allows the suite to continue running while reporting an error for the broken test file.

### DIFF
--- a/R/test-files.r
+++ b/R/test-files.r
@@ -155,7 +155,19 @@ test_file <- function(path, reporter = "summary", env = test_env(),
   fname <- basename(path)
   lister$start_file(fname)
 
-  sys.source2(fname, new.env(parent = env))
+  tryCatch({
+    sys.source2(fname, new.env(parent = env))
+  }, error = function(e){
+    context(fname)
+    
+    warning(
+      paste0("File sourcing failed, likely due to either an error in untested code ",
+             "or a malformed file name"))
+    
+    fail("File Sourcing Failed")
+    end_context()
+  })
+
   end_context()
 
   if (start_end_reporter) reporter$end_reporter()
--- a/tests/testthat/test-line_numbers.r
+++ b/tests/testthat/test-line_numbers.r
@@ -3,34 +3,34 @@ context("Line Numbers")
 test_that("line numbers are found and given to reporters", {
     ## a reporter that keeps its results
   GreedyReporter <- setRefClass("GreedyReporter", contains = "Reporter",
-      where = environment(), 
+      where = environment(),
       fields = list(results = "list"),
       methods = list(
           add_result = function(result) {
               results[[length(results) + 1]] <<- result
          }
     ))
-  
+
   # get the results supplied to the reporter by expectations
   .test_code <- function(code, reporter = GreedyReporter$new(),
                         path = tempfile(fileext = ".R")) {
     code <- sub("^\\n", '', code) # strip first empty line if any
     writeLines(code, path)
     on.exit(unlink(path))
-    
+
     test_file(path, reporter)
     reporter$results
   }
-    
+
   .test_and_fetch_lines <- function(code) {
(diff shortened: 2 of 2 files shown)


R/test-files.r
```diff
@@ -155,7 +155,15 @@ test_file <- function(path, reporter = "summary", env = test_env(),
   fname <- basename(path)
   lister$start_file(fname)
 
-  sys.source2(fname, new.env(parent = env))
+  tryCatch({
+    sys.source2(fname, new.env(parent = env))
+  }, error = function(e){
+    test_that(
```
hadley: I like the idea, but I'm not sure this will work correctly. I think it should be more like:

```R
context(fname)
fail("Parsing failed")
```

cfhammill:
I made the change you suggested, it allows my test suite to run properly. Any idea why the AppVeyor build failed?

hadley:
Looks like the change breaks the unit tests: https://travis-ci.org/hadley/testthat/builds/96101967#L5475

tests/testthat/test-line_numbers.r
```diff
@@ -29,7 +29,7 @@ test_that("line numbers are found and given to reporters", {
   ### ==== EDGE CASES ====
   
   # test file with errors, e.g. unknown function
-  expect_error(.test_code('expect_toto()'), regexp = 'Error')
+  expect_warning(.test_code('expect_toto()'), regexp = 'File sourcing')
```
robertzk: A potential (pedantic) comment may be to make the quoting consistent with your change on line 44 (both single quotes or both double quotes).

robertzk:
@cfhammill Tests are failing due to

```
ERROR: dependency ‘git2r’ is not available for package ‘devtools’
* removing ‘/usr/local/lib/R/site-library/devtools’
ERROR: dependency ‘devtools’ is not available for package ‘covr’
* removing ‘/usr/local/lib/R/site-library/covr’
```

Does not look like you were the genesis of the error. I would try to rebase and [squish your commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) into a single one to re-trigger a build (and de-clutter your commits!).

```
git rebase -i 
```

cfhammill:
All checks passing, thanks for the tip @robertzk. I also switched to consistent quoting.

R/test-files.r
```diff
@@ -155,7 +155,19 @@ test_file <- function(path, reporter = "summary", env = test_env(),
...
+  tryCatch({
+    sys.source2(fname, new.env(parent = env))
+  }, error = function(e){
+    context(fname)
+    
+    warning(
+      paste0("File sourcing failed, likely due to either an error in untested code ",
+             "or a malformed file name"))
+    
+    fail("File Sourcing Failed")
```
hadley: Seems like it would be better to put everything in the failure message, i.e. `Failed to source "fname".` (I don't think you need such a verbose warning)

hadley:
I tried this out locally and it didn't work. I'm not sure if something in the internals of testthat has changed, but I don't think this strategy is correct. I tried out a few other things but I can't see an obvious approach here.

cfhammill:
What was failing? 

I still think an error outside of an individual test but within test_file should be captured as a testing failure, not abort the entire test_dir. Although perhaps there is a way to do this more elegantly, I'm not sure. 

We have unisolated code in our test suite to avoid recalculating expensive expressions, something I'd imagine is relatively common. The better answer is probably to memoize the slow stuff.

hadley:
It just threw a regular error.

cfhammill:
hmm, seems strange the tryCatch wouldn't catch it. I'll have look when I get a chance

hadley:
It does catch it, but it makes no difference because it's not inside a test.

PR #338, opened 2015-12-23
compare.numeric() message fix for names and attributes

e3bo:
I developed this fix before noticing that a similar fix is already available for pulling at #313. I'm still sending this pull request just in case this approach is preferable for some reason.


hadley:
I fixed it by totally rewriting `compare.numeric` - I think it should be a lot more robust now

PR #340, opened 2016-01-05
Add new-style equivalent for takes_less_than

richfitz:
This pull request adds new-style expectation support for `takes_less_than`, so rather than

```r
expect_that(myexpression, takes_less_than(1))
```

we have

```r
expect_takes_less_than(myexpression, 1)
```

I've tried to follow the pattern in other expectation migrations and added the old-style function to `oldskool`; there are a lot of resulting man changes as a result.

This may have been omitted from the new-style tests for a reason - if so please just close this.


hadley:
Thanks - can you please add a bullet to news too?

hadley:
Oh, hmmm, I think I was soft-deprecating this function because it's a bit unreliable (i.e. it's not a great idea to have stochastic behaviour in your tests)

PR #346, opened 2016-01-24
add skip_if_no_internet()

jennybc:
Celebrating @jeroenooms's addition of `curl::nslookup()` (https://github.com/jeroenooms/curl/blob/master/R/nslookup.R).

I don't necessarily expect this to be merged as is. Some questions:

  * Is it ok to add a skip function that depends on another package (`curl` in this case)?
  * Is there a better way inside `skip_if_no_internet()` to say "require that curl is installed and version is >= the version declared in DESCRIPTION"?
  * How would you write a test for this?


gaborcsardi:
Unfortunately this is not so simple, for several reasons:
* Most importantly, a successful DNS lookup does not mean that you are online. Your home router can simply answer it from its cache, while still having no connection to the rest of the internet. Similarly, any router on the way can answer it, e.g. the router at your ISP.
* An unsuccessful DNS lookup does not mean that you are offline, depending on what you want to do. Maybe only your DNS server is down.
* If the DNS query fails, it might take a couple seconds to fail, so you would not want to use this function in too many tests. Or you would want to cache the result.
* Even if you can resolve a hostname, that does not mean that you can actually connect to it, it might still be down. You need to try to connect to it.

Plus there are additional difficulties as well, e.g. which host to test (if you just want a generic `am_i_online()` test), how to test it (ping it, trying to connect via HTTP, etc.).

I am not a network expert, so FIXME (@jeroenooms probably knows this better than me), but I think it is better to try an HTTP connection, especially if you eventually want to connect to a web server in the tests. Note that this is not perfect, either, because you might be behind a proxy, and the proxy might answer the query. Maybe if you querying a random URL, that the proxy has no chance of replying to, is better. 

Anyway, in summary, I suggest to have a `skip_if_host_down()` function instead.

I am doing this in several packages. (Althouh not the random URL part.) E.g.: https://github.com/MangoTheCat/remotes/blob/5d319f4d82d80cc5798e55db0397ea8a062fbcdd/tests/testthat/helper.R#L2-L10
which in turn uses the `pingr` package with a 0.2 second timeout:
https://github.com/gaborcsardi/pingr/blob/6112c9697d363082868d09b5a0613f26188c02f2/R/ping-package.r#L9-L35

@sckott also has a package for this: https://github.com/sckott/httping 
`httping` is R only, but it relies on `httr` to do the work, which is a pity, because `testthat` is lightweight currently, but won't be with `httr` as a dependency.

`pingr` is self-contained, but has compiled code, and my TCP code is probably inferior to libcurl, although as I (vaguely) remember, `pingr` handles timeouts better.

If you want a generic `is_online()` test, that is even harder, because you need to pick a host to check, and also a protocol. It probably makes sense to pick two root name servers, and connect to them directly for a DNS lookup. AFAIR some smart routers even intercept this, and answer, and some proxy servers will not let the direct query through. Another way would be to connect to some server that is almost always online, e.g. google or amazon.

jeroenooms:
Either approach will do, it depends on what exactly you consider a working connection. I think a dns lookup is the best test for general connectivity, R itself uses nsl as well to [test if internet](https://github.com/wch/r-source/blob/master/tests/internet.R#L1-L11) is available.

Yes there is a case when you have an ip address but your dns server is down. However I would argue that in this case you don't have a working internet connection because dns is required everywhere. Of course being able to resolve a hostname does not mean you can connect to it, but that is not what we want to test. We're just testing if we are online ourselves, not if any given host/service is online.

I don't think you have to worry about performance. Operating systems do a really good job caching dns queries, without false positives:

```r
> system.time(curl::nslookup("www.google.nl"))
   user  system elapsed 
  0.004   0.001   0.038 
> system.time(curl::nslookup("www.google.nl"))
   user  system elapsed 
  0.001   0.000   0.001
```

But as soon as I disable my connection it will fail:

```r
> system.time(curl::nslookup("www.google.nl"))
 Error in curl::nslookup("www.google.nl") : 
  Unable to resolve host: www.google.nl 
```

But I agree with Gabor that you should test what it is that you actually need. If you just want to know if you're online, a dns lookup is a nice minimal test. If your package talks to a certain API, then actually testing if you can connect to the API server is probably better.

gaborcsardi:
My main problem with a simple DNS query is that if you unplug your router's uplink, then it's built-in DNS server might reply from its cache. So it is not reliable. Maybe your ISP is reliable, but mine is terrible, so this actually happens to me.

> I don't think you have to worry about performance. 

Yes, I think timeouts are mainly a problem if you connect via HTTP and the server just ignores your packets. E.g. this takes 10 seconds. You can set the timeout to one second, but smaller values do not work, I don't know why.

```r
> system.time(GET("http://2.2.2.2"))
Error in curl::curl_fetch_memory(url, handle = handle) :
  Timeout was reached
Timing stopped at: 0.005 0.001 10.01
```

Btw. here is a somewhat sophisticated solution in JS:
https://github.com/sindresorhus/is-online/blob/46ec687974c7e0816bee6cfc595315c1a7a4bf49/index.js
It tries a direct DNS connection to a random root name server, and if that fails, then maybe your are behind a proxy, so it tries to http ping four web servers that are not supposed to be down at the same time (google, baidu, etc.).

Anyway, I think whatever the test does, it might be a good idea, to just `skip_on_cran()` everything that needs it, to be safe. :) E.g.:
https://github.com/MangoTheCat/remotes/blob/master/tests/testthat/test-install-github.R#L6-L8

Also, I would not mind avoiding the dependency on curl. :)

jennybc:
Wow.

I will react to the above as a `testthat` user.

> If your package talks to a certain API, then actually pinging the API server is probably better.

Yes, I am just trying to hit a specific API (GitHub, today), so I see that it makes sense to specify that in the `skip_*()` function.

> ...it might take a couple seconds to fail, so you would not want to use this function in too many tests. Or you would want to cache the result.

From my very limited use of the version I proposed, there is a noticeable pause in `testthat`'s progress when it hits affected tests. I had already wondered if/how to cache the result within a test run.

> Anyway, I think whatever the test does, it might be a good idea, to just skip_on_cran() everything that needs it, to be safe. 

Agreed! Abusing terminology, it seems like `skip_if_cant_reach(https://api.github.com)`, or whatever it gets called, should "inherit" from `skip_on_cran(). Do you actually build that in or do you just always call them together, like in @gaborcsardi's example?

So... will one of you propose a proper alternative? Everyone I ask about this problem has implemented it for themselves but it would be nice to have some of these solutions in `testthat` itself.

jennybc:
One more thing:

> @sckott also has a package for this: https://github.com/sckott/httping 
httping is R only, but it relies on httr to do the work, which is a pity, because testthat is lightweight currently, but won't be with httr as a dependency.

But you only have to `Suggest` whatever package you use (`httr` or `curl` or whatever), right? Plus anyone who is testing against, say, an API already depends on both of these anyway.

sckott:
> Plus anyone who is testing against, say, an API probably already depends on both of these anyway.

Probably, but I do understand gabor's desire to keep this pkg light weight

> httping ... relies on httr to do the work, which is a pity, because testthat is lightweight currently, but won't be with httr as a dependency.

was designed it to work alongside `httr` - I thought of it more for interactive use cases, not so much a light weight utility

gaborcsardi:
> > httping ... relies on httr to do the work, which is a pity, because testthat is lightweight currently, but won't be with httr as a dependency.

> was designed it to work alongside httr - I thought of it more for interactive use cases, not so much a light weight utility

@sckott Sorry, I did not mean that this is a deficiency of `httping`. It makes perfect sense to rely on `httr`, especially as there is effectively nothing in base R to rely on. I was just thinking about the current use case for `testthat`.

gaborcsardi:
@jennybc I would keep `skip_if_cant_reach()` separate from `skip_on_cran()`. They are functionally orthogonal.

I would still suggest to skip these tests on CRAN, just for the peace of mind.

Yes, you are right that a Suggests is not too bad. `devtools` is already there, and that brings in `httr` and `curl`, anyway.

sckott:
> I would still suggest to skip these tests on CRAN, just for the peace of mind.

i agree with this

jennybc:
Today's GitHub outage gave me an object lesson in why I would want my skipper to be specific to the resource I am trying to reach vs. general internet access. Message received.

jennybc:
Closing this. I clearly cannot implement the general skippers discussed above and I have discovered <https://status.github.com/api/status.json> to build around for my immediate need.

PR #348, opened 2016-01-25
Modified test_pkg_env to work with the byte-code compiler.

kalibera:
The original version of test_pkg_env does not work with the R byte-code compiler: test_pkg_env creates a top-level environment, which is not a namespace, but has some namespaces as its parents. In the modified version, the top-level environment is a namespace, which makes the compiler happy. The namespace is copied so that it will stay unlocked - some package tests rely on that they can trigger S4 registration, which requires writing to the namespace. The modification comes with a small semantic change: in the original version, the top-level environment of the test was not the same as the package namespace, so one could e.g. create a local variable overshadowing a package function, but still be able to call that package function. That is perhaps not a good style and it was found in only one package out of all CRAN+BIOC. It is easy to fix in the package tests.


jackwasey:
The package was mine, and @kalibera's pull request https://github.com/jackwasey/icd9/pull/79 is accepted. Not sure how long until this makes it to CRAN, however.

jimhester:
Do you have a simple example of the byte-code compiler failure with the previous code?

We could also do something akin to https://github.com/hadley/devtools/blob/81dd313e1d143a3bdaaa2e1f09c887801db66694/R/namespace-env.r#L46-L73 to make a more proper namespace.

I guess the main concern with this patch is it copies all the objects in the namespace twice, the original approach feels cleaner to me.

kalibera:
I will forward to you the email discussion we had on this with Hadley, it has much more background. To answer your question, the previous testthat code fails with bytecode compiler e.g. when checking the changepoint package:

`env R_ENABLE_JIT=3 R_COMPILER_OPTIMIZE=3 R_COMPILER_SUPPRESS_ALL=true _R_CHECK_FORCE_SUGGESTS_=false ./bin/R CMD check changepoint_2.2.tar.gz`

The (old) byte-code compiler (I tested now with revision 69996, and older revision will behave the same) will fall into an infinite loop, the error displayed will be "C stack usage XXX is too close to the limit". We have modified the compiler to report a (proper) error instead, but as it would break all code that depends on old testthat, we have hidden this error and the compiler falls back to AST interpreter (I tested now revision 70046 and newer should behave similarly - if you wanted to see the error, print it in tryCmpfun and tryCompile). So, after this modification the test will work, but still the problem has to be fixed, it is just hidden.

jimhester:
Tomas,

Thank you for filling me in by forwarding me the correspondence with hadley.

Using the following patch modifies the `test_pkg_env` environment to be a pseudo-namespace and return `TRUE` for `isNamespace()`. This approach preserves the prior behavior and does not have to copy any objects. This prevents the compiler error you were seeing and your command now runs without error on my machine (I verified it failed previously).
```diff
diff --git a/R/test-package.r b/R/test-package.r
index 15fe64b..cb218e1 100644
--- a/R/test-package.r
+++ b/R/test-package.r
@@ -1,6 +1,12 @@
 test_pkg_env <- function(package) {
+
+  # Create a pseudo namespace to avoid issues with the R compiler
   env <- new.env(parent = getNamespace(package))
-  # Supress warning messages from S4
+  info <- new.env(hash = TRUE, parent = baseenv())
+  assign(".__NAMESPACE__.", info, envir = env)
+  assign("spec", c(name = paste0("test_", package)), envir = info)
+
+  # Suppress warning messages from S4
   env$.packageName <- package
   env
 }
```
```shell
env R_ENABLE_JIT=3 R_COMPILER_OPTIMIZE=3 R_COMPILER_SUPPRESS_ALL=true _R_CHECK_FORCE_SUGGESTS_=false R CMD check changepoint_2.2.tar.gz
* using log directory ‘/Users/jhester/pkg/changepoint.Rcheck’
* using R version 3.2.3 (2015-12-10)
* using platform: x86_64-apple-darwin15.2.0 (64-bit)
* using session charset: UTF-8
* checking for file ‘changepoint/DESCRIPTION’ ... OK
* checking extension type ... Package
* this is package ‘changepoint’ version ‘2.2’
* checking package namespace information ... OK
* checking package dependencies ... OK
* checking if this is a source package ... OK
* checking if there is a namespace ... OK
* checking for executable files ... OK
* checking for hidden files and directories ... OK
* checking for portable file names ... OK
* checking for sufficient/correct file permissions ... OK
* checking whether package ‘changepoint’ can be installed ... OK
* checking installed package size ... OK
* checking package directory ... OK
* checking DESCRIPTION meta-information ... OK
* checking top-level files ... OK
* checking for left-over files ... OK
* checking index information ... OK
* checking package subdirectories ... OK
* checking R files for non-ASCII characters ... OK
* checking R files for syntax errors ... OK
* checking whether the package can be loaded ... OK
* checking whether the package can be loaded with stated dependencies ... OK
* checking whether the package can be unloaded cleanly ... OK
* checking whether the namespace can be loaded with stated dependencies ... OK
* checking whether the namespace can be unloaded cleanly ... OK
* checking loading without being on the library search path ... OK
* checking dependencies in R code ... OK
* checking S3 generic/method consistency ... OK
* checking replacement functions ... OK
* checking foreign function calls ... OK
* checking R code for possible problems ... OK
* checking Rd files ... OK
* checking Rd metadata ... OK
* checking Rd cross-references ... OK
* checking for missing documentation entries ... OK
* checking for code/documentation mismatches ... OK
* checking Rd \usage sections ... OK
* checking Rd contents ... OK
* checking for unstated dependencies in examples ... OK
* checking contents of ‘data’ directory ... OK
* checking data for non-ASCII characters ... OK
* checking data for ASCII and uncompressed saves ... OK
* checking line endings in C/C++/Fortran sources/headers ... OK
* checking compiled code ... OK
* checking examples ... OK
* checking for unstated dependencies in ‘tests’ ... OK
* checking tests ...
  Running ‘testthat.R’
 OK
* checking PDF version of manual ... OK
* DONE
```

kalibera:
Hi Jim,

I am sorry I forgot to tell you I tried this solution as well. It makes 
the compiler happy with changepoint. But it breaks the compiler e.g. it 
with evaluate - it confuses the compiler by the fake namespace. The 
packages I was mostly debugging my version on:

changepoint
evaluate
aoos
swirl
pryr
DBI
data.table

I run with
env R_ENABLE_JIT=3 R_COMPILER_OPTIMIZE=3 R_COMPILER_SUPPRESS_ALL=true 
_R_CHECK_FORCE_SUGGESTS_=false ./bin/R CMD check evaluate_0.8.tar.gz

Best,
Tomas

jimhester:
Alright from messing with this a little more trying to get it to work I am convinced your patch is the best solution. Sorry for the detour and thank you for your work on the R compiler!

kalibera:
Thanks!

kalibera:
I've added one more fix to this pull request, this one is for testthat internal tests (test-context.r). In original testthat, the CountReporter class is created in a fresh environment that pretends to be the testthat package namespace. Such environment confuses the byte-code compiler. With my fix of test_pkg_env, this environment is not needed anymore, so I've removed it.

R/test-package.r
```diff
@@ -1,8 +1,6 @@
 test_pkg_env <- function(package) {
-  env <- new.env(parent = getNamespace(package))
-  # Supress warning messages from S4
-  env$.packageName <- package
-  env
+  list2env(as.list(getNamespace(package), all.names=TRUE),
```
hadley: Can you please match existing style here and put spaces around `=`?

kalibera: Sure, added.

hadley:
Thanks!

PR #343, opened 2016-01-13
Summary of skipped tests

krlmlr:
Shows skipped tests, but only if there are no failures. Example:

```
Testing bigrquery
DBItest: Connection : ..SS..S.SS
1. Skip: DBItest: cannot_query_disconnected --------------------------------------------------------------------------------------------------------------
Not yet implemented

2. Skip: DBItest: cannot_disconnect_twice ----------------------------------------------------------------------------------------------------------------
by request

3. Skip: DBItest: stress_connections ---------------------------------------------------------------------------------------------------------------------
by request

4. Skip: DBItest: get_info_connection --------------------------------------------------------------------------------------------------------------------
Not yet implemented

5. Skip: DBItest: stress_load_connect_unload -------------------------------------------------------------------------------------------------------------
by request

DONE 
```

Will improve the coding once we agree on the concept.

Fixes #234.

CC @HarlanH @kenahoo-windlogics

Fixes #349 (=includes it).

PR #349, opened 2016-01-25
extract cat_reports()

krlmlr:
for #343.

This will also restrict the detailed output to length(labels) = 62 errors at most.


krlmlr:
Prefer S3 class with print method.

hadley:
i.e. `print.testthat_reports()` and a minimal constructor.

PR #355, opened 2016-01-27
Use with_reporter() internally

krlmlr:
for better control over the testing logic (e.g., stop if encountering any error: #308).

In the long run, I think the code that's now in test_file() should be vectorized to allow for several files (perhaps extract the logic to a new function test_file_one()). Then test_files() simply calls test_file().


R/test-files.r
```diff
@@ -40,10 +40,11 @@ test_files <- function(paths, reporter = "summary",
   if (length(paths) == 0) stop('No matching test file in dir')
 
   current_reporter <- find_reporter(reporter)
-  current_reporter$start_reporter()
-  results <- lapply(paths, test_file, env = env,
-    reporter = current_reporter, start_end_reporter = FALSE)
-  current_reporter$end_reporter()
+  with_reporter(
```
hadley: I think this would be clearer as `results <- with_reporter(current_reporter, ...)`

hadley:
Otherwise, looks good.

R/test-files.r
krlmlr: with_reporter() returns the reporter, not the results of the code. At least one package (checkmate) depends on this behavior. I'll add documentation.

krlmlr:
Documentation added.

hadley:
LGTM. Can you please add a news bullet and merge?

PR #358, opened 2016-02-03
set_reporter() returns old reporter invisibly

krlmlr:
to avoid surprises.


krlmlr:
This otherwise trivial patch also edits the NEWS in two places, please consider it first.

hadley:
Thanks

Issue #356, opened 2016-01-29
FR: DebugReporter


krlmlr:
A reporter that interrupts the test with a `recover()` call at the first failed expectation. Implementation sketch:

```
DebugReporter <- setRefClass("DebugReporter", contains = "Reporter",
  methods = list(
    add_result = function(result) {
      if (!result$passed) {
        recover()
      }
    }
  )
)
```

Would be great to handle conditions, too (errors, warnings, skips; #343, #310), but to me this looks more difficult to implement.

krlmlr:
On the other hand, if failures are special conditions (#215), this might be easier with `options(error = recover)`.

hadley:
It feels like this is better attacked using `options(error = recover).`

krlmlr:
This works for me with #360 and the following code in a test file, before the context() call:

```
library(testthat)
test_that <- function(name, code) {
  withr::with_dir("tests/testthat", code)
}
options(error = recover)
```

recover() is not the ideal choice here, because a bunch of calls to testthat's internals are included in the stack trace (cc @gaborcsardi). Also, remember to leave the debugger before restarting RStudio, otherwise the working directory remains set to tests/testthat.

Issue #310, opened 2015-10-16
Capture warnings and messages and print all at end


hadley:
Display on summary as W.

gaborcsardi:
:+1:

krlmlr:
Perhaps related: Would it be possible to print a dot or something for each evaluated top-level expression (in addition to the dot for each successful expectation)?

gaborcsardi:
@krlmlr: That would create a lot of dots imo. Why not just put the ones that you want output for in an expectation, e.g. `expect_silent`?

krlmlr:
Agreed. There will be more dots, and it could also hurt test performance. I can print a dot myself with `expect_null(NULL)`, too.

krlmlr:
I think that both warnings and messages should be handled by the reporter, the default handling for messages would be to muffle them. I'm currently struggling with a case where I'd really like to see the messages that occur within a test.

hadley:
I think that's reasonable, although I think the default should be to handle messages in the same way as warnings (i.e. with an M and a summary at the end)

hadley:
This is a first stab at it - please let me know if you have suggestions for improvement

krlmlr:
Thanks. This looks very nice, I like it.

- Are you also going to handle messages the same way?
- I've looked at dplyr results, most warnings come from testthat: "info"/"label" has been deprecated. I wonder if these should be handled specially.
- The warnings that occur outside of test_that() calls are collected separately.
- Do you plan to treat warnings as errors eventually? I understand that this is not an option for this release.

hadley:
* I think I'll leave messages for now. They seem like more like output where it's reasonable to ignore them by default - otherwise it makes for frustrating testing.  

* I'm as concerned as I used to be about capturing warnings etc that happen outside of tests.

* You mean treat as test failures? Maybe that could be an option, but I don't see it as being particularly important.

krlmlr:
1. I'd appreciate at least an "M" in the SummaryReporter, perhaps verbose output should be "opt-in".
2. I don't understand this -- does that mean you don't care about what happens outside tests?
3. I'd argue that warning=failure should be the default, but I fear that many packages that use testthat are not ready for this.

hadley:
1. I'm concerned that even that is too much output for most people. 

1. Right - if it's not in a test, why should testthat be involved?

1. It's not clear to me that warnings should be failures; regardless, I think it's too late to change now.

Issue #218, opened 2015-01-09
Deprecate label & info arguments


hadley:
In favour of teaching people how to accomplish the same thing with NSE.

krlmlr:
Could you please show an example?

hadley:
Look at "building your own testing tools" under http://r-pkgs.had.co.nz/tests.html#test-tests

sergiizaskaleta:
But how will `desc` from `test_that(..)` help when using TeamCity reporter? The only option for reasonable test names are labels

PR #360, opened 2016-02-05
Use conditions for reporting results

krlmlr:
Also: test_that() returns TRUE if successful (no failed expectations, errors, or skips) for programmatic use of testthat.

Bare expectations now simply stop if a test fails. The stop occurs upon the first event, unlike the StopReporter which stops after the first failed test.


R/expect-that.r
```diff
@@ -37,7 +37,18 @@ expect_that <- function(object, condition, info = NULL, label = NULL) {
...
-  get_reporter()$add_result(results)
+  report_results(results)
+}
+
+report_results <- function(results) {
+  cond <- structure(list(results = results, message = format(results)),
+                    class = c("test_result", "condition"))
+
+  withRestarts(
+    if (results$passed) signalCondition(cond) else stop(cond),
```
krlmlr: Could also call `message()` instead of `signalCondition()` here.

krlmlr:
@hadley @jimhester: PTAL. Part of https://github.com/hadley/testthat/issues/215#issuecomment-75367557.

R/test-that.r
```diff
@@ -73,6 +82,23 @@ test_code <- function(description, code, env) {
   invisible(ok)
 }
 
+report_results <- function(results) {
+  cond <- structure(list(results = results, message = format(results)),
+                    class = c("test_result", "condition"))
+
+  withRestarts(
+    {
+      signalCondition(cond)
```
krlmlr: Call stop() here.

R/test-that.r
```diff
@@ -73,6 +82,28 @@ test_code <- function(description, code, env) {
   invisible(ok)
 }
 
+report_results <- function(results) {
```
hadley: I think this would be better called `report_expectation()`, but I'm not sure `report()` is quite the right verb (but I'm not sure what would be better)

R/test-that.r
```diff
@@ -73,6 +82,28 @@ test_code <- function(description, code, env) {
...
+    raise_condition_from_result(results),
+    continue_test = function(e) NULL
+  )
+
+  invisible(results)
+}
+
+raise_condition_from_result <- function(results) {
+  if (results$passed) {
+    cond <- structure(list(results = results),
```
hadley: Could we combine this object representation with the expectation, so there's only one type of thing to worry about?

R/test-that.r
hadley: Or should this all be wrapped up into `expectation()`?

krlmlr: Do you mean that expectation() should create and raise the condition object? This means that we need to be *very* clever in expect_that(): Catch the condition (which is created in the condition(object) call), patch it, and re-raise it. To me, this looks like too much effort for too little gain. But I agree that expectation() should return a proper condition object right away. I'll update the PR soon.

hadley: Maybe we should have `expectation()` (noun) to create the object, and `expect()` (verb) raise the condition?

krlmlr: We already have expect() which accepts a logical and a message; we'll need as.expectation.logical() and as.expectation.expectation() to avoid changing too much unrelated code, but otherwise this is doable. How does that sound?

krlmlr: Then we'll add as.expectation.error(), as.expectation.skip() (as replacement for current expecation_error() and expectation_skip()) and later as.expectation.warning() (#310).

R/expectation.r
```diff
@@ -11,20 +11,73 @@
...
+#' @export
+as.expectation.logical <- function(x, message, ...) {
+  expectation(passed = x, failure_msg = message, srcref = find_test_srcref())
+}
+
+#' @export
+as.expectation.error <- function(x, ...) {
+  error <- x$message
+  calls <- x$calls
+  srcref <- x$srcref
```
krlmlr: The srcref handling needs to be improved. This is currently just a placeholder, should be filled later by the calling handler to support file+line reporting.

krlmlr:
PTAL: Now expectations are also conditions.

R/expectation.r
```diff
@@ -11,20 +11,73 @@
...
 #' @export
 expectation <- function(passed, failure_msg, success_msg = "unknown", srcref = NULL) {
-  structure(
+  new_expectation(passed = passed, failure_msg = failure_msg,
+                  success_msg = success_msg, srcref = srcref)
+}
+
+new_expectation <- function(failure_msg, success_msg, srcref, ...,
+                            passed = FALSE, error = FALSE, skipped = FALSE) {
+  if (passed) {
```
hadley: `passed && !error`?

hadley: Maybe it would be better to have `type = c("error", "failure", "skip", "warning", "message")` etc?

R/test-that.r
```diff
@@ -51,28 +51,55 @@ test_code <- function(description, code, env) {
     e$calls <- utils::head(sys.calls()[-seq_len(frame + 7)], -2)
     signalCondition(e)
   }
+  handle_expectation <- function(exp) {
+    get_reporter()$add_result(exp)
+    if (!isTRUE(exp$passed)) {
```
hadley: Maybe pull this out to separate `expectation_ok()` ?

hadley:
This is looking really good!

R/expectation.r
krlmlr: If `error`, then `passed` is `FALSE` by construction (although not enforced). I agree that a type is better, but I'd like to deal with this in a new batch of changes, and also change the reporters -- this PR is already big enough.

hadley: Sure. I think I just introduced some merge conflicts, but if you squash this, I'll fix.

krlmlr:
Are you planning further refactorings? Then it would be great to merge this beforehand, otherwise I'll work further on this branch.

hadley:
I don't have anything else major planned - I think most of the other issues need this PR completed first.

krlmlr:
Extracted expectation_ok(), got rid of success_msg in my code, added two TODO markers.

I think test_code() can be simplified considerably, so that there's basically just one handler function taking care of everything. The reporter should be able to decide if tryCatch() catches the errors or not (#356) :

```
tryCatch(stop(), error = NULL)
tryCatch(stop(), error = function(e) NULL)
```

I'll merge the other PR before adding NEWS to this PR.

R/expectation.r
```diff
@@ -10,19 +10,76 @@
...
+      failure_msg = failure_msg
     ),
-    class = "expectation"
+    class = class
   )
+
+  update_expectation(exp, srcref)
+}
+
+update_expectation <- function(exp, srcref, info = NULL, label = NULL) {
```
hadley: It feels like we should be able to eventually eliminate this function, wrapping it into the constructor

R/test-that.r
```diff
@@ -52,28 +52,55 @@ test_code <- function(description, code, env) {
     e$calls <- utils::head(sys.calls()[-seq_len(frame + 7)], -2)
     signalCondition(e)
   }
+  handle_expectation <- function(exp) {
+    get_reporter()$add_result(exp)
+    if (!expectation_ok(exp)) {
```
hadley: `ok <<- expectation_ok(exp)` ?

R/test-that.r
```diff
@@ -52,28 +52,55 @@ test_code <- function(description, code, env) {
...
+  withRestarts(
+    raise_condition(exp),
+    continue_test = function(e) NULL
+  )
+
+  invisible(exp)
+}
+
+raise_condition <- function(exp) {
+  if (exp$passed) {
```
hadley: `expectation_ok(exp)`

R/expectation.r
krlmlr: Difficult without finally deprecating info and label arguments (#218).

tests/testthat/test-return.R
```diff
@@ -0,0 +1,23 @@
+context("Return values")
```
hadley: I think this should probably go in `test_test_that.R`

tests/testthat/test-bare.r
```diff
@@ -12,3 +12,18 @@ expect_gte(3, 3)
 expect_less_than(2, 3)
 expect_lt(2, 3)
 expect_lte(2, 2)
+
+expect_error(expect_false(FALSE), NA)
```
hadley: I think you accidentally re-added this?

R/expectation.r
hadley: Yeah, lets look again once we've eliminated those args and refactored the expectations

R/test-that.r
krlmlr: `ok <<- ok && expectation_ok(exp)` if you prefer.

tests/testthat/test-bare.r
krlmlr: This was intentional, the commented code that was below is now gone.

krlmlr:
Done. One more thing: expectation() is exported, you have changed its interface. Perhaps it's safer to keep success_msg (with a warning), or introduce a ... .

hadley:
Yeah, I'm not too worried about that. I think the next release will be testthat 1.0.0 so it's ok to make a few breaking changes (and I doubt this one will affect many people)

hadley:
And looks good to merge - feel free whenever you think it's ready

PR #363, opened 2016-02-09
add class check to throws_error

dchudz:
I wasn't sure about adding this argument to `throws_error` rather than having two separate functions (one for the `regexp` check and one for the `class` check). I decided to put it all together so that people can concisely check both at the same time if they want to (and don't have to learn about a new function). Thoughts?

(As expressed in my unit tests, the `expectation` fails if either the `regexp` check or the `class` check fails.)

Also, I struggled a bit to concisely explain what this is about in the help text. I really wanted to link to http://adv-r.had.co.nz/Exceptions-Debugging.html, since I haven't seen this explained well anywhere else and many users will need a full explanation if they want to use this option.


hadley:
Can you please merge/rebase?  I think you'll probably need a little extra logic since @krlmlr's fixes.

dchudz:
@hadley: Hopefully you can help me with some questions this raises...

What would cause `length(errors) > 1`? Parallel processing?

If I can reproduce that situation, I'll know what kind of object `res` is in that case. (A list? A single object that encodes multiple errors?).

I guess my `expectation` should succeed if any of the errors are of the right class?

dchudz:
For context, the new code includes:

```
      expectation(
        length(errors) > 0,
        "no errors raised",
        paste0(length(errors), " errors raised")
```

... which makes me think `length(errors)` might be something other than just `0` or `1`. But I don't see any tests w/ `length(errors) > 1` so I'm not sure.

hadley:
Hmmmm, I don't think it's possible. It might be better to clean up the preceding code to make that more clear. Maybe something like:

```R
error <- tryCatch(
  {
    expr
    NULL
  },
  error = function(e) {
    e
  }
)
```

@krlmlr any thoughts?

dchudz:
... and (if that's correct) clean up the bit I quoted, right? No reason for "errors" to be plural if plural is never correct.

krlmlr:
My implementation of throws_error() is essentially a copy of gives_warnings(), and ignores the fact that there really can be only one error. Do you want to use

```r
error <- tryCatch(
  {
    expr
    NULL
  },
  error = function(e) {
    e$message %||% ""
  }
)
```

and later check for `is.null(error)`?

dchudz:
I can include the change with my PR, probably tonight.

dchudz:
Btw, I'm also thinking: 

- I should change my `class` argument to `expected_class` (for better clarity and to avoid masking the builtin `class` function)
- calling `throws_error` with `regexp=NA` and non-null `expected_class` shouldn't be allowed. (`regexp=NA` means no error is expected, which is inconsistent w/ expecting a certain error class.)

krlmlr:
Can `expected_class` be a vector?

dchudz:
You'd want a vector where we expect any one of the classes in the vector? (Makes sense, I could do that.)

krlmlr:
I thought about checking that the condition inherits from all classes, but it looks like the "any" check is the default behavior of inherits():

```
inherits(iris, c("POSIXt", "data.frame"))
## TRUE
```

But:

```
> inherits(iris, list("POSIXt", "data.frame"))
## Error in inherits(iris, list("POSIXt", "data.frame")) : 
##   'what' must be a character vector
```

dchudz:
The `any` check would be consistent with `assertRaises` in [Python's `unittest`](https://docs.python.org/2/library/unittest.html)

hadley:
I'd prefer to keep it consistent with the existing behaviour of inherits

dchudz:
Great - consistent with `inherits` and similar to `unittest` are both good things from my perspective, so I'm glad they're the same.

dchudz:
@hadley okay, this should be ready. 

The Travis build failure seems unrelated to anything I did, unless I'm misreading it.

I also introduced some minor formatting changes via RStudio's autoformatting (Cmd-I). I think they're all (minor) improvements, but I'll take them out if you want.

R/expectations-matches.R
```diff
@@ -126,39 +129,58 @@ prints_text <- function(regexp, ...) {
...
+    error <- tryCatch({
+      expr
+      NULL
+    },
+    error = function(e) {
+      e
+    })
+    
+    paste_with_commas <- function(x, y) paste(x, y, sep=", ")
+    expected_classes_as_string <- Reduce(paste_with_commas, expected_class)
```
hadley: `paste(x, collapse = ",")`?

R/expectations-matches.R
```diff
@@ -79,11 +83,11 @@ matches <- function(regexp, all = TRUE, ...) {
     matches <- grepl(regexp, char, ...)
     if (length(char) > 1) {
       values <- paste0("Actual values:\n",
-        paste0("* ", encodeString(char), collapse = "\n"))
+                       paste0("* ", encodeString(char), collapse = "\n"))
     } else {
```
hadley: I prefer to aligning indent for function definitions, but fixed 2 space indent for function calls.

R/expectations-matches.R
```diff
@@ -126,39 +129,58 @@ prints_text <- function(regexp, ...) {
...
+    
     if (identical(regexp, NA)) {
       expectation(
-        length(errors) == 0,
-        paste0("expected no errors:\n", paste("* ", errors, collapse = "\n")),
-        "no errors raised"
+        is.null(error),
+        paste0("expected no errors:\n", as.character(error),
+               no_error_string),
+        success_msg = no_error_string
```
hadley: Why did you name the argument here? Why does `no_error_string` appear in both the success and failure message?

R/expectations-matches.R
```diff
@@ -126,39 +129,58 @@ prints_text <- function(regexp, ...) {
...
-    } else if (!is.null(regexp) && length(errors) > 0) {
-      matches(regexp, ...)(errors)
-    } else {
-      expectation(
-        length(errors) > 0,
-        "no errors raised",
-        paste0(length(errors), " errors raised")
+    } else if (is.null(error)) {
+      expectation(FALSE, no_error_string) 
+    } else if (!is.null(expected_class) && !inherits(error, expected_class)) {
```
hadley: The check should be the first argument to `expectation()`

R/expectations-matches.R
```diff
@@ -126,39 +129,58 @@ prints_text <- function(regexp, ...) {
...
       )
-    } else if (!is.null(regexp) && length(errors) > 0) {
-      matches(regexp, ...)(errors)
-    } else {
-      expectation(
-        length(errors) > 0,
-        "no errors raised",
-        paste0(length(errors), " errors raised")
+    } else if (is.null(error)) {
+      expectation(FALSE, no_error_string)
```
hadley: This doesn't look right - surely it needs to check `!is.null(error)`

hadley: Oh I see - your if construction is confusing. I think it would be better to branch based only on the values of `regexp` and `expected_class`

dchudz: You're right it's confusing - I did it the way I did b/c I thought the failure message should depend on whether there's no error, or an error of the wrong type. So some branching on `!is.null(error)` is required (unless we want to give up on distinguishing in the message), but I can make it cleaner.

R/expectations-matches.R
dchudz: hmm, okay, I'll revert. You work at RStudio, right? You could try to convince them to do it your way. :)

R/expectations-matches.R
dchudz: Thanks, didn't know about that!

R/expectations-matches.R
dchudz: `no_error_string` is a success message when there is no error (as expected), but a failure message when we expect an error but there's none.

Btw, I think `regexp=NA` meaning we expect no error is part of what makes this confusing. It looks to me like that behavior was documented months ago but only actually worked after @krlmlr's fixes.

hadley:
Yeah, you're right, this function is a nightmare. I'm going to rewrite it, then you can take another stab if you want.

dchudz:
Sounds good. I also won't mind if you include my feature in the require yourself. :)

PR #372, opened 2016-02-19
colourise() indexes colors by test type

krlmlr:
- uses test types as color names

@hadley: This might conflict with other changes in the reporters.


krlmlr:
Completed, tests pass locally. The idea here is that the test colors are named the same as the expectation types.

hadley:
Nice - I was just thinking about that!

PR #375, opened 2016-02-22
Don't duplicate "error" class

krlmlr:
If type == "error", the "error" class doesn't need to be added

### DIFF
--- a/R/expectation.r
+++ b/R/expectation.r
@@ -22,7 +22,7 @@ expectation <- function(type, message, srcref = NULL) {
     class = c(
       "expectation",
       type,
-      if (type %in% c("failure", "error")) "error",
+      if (type == "failure") "error",
       "condition"
     )
   )


hadley:
I'm not sure whether or not its even necessary to add the `error` class. It's not obvious to me how the class and use of `stop()` interacts

krlmlr:
Only if you add the error class, tryCatch(error = ...) catches it:

```
tryCatch(signalCondition(structure(list(message = "Hi"), class = c("failure", "condition"))),
  error = function(x) "Caught")
## NULL
tryCatch(signalCondition(structure(list(message = "Hi"), class = c("error", "condition"))),
  error = function(x) "Caught")
## [1] "Caught"
```

hadley:
Yeah, but is a failure really an error? Is it a feature for it to be caught this way?

krlmlr:
Pro: Other similar packages raise errors, too:

```
> tryCatch(checkmate::assertInt(5.5), error = function(e) "Caught")
[1] "Caught"
> tryCatch(assertthat::assert_that(assertthat::is.number("a")), error = function(e) "Caught")
[1] "Caught"
> tryCatch(ensurer::ensure_that(5.5, is.integer), error = function(e) "Caught")
[1] "Caught"
```

Contra: Are there use cases where you want to catch an error but *not* a failure?

hadley:
I think it's more clear in those situations that you want an error. A failure is a bit different because you don't usually want to stop executing the rest of the code in the block.

krlmlr:
I see several design issues here:

1. Continue test after failure?
    - Current behavior: yes, sometimes undesirable
    - How to change behavior: wrap with function must() or the like: must(expect_true(some_test())) .
2. Behavior of bare expectations
    - I'd argue that failures should be signaled with stop(), like the other packages above (and also testit)
    - Adding the "error" class allows tryCatch(error = ...) for bare expectations, I think this is a good thing -- minimizes surprise
3. Compatibility with other tester packages
    - They could call testthat::fail(), which should continue raising a "proper" error in this case (see point 2)
    - Alternatively, if a tester package prefers a custom failure class, testthat could be made aware of it

PR #378, opened 2016-02-29
Use stopifnot() to test error behavior of bare expectations

krlmlr:
Otherwise the tests don't catch broken error behavior.


hadley:
Thanks!

PR #379, opened 2016-03-01
Fail test if reporter output differs

krlmlr:
Tests status of tests/testthat/reporters directory using git2r.

### DIFF
--- a/tests/testthat/test-reporter.R
+++ b/tests/testthat/test-reporter.R
@@ -38,9 +38,17 @@ test_that("character vector yields multi reporter", {
 })
 
 test_that("reporters produce consistent output", {
+  out_dir <- tempfile("testthat")
+  dir.create(out_dir)
+
   save_report <- function(name, reporter = find_reporter(name)) {
+    out_path <- file.path(out_dir, paste0(name, ".txt"))
     path <- file.path("reporters", paste0(name, ".txt"))
-    capture.output(test_file(test_path("reporters/tests.R"), reporter), file = path)
+    writeLines(character(), out_path)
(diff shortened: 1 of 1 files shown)


krlmlr:
Will rework to use plain text file comparison.

hadley:
How about a generic `expect_no_git_changes()` or similar?  

I also wondered if we wanted something specialised for regression testing - i.e. a variant of `test_that()`

krlmlr:
This PR is currently broken -- I don't understand why file.copy() isn't invoked.

How about `expect_output(..., file = "reporters/...txt")`, which would collect output and write it to the file (if it is different)? I'm not too convinced about integrating this with Git anymore -- these tests should run on CRAN (and for an installed package) as well.

hadley:
Also they should definitely be saved and checked into git as this makes it easy for me to visually inspect changes and verify whether or not they're ok.  This is more like a regression test, so I don't think it should fail automatically if different to previous.

krlmlr:
Agreed. But what if a third party changes the code but forgets to update the "golden" files? Failing the test gives you an early clue. If the files are updated in-place, the tests fail only once.

hadley:
That doesn't seem likely in this scenario, because I'd notice in the pull request. I really don't want automation to make this more fragile for me.

krlmlr:
You notice in the PR only if the contributor updates the "golden" files. Failing the test in case of a difference errs on the safe side and has only a minor impact for your work on the code (because files are still updated in-place). That said, feel free to close this if you don't like it.

hadley:
I'd rather think it through in general in the future. The current code suits me well for now

PR #380, opened 2016-03-01
Error not recorded if it happens in expect_message()

krlmlr:
Adapted reporter test, currently failing.

Closes #387 (=includes it).


krlmlr:
The error is recorded, it's just not printed, because it's handled when sink() is still active. This also affects expect_warning() and expect_output(). I've added the resulting differences to this PR. Please advise.

krlmlr:
Seems tricky. Best idea so far: Use a new `CachingReporter` in evaluate_promise() that defers actual reporting until the sink is closed again. Will try.

krlmlr:
Works for me now, PTAL. I'm not inheriting from SilentReporter because I need to forward context and test, too.

tests/testthat/reporters/check.txt
```diff
@@ -18,10 +18,18 @@ stop
 
 6. Error: Error:3 (@tests.R#36) ------------------------------------------------
 !
-1: f() at reporters/tests.R:36
-2: g() at reporters/tests.R:32
-3: h() at reporters/tests.R:33
-4: stop("!") at reporters/tests.R:34
+1: expect_message(f(), NA) at reporters/tests.R:36
```
hadley: You'll also need to tweak the constants in `test_code()` to avoid the extra stuff here

krlmlr: Done, requires #387.

krlmlr:
Depends on #387 (which needs #390); let's wait for these two.

hadley:
Can you please rebase/merge and I'll give a detailed review

tests/testthat/reporters/check.txt
```diff
@@ -18,7 +18,7 @@ stop
 
 6. Error: Error:3 (@tests.R#36) ------------------------------------------------
 !
-1: f() at reporters/tests.R:36
+1: f()
```
krlmlr: Rebased. I wonder if we can do anything about this srcref loss here.

R/evaluate-promise.R
```diff
@@ -39,9 +39,12 @@ evaluate_promise <- function(code, print = FALSE) {
   temp <- file()
   on.exit(close(temp))
 
-  result <- with_sink(temp,
-    withCallingHandlers(
-      withVisible(code), warning = wHandler, message = mHandler
+  with_reporter(
```
hadley: What about wrapping a `tryCatch` outside of the `with_sink()` so it can capture and rethrow the error?

tests/testthat/reporters/tests.R
```diff
@@ -33,7 +33,7 @@ test_that("Error:3", {
   g <- function() {h()}
   h <- function() {stop("!")}
 
-  f()
+  expect_message(f(), NA)
```
hadley: Do you need to change this line? The main test is below, right?

hadley: Ooooh, because the problem isn't catching the error message, it's that the `end_test()` method gets called inside the sink.

krlmlr: Right. We can only call "true" reporter functions outside of with_sink().

R/evaluate-promise.R
krlmlr: Tried that. If I capture the error, I can't handle multiple warnings; if I simply handle the condition, the top-level handler still is invoked, which means that the reporter prints to the sink.

hadley: Why can't you handle multiple warnings? Oh because the reporter also needs to print them?

krlmlr: The second warning wouldn't be run with tryCatch():

```
> tryCatch({warning(1); warning(2)}, warning = identity)$message
[1] "1"
```

Can't use withCallingHandlers() because this doesn't eat warnings (or I'd have to reimplement most of the logic in test_code() ).

Am I missing anything obvious?

hadley: I don't think so, but I'm not in love with this approach - it adds a lot of extra complexity. Let me think on it a bit more.

hadley:
I think I need a bit more info on this problem. Your test:

```R
reporter <- with_reporter("silent", {
  test_that("", expect_warning(stop()))
})
expect_equal(length(reporter$expectations()), 1)
expect_true(expectation_error(reporter$expectations()[[1L]]))
```

Currently works for me, and that makes sense - the error causes a premature exit from `evaluate_promise()`, so by the time the reporter prints something the sink has been terminated.

It's not clear your patch preserves the desirable behaviour of suppressing warnings when called from `expect_warning()`, while propagating them when called from `expect_message()`. I think what we actually need is finer control over evaluate promise so we can opt out of capturing messages or warnings. I'll tweak `evaluate_promise()` to do this.  

Can you have a go at creating a simple test case that illustrates the current problem?

krlmlr:
Summary of commits:

- fe2ea is the demo
- 133c7 is the fix, which solves the issue (with minor flaws)
- 27536 is the correction of the call stack, which (almost) restores the original behavior.

Could you sketch an example where this implementation would misbehave?

krlmlr:
I don't like this anymore either: My new DebugReporter (internal WIP) doesn't work here. Perhaps we can teach evaluate_promise() to temporarily close its sink when a reporter is activated.

hadley:
I think the way forward is to break `evaluate_promise()` into 3 separate pieces, one that captures outputs, one that captures warnings, and one that captures messages.

PR #382, opened 2016-03-01
Discussion: Don't use class "error" for failures?

krlmlr:
Follow-up to https://github.com/hadley/testthat/pull/375#issuecomment-187375108

Currently, the following code executes only the first expectation:

```
test_that("expect", {
  multi <- function() {
    expect_true(FALSE)
    expect_false(TRUE)
  }

  expect_error(multi(), NA)
})
```

(Added to the regression tests.)

This PR resolves this, and also the problems seen in https://github.com/hadley/testthat/pull/379#issuecomment-190721317. I've tried changing the implementation of expect_error() to no avail.

On the other hand, this behavior is not wrong if failures are explicitly considered as errors.

### DIFF
--- a/R/expectation.R
+++ b/R/expectation.R
@@ -22,7 +22,8 @@ expectation <- function(type, message, srcref = NULL) {
     class = c(
       "expectation",
       type,
-      if (type == "failure") "error",
+      # Failures cannot have class "error", otherwise each failed expectation
+      # inside expect_error() aborts the test
       "condition"
     )
   )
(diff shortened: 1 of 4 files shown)


hadley:
I'd rather you didn't test that in the reporters tests, but instead used the silent reporter to verify that you get two expectations for that code.

krlmlr:
Updated, PTAL.

R/expectation.R
```diff
@@ -22,7 +22,8 @@ expectation <- function(type, message, srcref = NULL) {
     class = c(
       "expectation",
       type,
-      if (type == "failure") "error",
+      # Failures cannot have class "error", otherwise each failed expectation
```
hadley: I think you can drop this comment.  Otherwise, feel free to merge

hadley:
That looks good to me.

PR #385, opened 2016-03-01
Remove old call-capturing code

krlmlr:
now handled in register_expectation().

### DIFF
--- a/R/test-that.R
+++ b/R/test-that.R
@@ -37,6 +37,8 @@ test_code <- function(test, code, env = test_env()) {
 
   ok <- TRUE
   register_expectation <- function(e, start_frame, end_frame) {
+    # Capture call stack, removing calls from end (added by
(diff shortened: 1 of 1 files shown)


hadley:
I'm pretty sure that's still needed because of `signalCondition(e)`.  It's a bit ugly and confusing.

krlmlr:
That depends if you want to rethrow the original error, or an updated version. I think this doesn't matter now. Who's going to look at the $call component of the rethrown error? In the original code, this was the tryCatch(error = ) handler; in the current code, there is an empty error handler.

hadley:
Maybe for the default reporter so you get a more informative stacktrace for a bare test? I'm almost certain I put it there deliberately.

krlmlr:
You're right -- reporter output gets distorted (and the tests don't catch it -- #379)...

PR #387, opened 2016-03-01
Use format(result) instead of result

krlmlr:
when printing expectations.


R/expectation.R
```diff
@@ -136,15 +136,9 @@ as.expectation.error <- function(x, ..., srcref = NULL) {
...
-    user_calls <- paste0(traceback, collapse = "\n")
-    msg <- paste0(msg, "\n", user_calls)
-  } else {
-    # Need to remove trailing newline from error message to be consistent
-    # with other messages
-    msg <- gsub("\n$", "", msg)
-  }
+  # Need to remove trailing newline from error message to be consistent
+  # with other messages
+  msg <- gsub("\n$", "", msg)
```
hadley: Should this also go in `format()`?

R/expectation.R
```diff
@@ -178,10 +172,15 @@ print.expectation <- function(x, ...) cat(format(x), "\n")
 #' @export
 format.expectation <- function(x, ...) {
```
hadley: I wonder if we should a separate method for `success` (and I wonder if the class names should actually `expectation_success()`, `expectation_skipped()` etc)

krlmlr: format.success() will never be called if format.expectation() exists.

hadley: Then we have our classes around the wrong way

R/expectation.R
krlmlr: Don't know -- this was code specific to error conditions.

PR #388, opened 2016-03-01
Satisfy R CMD check

krlmlr:
- Qualified access to setOldClass
- utils::capture.output()
- Fix error in example


hadley:
Thanks!

PR #396, opened 2016-03-05
Test that functions are exported from the package namespace

nealrichardson:
More times than I'd like to admit, I've added new functions to a package and forgotten to export them from the namespace. It's really easy to do because `test_check` runs tests inside the package namespace. That's great for unit testing, but it's less helpful for validating the package interface being provided to your users. 

In the interest of avoiding past mistakes, in the [httpcache](https://github.com/nealrichardson/httpcache) package tests, I experimented with a `public` context that evaluates test code in the global environment, rather than the package environment: 

    public <- function (...) with(globalenv(), ...)

It had the [desired effect](https://github.com/nealrichardson/httpcache/blob/master/tests/testthat/test-zzz-helper.R#L3-L12) in that only exported functions could be found inside a `public()` block, so it would fail if I had failed to document and export something. But, I found that I was wrapping entire test files in `public({...})`, which was a bit repetitive. And it was still a little too manual, and thus forgettable. I wanted something that would be more foolproof and obvious.

So, I made a few tweaks to `run_tests` and `test_files`, such that if there is a subdirectory in your test path called "public", the code/tests in it will be run in the global environment rather than the package environment. The two main changes are (1) the check for that "public" directory and running `test_dir` in it, and (2) passing the `start_end_reporter` argument around a couple more places so that `run_tests` could call `test_dir` twice with the same reporter. The added tests assert that non-exported functions aren't visible in the public tests. 

Since this seemed like a solution to a problem that others may have, I figured I'd submit the pull request for your consideration. Happy to rework it if you see a better way to achieve this, and open to renaming if "public" doesn't seem like the right label.

### DIFF
--- a/R/test-files.R
+++ b/R/test-files.R
@@ -23,16 +23,17 @@ test_env <- function() {
 #'   regular expression will be executed.  Matching will take on the file
 #'   name after it has been stripped of \code{"test-"} and \code{".R"}.
 #' @param env environment in which to execute test suite.
+#' @param start_end_reporter whether to start and end the reporter
 #' @param ... Additional arguments passed to \code{grepl} to control filtering.
 #'
 #' @return the results as a "testthat_results" (list)
 #' @export
 test_dir <- function(path, filter = NULL, reporter = "summary",
-                                          env = test_env(), ...) {
+                     env = test_env(), start_end_reporter=TRUE, ...) {
   source_test_helpers(path, env)
   paths <- find_test_scripts(path, filter, ...)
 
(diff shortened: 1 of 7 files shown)


nealrichardson:
Not sure how to interpret that Travis failure in test-watcher.R. It passes locally and on my fork: https://travis-ci.org/nealrichardson/testthat/builds/113834565

hadley:
I usually use examples to catch that sort of problem. Using a subdirectory doesn't feel like the right solution to me.

PR #404, opened 2016-03-08
NULL reporter

krlmlr:
Does nothing in with_reporter(), provided that an active reporter exists.

Needed for dplyr testing as part of the dtplyr tests.

### DIFF
--- a/R/reporter-zzz.R
+++ b/R/reporter-zzz.R
@@ -41,8 +41,17 @@ get_reporter <- function() {
 with_reporter <- function(reporter, code, start_end_reporter = TRUE) {
   reporter <- find_reporter(reporter)
 
-  old <- set_reporter(reporter)
-  on.exit(set_reporter(old), add = TRUE)
+  if (!is.null(reporter)) {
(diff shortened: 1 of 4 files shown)


R/reporter-zzz.R
```diff
@@ -41,8 +41,17 @@ get_reporter <- function() {
 with_reporter <- function(reporter, code, start_end_reporter = TRUE) {
   reporter <- find_reporter(reporter)
 
-  old <- set_reporter(reporter)
-  on.exit(set_reporter(old), add = TRUE)
+  if (!is.null(reporter)) {
```
krlmlr: Alternative implementation:

```
find_reporter <- function(reporter, use_current = FALSE) {
  if (is.null(reporter)) {
    if (use_current) return(get_reporter()) else return(NULL)
  }
  ...
}
```

krlmlr:
Turns out this isn't needed (as get_reporter() is exported); closing for now. (Although we eventually might want to define sensible behavior for a "null" reporter.)

PR #406, opened 2016-03-10
Export as.expectation()

krlmlr:
as a drop-in replacement for the old expectation() call


krlmlr:
Prefer expect_true() instead (#407).

PR #408, opened 2016-03-10
Broken test: labelling if function value has been forced

krlmlr:
This occurred in one of the existing tests in RSQLite. @hadley: PTAL.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # testthat 0.11.0.9000
 
+## Breaking changes
+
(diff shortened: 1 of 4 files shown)


hadley:
I don't see any easy way around this - I think you'll just have to supply an explicit label.

krlmlr:
Could this be fixed in lazyeval?

hadley:
Possibly - the dev version at least has a better error message, but we'd need a variation of the (C) lazy function that only returned the expression, not the env (so being forced wouldn't be a problem)

PR #411, opened 2016-03-12
avoid use of now-removed 'copy()' method

kevinushey:
Fixes https://github.com/hadley/testthat/issues/405.


hadley:
Are you sure that's sufficient? I was thinking an R6 clone would be needed

kevinushey:
It looks like `find_reporter()` will just construct a new reporter when the argument is a string, but I think you're right that explicitly calling `find_reporter()` first and then cloning is the way to go.

kevinushey:
Alright, now done using R6 clone rather than copy.

hadley:
Thanks!

PR #395, opened 2016-03-03
Test error output in expect_message()

krlmlr:
with update of golden files to show the problem. Demo of the problem #380 tries to solve.

PR #420, opened 2016-03-22
Reporters always write to true stdout

krlmlr:
Works by memoizing `stdout()` on creation, and writing strictly to the memoized connection (and not to the possibly overridden standard output).

```
> std <- stdout(); with_sink(file(), cat("a", file = std))
a
```

Adding the functionality to Reporter was easiest, other designs might be better.

Follow-up for #395.

Closes #419 (=contains it).


R/reporter.R
```diff
@@ -18,6 +18,24 @@ Reporter <- R6::R6Class("Reporter",
...
+    cat = function(..., file = NULL, sep = " ", fill = FALSE, labels = NULL,
+                   append = NULL) {
+      if (!is.null(file)) {
+        warning("file ignored", call. = FALSE)
+      }
+      if (!is.null(append)) {
+        warning("append ignored", call. = FALSE)
+      }
+
+      cat(..., file = self$out, sep = sep, fill = fill, labels = labels)
```
hadley: Rather than having our own clone of `cat()`, I'd rather have something that's tweaked to make like a bit easier:

```R
cat_line <- function(...) {
  cat(..., "\n", sep = "")
}
```

Not sure what to call the version that doesn't add a nl

hadley:
This looks like a great approach!

krlmlr:
Thanks. From a design perspective it probably should be a mixin (can R6 do it?), or a separate class that is instantiated and used by each interested reporter.

R/reporter.R
krlmlr: I'd like to tackle cat() refactoring in a separate PR, this is already big enough.

Caveat: Not all cat() calls have sep = "" in place, but the integration tests should catch this.

krlmlr:
Added NEWS.

krlmlr:
Also added cat_tight(), cat_line() and cat_paragraph().

R/reporter-tap.R
```diff
@@ -29,20 +29,21 @@ TapReporter <- R6::R6Class("TapReporter", inherit = Reporter,
       if (!self$has_tests)
         return()
 
-      cat("1..", self$n, '\n', sep = '')
+      self$cat_line("1..", self$n)
       for (i in 1:self$n) {
         if (!is.na(self$contexts[i])) {
-          cat("# Context", self$contexts[i], "\n")
+          self$cat_line("# Context ", self$contexts[i], " ")
```
krlmlr: @dkesh: I'm reworking output for the TAP reporter. Does it require trailing whitespace, or would it work without them?

hadley:
Thanks - this is great :smile:

R/reporter-tap.R
dkesh: Trailing whitespace isn't required by TAP.

PR #423, opened 2016-03-23
Use covr v2

jimhester:
New version works locally, whereas I am getting strange errors with the
CRAN version locally


hadley:
Thanks!

PR #426, opened 2016-03-24
Remove trailing whitespace

krlmlr:
for TAP output:

@dkesh: Does output look right to you?


dkesh:
Honestly, we've still been having trouble with jenkins reading TAP output--not for tests that succeed, but on parsing failing tests. I just haven't put enough effort into figuring out why.  But there is no regression in this commit, AFAICT.

PR #439, opened 2016-04-07
implement standalone signal mechanism for reporters

kevinushey:
This PR implements a simple signal / slot system, and uses it for signalling expectation results outside of `test_that` calls.


hadley:
Would you mind commenting somewhere that this is a bit of a hack? (And obviously only helps with the summary reporter)

Alternatively, maybe I should just decide that this is a feature?

PR #449, opened 2016-04-10
don't use partial argument match in deparse

jackwasey:
Following the advice of Kevin Ushey at https://kevinushey.github.io/blog/2015/02/02/rprofile-essentials/ I turned on warnings for partial argument matching and found this.


krlmlr:
Could you please merge/rebase against master to rerun the test?

PR #443, opened 2016-04-09
New expect_output_file()

krlmlr:
Compares output against file contents, and optionally updates it.

With test, and using it already in this package.

I'm not really happy with the name, any suggestions?

PR #455, opened 2016-04-16
don't use partial argument matching

jackwasey:
updated version of #443, with three updates to use full rather than partial argument matching.


krlmlr:
LGTM.

hadley:
Thanks!

PR #464, opened 2016-04-22
fix for solaris

jeroenooms:


hadley:
Thanks - can you confirm that once this is in place the unit tests pass on Solaris?

jeroenooms:
Yes it now passes check on both sparc and x86.

jeroenooms:
We can only guess what timezone value BDR is going to get on his box though... maybe you want to ignore the timezone attribute alltogether?

hadley:
I wish I could, but I think it's important.  Let me resubmit again, with an explanatory note.

PR #465, opened 2016-04-25
bugfix: auto_test_package does not work on windows machines

FrancoisGuillem:
On windows, auto_test_package does not automatically run tests when a file is changed. This is because on windows, paths may contain "/" or "\".

The problem is solved by calling normalizePath.


hadley:
But where are the bad slashes coming from?

cuche27:
file.path puts "/" but normalizePath uses only "\\". Since the watcher calls normalizePath, it expects to see only paths with "\\".

kevinushey:
For reference:

```
kevinushey@Kevin-MBP-2:~/git/testthat [master]
$ ag -Q "normalizePath"
R/auto-test.R
36:  code_path <- normalizePath(code_path)
37:  test_path <- normalizePath(test_path)
45:    changed <- normalizePath(c(added, modified))
96:    changed <- normalizePath(c(added, modified))

R/source.R
34:  files <- normalizePath(sort(dir(path, pattern, full.names = TRUE)))

R/test-path.R
33:  path <- normalizePath(path)

tests/testthat/test-source_dir.R
7:  res <- source_dir(normalizePath('test_dir'), pattern = 'hello', chdir = TRUE)
13:  res <- source_dir(normalizePath('test_dir'), pattern = 'hello', chdir = FALSE)
```

Should these all have `winslash = "/"`?

cuche27:
It could help for future developments, but I think the important thing is to always use normalizePath when manipulating paths.

hadley:
Could you please add a bullet point to NEWS?

cuche27:
Done ! Sorry for the delay.

hadley:
Thanks!

PR #470, opened 2016-05-07
New DebugReporter

krlmlr:
Calls a better version of `recover()` that shows only the interesting part of the call stack. Fully tested.

Fixes #360.


R/reporter-debug.R
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        recover(start_frame = result$start_frame,
+                end_frame = result$end_frame)
+      }
+    }
+  )
+)
+
+# Copied from utils::recover()
```
hadley: Can you please also copy the copyright statement?  It might be better to move into completely a new file. Will also need to update `Authors@R` to reference R core

hadley:
LGTM

PR #472, opened 2016-05-12
adding a mock object API

lbartnik:
this is a proposal of a new functionality in testthat - mock objects;
I haven't really made it complete yet and it is meant to serve as an example of what could be done

my goal here is to see whether this is interesting for other developers at all


krlmlr:
So the purpose is to record the calls to the mock object? Very nice!

I'm not sure we really need to cycle the return values. Also, it would be nice if unevaluated expressions can be passed:

```
mock <- function(..., envir = parent.frame()) {
 ...
}

mock(1, 2, 3)
mock(NULL, fun1("a"), fun2("b"), NULL)
```

In the second example, fun1 and fun2 are called only upon the second and third invocation of the mock object.

lbartnik:
I thought of modelling this around Python's unittest.mock (https://docs.python.org/3/library/unittest.mock.html). Here we are dealing with a functions, not an object - so parts of Python's API are not relevant.

- recording calls: yes, that's an important functionality, I'm happy it looks OK
- cycling return values: maybe that's not a good idea - e.g. if you want to make sure only a given number of calls is made in the code under testing
- passing unevaluated expressions: could you please elaborate on that? what is the difference of passing such expressions versus just a plain value?

Also, unittest.mock has the `side_effect` parameter which is a function called upon each call to the mock object. It can change the `return_value` parameter and throw exceptions, among others. Maybe we could keep the `returns` or `return_value` parameters and add `side_effect` in `testthat.mock` ?

krlmlr:
Cycling may or may not be desired, we could add an argument `cycle = FALSE` that controls this.

Unevaluated expressions will solve both side effects and return values. Instead of evaluating the arguments automatically, you capture the expressions behind them and evaluate in a controlled fashion. See `with_mock()` for the case of unnamed arguments, and http://adv-r.had.co.nz/Computing-on-the-language.html and http://adv-r.had.co.nz/Expressions.html for more detail.

lbartnik:
Hi,
Actually, I like all of your ideas. I also thought it would be good to record not only the call but also the values of parameters passed in. Unfortunately, I cannot really work to implement all of this right now. I've made a few changes to prepare the API: it's `...` now instead of `returns` and I added `cycle`. I hope this will be enough to merge the change in because I'm already testing my software with `mock()`.

The funny thing is I checked locally with `covr` that I have 100% coverage but the report above says something else. However, the link takes me to 404 so it looks like something is broken...

R/mock-object.R
```diff
@@ -0,0 +1,81 @@
...
+#' @param mock A \code{\link{mock}} object.
+#' @param no Number of the call made on the \code{mock} object.
+#' @param expected_call Expected call.
+#' @family expectations
+#' @export
+#' @examples
+#' \dontrun{
+#' expect_call(mock, 1, summary(iris))
+#' }
+expect_call <- function (mock, no, expected_call) {
```
krlmlr: I'm not sure we need the new expect_call(). How about implementing `[[.mock()` which looks up a call (basically the first part of your expect_call() )? The rest could be handled with the existing expect_equal(). Perhaps `[.mock()` and `as.list.mock()` would be helpful, too.

R/mock-object.R
```diff
@@ -0,0 +1,81 @@
...
+#' \dontrun{
+#'   m <- mock(1)
+#'   with_mock(summary = m, {
+#'     expect_equal(summary(iris), 1)
+#'     expect_call(m, 1, summary(iris))
+#'   })
+#' }
+#'
+mock <- function (..., cycle = FALSE) {
+  returns <- list(...)
```
krlmlr: You could also say `returns <- eval(substitute(alist(...)))` and then use `eval(returns[[...]])` when evaluating the mocks. This will make sure the expressions are evaluated only when the mock is called, not at the start of the function.

R/mock-object.R
lbartnik: But then you'd have to wrap value of `expected_call` in `bquote` - today `expect_call` calls `substitute` on `expected_call` and test code is clearer and easier to write (no explicit `bquote`). Also, however less importantly, `expect_call` gives a very clear hint about test`s goal - improves readability, I'd say.

R/mock-object.R
lbartnik: That's a good point, I'll make the change.

lbartnik:
@krlmlr could you please help me with codecov? I cannot understand why my tests do not meet the expected coverage goals - and why "Details" take me to 404. Any ideas what I'm doing wrong?

krlmlr:
I'don't know which "Details" link you mean. Also, I don't understand why covr doesn't seem to catch your new tests. What do you get when you run `covr::package_coverage()` locally?

R/mock-object.R
krlmlr: Good point. How about `expect_call(m[[1]], summary(iris))`?

lbartnik: It looks good. Is there a strong preference for m[[1]] instead of m[1] ?

lbartnik:
Running this locally gives me 100% code coverage. If it doesn't catch up after my next commit I'll cancel PR and re-open it. Maybe then it'll work.

R/mock-object.R
krlmlr: I'd expect m[1] to return a list with one call, and m[[1]] to return just the call.

krlmlr:
I also get 100% coverage for the new mock.R file. That's strange. @jimhester: Any idea why local coverage might differ from that on Travis?

jimhester:
Not sure could be an issue with the new codecov site or some other issue, I will look at it tomorrow.

R/mock-object.R
lbartnik: That sounds good but...

- this is not a list nor a vector, but a function - so we don't have to follow their conventions
- I don't imagine anyone using single brackets at all but always double
- if only double brackets are going to be used, I don't see a reason to even implement both versions - I'd just do single-bracket operator and thus simplify the API and avoid typing double-brackets each time `expect_call` is invoked

hadley: `[[` is for extracting a single element, so that's what you should implement.

lbartnik: I changed the API but I almost like the old way better.

lbartnik: @hadley I'll change brackets then. Is the rest OK?

R/mock-object.R
```diff
@@ -0,0 +1,103 @@
...
+  )
+
+  calls[[i]]
+}
+
+
+#' Expectation: does the given call match the expected?
+#'
+#' @inheritParams expect_that
+#' @param mocked_call A mocked call, obtained from a \code{\link{mock}} object.
```
krlmlr: Could be also any other kind of call now.

R/mock-object.R
```diff
@@ -0,0 +1,103 @@
...
+#' @param expected_call Expected call.
+#' @family expectations
+#' @export
+#' @examples
```
krlmlr: `expect_call(quote(summary(iris)), summary(iris))` -- would that work?

krlmlr:
If the tests fail, coverage won't be updated: https://travis-ci.org/hadley/testthat/builds/130754313#L625

R/mock-object.R
lbartnik: Actually, it will. I added a new example.

R/mock-object.R
lbartnik: True, I say "preferably" now.

lbartnik:
I've just fixed that but I'm almost sure there were no such errors in the package before and yet the coverage check failed. We'll see in a few minutes.

krlmlr:
LGTM. No idea why coverage is reported oddly on codecov -- perhaps it's because a new file is added with this PR?

lbartnik:
A new file and a new test file.

lbartnik:
@krlmlr The interesting part is that when I go to "codecov/project" I get this message "404
No report found to at 495b644 (head) to compare against."

jimhester:
There seems to be an issue in general with testthat tests on travis.

Running `covr::package_coverage(line_exclusions = c("inst/include/testthat/testthat.h", "inst/include/testthat/vendor/catch.h"))` locally yields an overall coverage of 81.74%, but on travis/codecov the coverage is only [45.29%](https://codecov.io/github/hadley/testthat?branch=master). I am guessing some of the tests are not being run, not sure why that is the case though.

krlmlr:
If covr showed the payload that it sends to codecov.io, we'd know a bit more ;-)

jimhester:
Not really, there are no transformations being done, save the coverage and pass it along if you want to inspect things.

```r
Rscript -e '(cov <- covr::package_coverage(line_exclusions = c("inst/include/testthat/testthat.h", "inst/include/testthat/vendor/catch.h")));covr::codecov(cov = cov)'
```

jimhester:
I am able to reproduce this, seems to be due to the methods package not being loaded when using RScript. The following fixes the issue.

```r
Rscript --default-packages='datasets,utils,grDevices,graphics,stats,methods' -e 'covr::codecov(line_exclusions=c("inst/include/testthat/vendor/catch.h", "inst/include/testthat/testthat.h"))'
```

kevinushey:
Life is probably better if we pretend that `Rscript` doesn't exist and just always call `R`. :)

markhocky:
Hi all, just saw this pull request. I've actually been working on something very similar:
https://github.com/markhocky/testthatMockr
I was in the process of submitting to CRAN, never thought to make it a pull request...
I have a slightly different implementation.
Rather than going ahead with a separate package, would I be better off trying to merge any extra features I have (if there are any) into this branch?

lbartnik:
I think it won't hurt to see what could be merged.

markhocky:
Hi Lukasz, I'll take a more in depth look at your work and see if there's
anything I can add.
Mark.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+
+  class(mock_impl) <- 'mock'
+  mock_impl
+}
+
+
+is_mock <- function (object) inherits(object, 'mock')
+
+
+#' \code{$.mock} returns one of two lists: \code{calls} or \code{args}.
```
hadley: These methods do not need to be documentation indvidually

lbartnik:
I'm not sure which methods you have in mind. Could you please name them? the `$.mock` and `length.mock` have `rdname` set to `mock`. Is there something else you say is incorrect?

R/mock-object.R
hadley: Sorry, I missed the `@rdname mock`.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+#'   expect_equal(summary(iris), 1)
+#'   expect_length(m, 1)
+#'   expect_call(m, 1, summary(iris))
+#'   expect_args(m, 1, iris)
+#' })
+#'
+mock <- function (..., cycle = FALSE, envir = parent.frame()) {
+  stopifnot(is.environment(envir))
+
+  return_values     <- eval(substitute(alist(...)))
```
hadley: Why not `list(...)`?

hadley: Oh because you're capturing them unevaluated, assuming that they're coming from `envir`. That's not a great assumption because each element of `...` could actually come from a different environment. I think this will have to wait until the next release where testthat will depend on lazyeval, and you can do this correctly.

hadley: Alternatively, you could just evaluate the values once and use `list(...)`.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+  return_values_env <- envir
+  call_no           <- 0
+  calls             <- list()
+  args              <- list()
+
+  mock_impl <- function (...)
```
hadley: `{` needs to be on previous line. No space after `function`

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+#' @rdname mock
+#' @export
+#' @examples
+#' m <- mock()
+#' m(x = 1)
+#' m(y = 2)
+#' expect_equal(length(m), 2)
+#' expect_equal(m$calls[[1]], quote(m(x = 1)))
+#' expect_equal(m$calls[[2]], quote(m(y = 2)))
+`$.mock` <- function (m, n) {
```
hadley: I think overriding `$` in a way that is inconsistent with `length()` and `names()` is dangerous. I'd recommend providing `mock_calls()` and `mock_args()` functions instead.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+  length(environment(x)$calls)
+}
+
+
+#' Expectation: does the given call match the expected?
+#'
+#' @inheritParams expect_that
+#' @param mock_object A \code{\link{mock}} object.
+#' @param n Call number.
+#' @param expected_call Expected call.
```
hadley: You need to make clear that this is unevaluated

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+  expect(
+    0 < n && n <= length(mock_object),
+    sprintf("call number %s not found in mock object", toString(n))
+  )
+
+  expected_call <- substitute(expected_call)
+  mocked_call <- mock_object$calls[[n]]
+
+  expect(
+    mocked_call == expected_call,
```
hadley: `==` is vectorised - you should be using `identical()` to compare calls here.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+            format(expected_call), format(mocked_call))
+  )
+
+  invisible(TRUE)
+}
+
+
+#' Expectation: does the given argument values match expected?
+#'
+#' Unlike \code{\link{expect_call}}, \code{expect_args} compares the
```
hadley: I think you should document these two functions together.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
+
+#' Create a mock object.
```
hadley: Create a mocked function?

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+#'
+#' Mock object's primary use is to record calls that are made on the
+#' mocked function.
+#'
+#' Optionally values/expressions can be passed via \code{...} for the
+#' mock object to return them upon subsequent calls. Expressions are
+#' evaluated in environment \code{envir} before being returned. If no
+#' value is passed in \code{...} then \code{NULL} is returned.
+#'
+#' Passing an expression or a function call via \code{...} is also a
```
hadley: If this is important, it needs to be accompanied by an example.

R/mock-object.R
```diff
@@ -0,0 +1,174 @@
...
+#' way to implement side effects: keep track of the state of code
+#' under testing, throw an exception when a condition is met, etc.
+#'
+#' @param ... Values returned upon subsequent calls.
+#' @param cycle Whether to cycle over the return values. If \code{FALSE},
+#'        will fail if called too many times.
+#' @param envir Where to evaluate the expressions being returned.
+#' @return A mock object.
```
hadley: This is not a useful description

R/mock-object.R
lbartnik: True, but waiting with evaluation until actual call happens let's us implement side effects - e.g. throwing an exception.

R/mock-object.R
lbartnik: coding style wars...

lbartnik:
@hadley I've tried to address all the issues that you reported. Please take a look and let me know if those changes work for you.

n-s-f:
@lbartnik It might be relevant to note that I've done some work on mocking at [mockery](https://github.com/n-s-f/mockery). 

That library is a little like the [patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) function in python's mocking library in that it changes functions (including namespaced functions) in the environment of a calling function, but only for the duration of the test function in which it's used.  

I believe in that sense it's a lot like the (with_mock)[https://github.com/hadley/testthat/blob/a102b89a8f23120ea5e614a1adf229b947e14850/R/mock.R#L1-L63] function. The difference is that mockery does not modify C code, and, I think, gives you nicer syntactic options in that you don't have to assert expectations as arguments to another function.

Probably the 'right' thing to do in some sense is to combine this mock-object idea and that library.  This would replicate the power of python's mocking library, and allow for mocks without dependency injection.

I believe @hadley [suggested](https://twitter.com/hAadleywickham/status/755509418406453248) that it would be better if the mocking code was altogether in a different package, which makes sense to me.  

@lbartnik, perhaps you can pull this out into it's own package, and I'll rework mockery to work well with your code and submit a pull request?

lbartnik:
Hey, I'll look at the code you mentioned this week to see how much work that would be.

lbartnik:
Hi @n-s-f,

I don't think I'll be able to work on that, after all.

Lukasz

krlmlr:
@n-s-f: I have looked into your mockery package, and I like the way stub() selectively replaces a function definition in a precise location, and not everywhere like with_mock() does. Picking where to replace a function definition may be cumbersome, but replacing everywhere has its own problems, for instance, you can't fall back to the original implementation *at all*.

Perhaps you could support several levels of mocking: Selective mocking like stub(), mocking via C like with_mock(), and maybe mocking similar to the way base::trace() -> methods::.TraceWithMethods() injects code.

n-s-f:
@krlmlr Thanks for your comments. I'm not sure I fully understand, but here's my take.

My thinking behind not supporting C mocking (as does with_mock) is that I don't think that in a unit test, it doesn't make sense to mock out a function that is not directly called by the function under test. E.g. if _f()_ calls _g()_ and _g()_ calls _h()_, I'd never want to mock out _h()_ while testing _f()_ - I'd just mock out _g()_ (and perhaps in _g()_'s unit tests I'd mock out _h()_).

Is there a case where you think that approach is insufficient?

I not very familiar with `base::trace`, and on a brief look, I'm having a hard time figuring out exactly how `methods::TraceWithMethods` works (it's a rather involved function). What do you see as the advantage of stubbing out functions in this way as well?

I'm afraid I've somewhat derailed the conversation about the original pull request, so I may open a different issue on this topic. 

Thanks

lbartnik:
@krlmlr , @n-s-f I decided to move mocking code into a separate package, I'm also going to go through the submission process soon. Maybe we could work together for a bit and make its functionality more mature?
Here's the repo: https://github.com/lbartnik/mock

n-s-f:
@lbartnik https://twitter.com/lukaszbartnik/status/783341623824437249, you can reach me at noam.finkelstein@jhu.edu

lbartnik:
Hi, this functionality has been turned into a new package (https://cran.r-project.org/web/packages/mockery/index.html) so I think we can close this PR.

PR #474, opened 2016-05-17
Report endless recursion

krlmlr:
Currently only contains a failing test.

Expected behavior: An error should be reported.


krlmlr:
@hadley: The reason here is that handle_error() can't execute properly because the stack is already "almost full". 2d8429d399f2 demonstrates the principle by temporarily raising the limit, but I'll be looking for a better solution.

krlmlr:
This now builds upon #494 and #495, and includes a better refactoring of the expectation code. Infinite recursions are now caught with a proper stack trace.

krlmlr:
The output is way too large: 5b17d3ed0. Need to limit to first/last 20 entries?

hadley:
@krlmlr I think keeping the first and last 10 would be great if >20 entries

krlmlr:
Should be much better now. Setting options(expression = 200) for the output tests.

hadley:
LGTM.

krlmlr:
Thanks. Merging because it affects #470.

PR #493, opened 2016-07-03
Fix #492: only remove trailing [rR] from filename

### LINKED ISSUE #492: find_test_scripts should only strip trailing [rR] and not any [rR]
This [line](https://github.com/hadley/testthat/blob/master/R/test-files.R#L76):

```
test_names <- gsub("\\.[rR]", "", test_names)
```

Should be (I think):

```
test_names <- sub("\\.[rR]$", "", test_names)
```

I'll submit a pull request in a few minutes to fix.

brodieG:
Note I changed both `gsubs` to `sub` although not technically necessary, seems more correct since both tokens being replaced should only occur once per element in the `test_names` vector due to the `^` and `$`.  Also, fwiw the file name that was causing me issues was of the form "test-blah.rblah.R" where the first `.r` was getting dropped and then a filter searching for it with `rblah` wasn't finding it.


hadley:
Would you mind adding a unit test?

brodieG:
Not at all; just give me five.

brodieG:
Ok, done, let me know if you would prefer them in their own file instead of in the `test_dir` file.  Sorry for not including them in the first place.  I have to run now so can't check that build succeeds, but R CMD check worked and I'll check in later to make sure the builds succeed.

tests/testthat/test-test_dir.R
```diff
@@ -45,3 +45,23 @@ test_that('test_dir() helpers', {
   df <- as.data.frame(res)
   expect_true(all(!df$error & df$failed == 0))
 })
+
+test_that('find_test_scripts() with tricky names', {
```
hadley: Rather than mocking, I'd rather that you extract out the pure code from `find_test_scripts()` - maybe call it `filter_test_scripts()`?

brodieG: I'm no sure I understand.  I mocked `dir` so that I did not have to create a directory structure with the problematic file names.  Do you want me to create a new function that accepts the output of `dir` and does the filtering, and then modify `find_test_scripts` to use that function instead?

hadley: Yes. I don't like mocks in general.

brodieG:
Makes sense.  I'll address that tomorrow.

brodieG:
Hmm, I messed something up with the docs.  I ran out of time now, so I'll fix later this afternoon.

brodieG:
fwiw, this should be good to go now.

R/test-files.R
```diff
@@ -59,21 +59,25 @@ test_files <- function(paths, reporter = "summary",
...
 #' @param filter cf \code{\link{test_dir}}
 #' @param invert If \sQuote{TRUE} return files which do \emph{not} match.
 #' @param ... Additional arguments passed to \code{grepl} to control filtering.
 #' @return the test file paths
 #' @keywords internal
 #' @export
-find_test_scripts <- function(path, filter = NULL, invert = FALSE, ...) {
-  files <- dir(path, "^test.*\\.[rR]$", full.names = TRUE)
+
+filter_test_scripts <- function(files, filter = NULL, invert = FALSE, ...) {
```
hadley: This doesn't need to be exported or documented - it's just an internal helper.

brodieG: I'm just following the pattern for `find_test_scripts` of documenting with `@keywords internal` and exporting.  Do you want me to remove docs for both the original and the new function, or just for the new function.

hadley: Just leave all the documentation exactly as it was

brodieG:
Okay, hopefully fixed to your liking.  I restored the docs, although due to how the diff matches it looks like they are changed.  I restored them with `devtools::document`, so the `.Rd` files should reflect the `roxygen` in the file I modified.

hadley:
Can you please add a bullet to NEWS, and then I'll merge 😄

brodieG:
NEWS updated.

hadley:
Thanks!

PR #497, opened 2016-07-04
Increase options(expressions=) by 500

krlmlr:
when handling errors. Also, try to register errors in top-level handler first, revert to fatal handler only if error cannot be handled.

This is still ugly, but necessary for the debug reporter (#470).


hadley:
Thanks!

PR #499, opened 2016-07-04
New argument minimum_version to skip_if_not_installed()

### LINKED ISSUE #487: FR: New argument "version" to skip_if_not_installed()
for tests that differ depending on which version of a dependent package is installed.

krlmlr:
The package isn't loaded anymore as a side effect, and the check also works if the package is loaded already. With test.

Closes #487.


R/skip.R
```diff
@@ -44,13 +44,19 @@ skip_if_not <- function(condition, message = deparse(substitute(condition))) {
 
 #' @export
 #' @param pkg Name of package to check for
+#' @param minimum_version Minimum required version for the package
 #' @rdname skip
-skip_if_not_installed <- function(pkg) {
-  if (requireNamespace(pkg, quietly = TRUE)) {
-    return(invisible(TRUE))
+skip_if_not_installed <- function(pkg, minimum_version = NULL) {
+  tryCatch(
```
hadley: I'd rather keep the existing check - this feels a bit too clever to me.

krlmlr: I don't like the existing check for its side effect, it causes the package to be loaded. Furthermore, requireNamespace() doesn't check the version once the package is loaded. I could move the skip() call outside the tryCatch() if that helps

krlmlr: Or I could call requireNamespace() in addition to the manual version check -- this would work even if the installed package is somehow damaged so that DESCRIPTION exists but the package cannot be loaded.

hadley: Loading doesn't seem like a bad side effect to me - you're assuming checking it's present before you attempt to use it to do something.

krlmlr:
Reinstalled the old check, but still need to test version number manually.

krlmlr:
Now requiring package first, then checking version if appropriate.

hadley:
LGTM. I'll merge when I'm next at my computer.

Issue #486, opened 2016-06-21
FR: Run tests in parallel


krlmlr:
to speed up TDD.
- One task per context
- Need special care to avoid mixing output
  - Can be achieved by making reporter$cat() collect all output and print it in reporter$.end_context()
- Can use the "parallel" package for simplicity
  - Some backends need to load the package and the helpers on each node

russellpierce:
Just a thought: Being able to opt out for some tests might be good.(unless testthat is only intended for use with unit tests and excludes integration tests)

jeffpollock9:
Hi, what needs to be done to get this (or similar) accepted? I think it would be a great addition.

PR #501, opened 2016-07-07
Sketch for parallel testing

krlmlr:
in preparation for parallel processing (#486).

This is still very rough, but from here on adding parallel support is almost trivial. We could use e.g. the new `future` package.


krlmlr:
With multicore backend: dplyr test times drop to ~6 seconds (sequential: ~10 seconds).

R/test-files.R
```diff
@@ -42,22 +42,36 @@ test_files <- function(paths, reporter = "summary",
     stop('No matching test file in dir')
   }
 
+  # Make sure reporter can be found before running the tests
   current_reporter <- find_reporter(reporter)
+
+  # Run without reporter first, will collect results anyway
+  if (interactive()) {
```
krlmlr: Working directory seems to be set incorrectly in parallel workers.

hadley:
I'm not going to review for the next release since I need to get out something fairly quickly for the changes in R

hadley:
I'm going to close this because I think the implementation will look significantly different when we use one of the higher level constructs currently under development.

PR #505, opened 2016-07-08
New load_helpers arg to test_dir()

krlmlr:
So that devtools doesn't need to load the helpers twice when running the tests.


krlmlr:
Merging now to support https://github.com/hadley/devtools/issues/1256, CC @jimhester.

PR #522, opened 2016-09-03
Escape regular expression characters in error output.

jimhester:
This should make it more clear why something that looks identical did
not match.

e.g. expect_match("foo() is not bar", "foo() is not bar")


hadley:
This is good to merge. I'll merge when I'm next at my computer so I can resolve the news conflict

hadley:
Closed by 1483f356d2cd8281719be00142f69ff990a20994. Not sure why it didn't automatically close

PR #524, opened 2016-09-06
Avoid using find_label

kalibera:
This patch replaces the use of find_label in testthat by substitute. The problem of find_label is that it depends on that function arguments are always promised (e.g. even when they are constants, always re-promised, etc). This is an internal behavior of R that is not safe to depend on, this behavior can change (mostly for performance reasons), and it already changed in the byte-code compiler where constants are not wrapped into promises. So this patch is needed for testthat to work with the byte-code compiler. It may be worth removing find_label from the code completely (I already commented it out for now). There is also some maintenance advantage of this patch in avoiding native code.

### DIFF
--- a/R/expectation.R
+++ b/R/expectation.R
@@ -53,8 +53,6 @@ add_info <- function(message, info = NULL) {
 }
 
 label <- function(x) {
-  x <- find_label(x)
-
   if (is.character(x)) {
     encodeString(x, quote = '"')
   } else if (is.atomic(x)) {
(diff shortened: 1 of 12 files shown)


R/expectation.R
```diff
@@ -70,10 +68,14 @@ label <- function(x) {
   }
 }
 
-#' @useDynLib testthat find_label_
-find_label <- function(x) {
-  .Call(find_label_, quote(x), environment())
-}
+# This function should not be used as it depends on that all function
```
hadley: Can you please just delete this and the related C code?

tests/testthat/test-label.R
```diff
@@ -8,21 +8,21 @@ test_that("labelling compound {} expression gives single string", {
...
-})
-
-test_that("can find label even when deeply nested", {
-  f <- function(x) g(x)
-  g <- function(y) h(y)
-  h <- function(z) find_label(z)
-
-  expect_equal(f(abc), quote(abc))
-})
+#test_that("can find label after it's been forced", {
```
hadley: And you can delete this too.

PR #532, opened 2016-10-06
Refactor the test_examples function

jimhester:
Previously it would never run examples during R CMD check because there
was no '/man' directory for installed packages. This instead uses
tools::Rd_db to parse the Rd files into a database and uses that
directly, which does work for installed packages.

Fixes jimhester/covr#205


R/test-example.R
```diff
@@ -4,13 +4,26 @@
 #' Each example counts as one test, and it succeeds if the code runs without
 #' an error.
 #'
-#' @param path For \code{test_examples}, path to directory containing Rd files.
+#' @param path For \code{test_examples}, path to the package directory.
 #'   For \code{test_example}, path to a single Rd file. Remember the working
 #'   directory for tests is \code{tests/testthat}.
+#' @param rd A parsed Rd object, obtained from \code{\link[tools]{Rd_db}}.
```
hadley: "or otherwise"

R/test-example.R
```diff
@@ -4,13 +4,26 @@
...
 #' @export
-test_examples <- function(path = "../../man") {
-  man <- dir(path, "\\.Rd$", full.names = TRUE)
-  lapply(man, test_example)
+test_examples <- function(path = "../..") {
```
hadley: I _think_ this would be clearer with two functions - maybe `test_examples_installed` and `test_examples_source` ?

jimhester: Ok I split it up, I agree the logic is more clear now

jimhester:
I haven't come up with a good way to test this, since we don't have a source package when using `R CMD check` there isn't a great way to test running source examples. The current test runs the the code to find the examples, but does not actually run them. The main issue with running them is the examples contain expectations themselves, which makes the reporting a little confusing. They also have additional output that would need to be suppressed using something like `withr::with_output_sink()`.

hadley:
Thanks!

Issue #513, opened 2016-08-06
`format.mismatch_character` produces warning  and erroneous output when there are many differences


brodieG:
An example:

```
> library(testthat)
Warning message:
package ‘testthat’ was built under R version 3.2.5 
> expect_equal(letters[1:6], LETTERS[1:6])
Error: letters[1:6] not equal to LETTERS[1:6].
6/6 mismatches
x[1]: "a"
y[1]: "A"

x[2]: "b"
y[2]: "B"

x[3]: "c"
y[3]: "C"

x[4]: "d"
y[4]: "D"

x[5]: "e"
y[5]: "E"

x[6]: "a"
y[6]: "A"
In addition: Warning messages:
1: In mapply(FUN = f, ..., SIMPLIFY = FALSE) :
  longer argument not a multiple of length of shorter
2: In mapply(FUN = f, ..., SIMPLIFY = FALSE) :
  longer argument not a multiple of length of shorter
```

Notice that in addition to the warnings, the first 5 elements of the two vectors are being recycled (see `x[6]` and `y[6]`).

The problem is in [`format.mismatch_character`](https://github.com/hadley/testthat/blob/a102b89a8f23120ea5e614a1adf229b947e14850/R/compare-character.R#L83), where anytime there are more differences than `max_diffs` (currently defaults to 5), `show_x`, `show_y`, and `x$i` are no longer the same lengths in the [`Map` call](https://github.com/hadley/testthat/blob/a102b89a8f23120ea5e614a1adf229b947e14850/R/compare-character.R#L93).

Let me know if you want me to submit a PR to fix this, assuming you agree with what I laid out above.

PR #547, opened 2016-11-21
Issue513: compare Problems with > 5 Character Mismatches 

brodieG:
Fix issue #513

### DIFF
--- a/tests/testthat/test-compare-character.R
+++ b/tests/testthat/test-compare-character.R
@@ -71,3 +71,11 @@ test_that("not all lines are shown", {
   expect_equal(lines[1], "1/1 mismatches")
   expect_equal(length(lines), 8)
 })
+
+test_that("longer vectors (#513)", {
+  expect_warning(comp <- compare(letters[1:6], LETTERS[1:6]), NA)
+  expect_equal(
+    comp,
+    structure(list(equal = FALSE, message = "6/6 mismatches\nx[1]: \"a\"\ny[1]: \"A\"\n\nx[2]: \"b\"\ny[2]: \"B\"\n\nx[3]: \"c\"\ny[3]: \"C\"\n\nx[4]: \"d\"\ny[4]: \"D\"\n\nx[5]: \"e\"\ny[5]: \"E\""), class = "comparison")
+  )
+})
--- a/tests/testthat/test-compare-character.R
+++ b/tests/testthat/test-compare-character.R
@@ -72,10 +72,9 @@ test_that("not all lines are shown", {
   expect_equal(length(lines), 8)
 })
 
-test_that("longer vectors (#513)", {
-  expect_warning(comp <- compare(letters[1:6], LETTERS[1:6]), NA)
-  expect_equal(
-    comp,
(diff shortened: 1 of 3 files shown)


hadley:
Can you please change the title of the pull to something informative?

tests/testthat/test-compare-character.R
```diff
@@ -71,3 +71,11 @@ test_that("not all lines are shown", {
   expect_equal(lines[1], "1/1 mismatches")
   expect_equal(length(lines), 8)
 })
+
+test_that("longer vectors (#513)", {
+  expect_warning(comp <- compare(letters[1:6], LETTERS[1:6]), NA)
```
hadley: I don't understand why you're checking for the absence of a warning here.

tests/testthat/test-compare-character.R
```diff
@@ -71,3 +71,11 @@ test_that("not all lines are shown", {
...
+  expect_equal(
+    comp,
+    structure(list(equal = FALSE, message = "6/6 mismatches\nx[1]: \"a\"\ny[1]: \"A\"\n\nx[2]: \"b\"\ny[2]: \"B\"\n\nx[3]: \"c\"\ny[3]: \"C\"\n\nx[4]: \"d\"\ny[4]: \"D\"\n\nx[5]: \"e\"\ny[5]: \"E\""), class = "comparison")
```
hadley: This test hard to understand. It would be clearer if you tested the individual components, splitting the `message` into lines and comparing only the most important.

hadley:
This also needs a bullet point in `NEWS.md`

tests/testthat/test-compare-character.R
brodieG: The code this PR fixes used to produce a warning when there are more than 5 mismatches.  I guess this is not strictly needed going forward assuming no one re-introduces the same error, but thought it might make sense to try to avoid an accidental regression.

Please confirm that you want me to remove it.

tests/testthat/test-compare-character.R
brodieG: Ok, will do.

brodieG:
Re NEWS, I think that's done already [NEWS](https://github.com/hadley/testthat/pull/547/files#diff-8312ad0561ef661716b48d09478362f3)

tests/testthat/test-compare-character.R
hadley: Yes, please remove. Warnings are captured anyway by the reporter

tests/testthat/test-compare-character.R
brodieG: Upon further reflection, I'm struggling on how to do this in a more meaningful way.  Certainly I admit the test is not super clear, but part of the problem is that the original failure was because too much stuff was generated (incorrect recycling shown in blue under "new"):

screen shot 2016-12-16 at 4 29 55 pm

I can write something that counts number of lines and spot checks a couple of the lines and should be easier to grasp at a glance, but this seems incomplete.  Would you be satisfied if I instead tested each component separately, but used the totality of the output of:

```
> compare(letters[1:2], LETTERS[1:2], max_diffs=1)$message
[1] "2/2 mismatches\nx[1]: \"a\"\ny[1]: \"A\""
```

hadley: In that case, just use `expect_output_file()`

brodieG: I ended up just testing each component equally, only because I do not know where to put the reference file for `expect_output_file`.  I didn't see any uses of `expect_output_file` in the tests that weren't just tempfiles.  If you don't like what I just committed just tell me where you want me to save the file for use with `expect_output_file`.

hadley:
Thanks!

PR #518, opened 2016-08-30
Add encoding support

hansharhoff:
This pull request adds support for an encoding argument to test_dir (and the functions called by this). This is related to an [issue reported in devtools](https://github.com/hadley/devtools/issues/1306). A separate pull request is incoming for devtools.

PR #550, opened 2016-12-07
Add encoding support

krlmlr:
Now with proper recoding of file read by `readLines()` call. @hadley: PTAL.

Closes #518 (=includes it).

CC @hansharhoff.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # testthat 1.0.2.9000
 
-* New argument`encoding` in `test_file()` and `source_file()` (@hansharhoff, hadley/devtools#1306)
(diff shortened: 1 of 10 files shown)


hadley:
Do you think you could come up with a simple test too?

krlmlr:
@hadley: The simplest test I can think of is to create a dummy package in `tests/testthat`, like `devtools` does. Where should that test live anyway?

hadley:
Maybe test one level up? (i.e. `test_file()`)

krlmlr:
@hadley: Added simple test for Latin-1 and UTF-8.

hadley:
Thanks!

Issue #527, opened 2016-09-15
testthat_results is empty with test_dir


billdenney:
When I am testing the current GitHub version of the PKNCA package, test_dir gives an empty return value while when I test a specific file with test_file, I get the expected return value.

```
testresult <- test_dir("tests")
testthat results ===============================================================================================================================================
OK: 855 SKIPPED: 0 FAILED: 0

DONE ==========================================================================================================================================================
> length(testresult)
[1] 0
> testresult <- test_file("tests/testthat/test-auc.R")
AUC and AUMC: .....................................................................................................................

DONE ==========================================================================================================================================================
> length(testresult)
[1] 9
> 
```

Also, generally, examples of how to use Reporters and testthat_results would help.  Perhaps a vignette could be generated?

hadley:
What are you trying to do? Why are you using `test_dir()` in the first place?

billdenney:
By your message, I probably should be using a different function than `test_dir()`.  Can you clarify what function I should be using (and preferably update the documentation to prevent others' similar confusion)?

I'm trying to "Test files start with test" (from the test_dir documentation) in my package, and I'd then like to generate a report with the information on the tests running successfully.  In the report, I'm hoping to have at least the context, the test_that description, and the expect_* info arguments along with pass or fail status.

It is unclear to me how to use the return value of `test_dir()`.

hadley:
Is there some reason you're not using `devtools::test()`?

billdenney:
It doesn't look like it has a return value (at least nothing is noted in its documentation, because nothing was noted in the documentation I didn't start there).  As described above, I'd like more detail than is typically indicated by X tests passed, Y tests failed.

hadley:
I'm trying to understand what your end goal is here - why are you trying to run tests in a way that's different to everyone else?

billdenney:
I'm trying to get a detailed report from the tests with more detail than "OK: 855 SKIPPED: 0 FAILED: 0".

It looks like I can get this from `test_file`, but I do not get this from `test_dir`.  The documentation for both functions say that they return "the results as a "testthat_results" (list)".  In my experience above, `test_dir` returns an empty result while `test_file` returns something like what I'm wanting.  The empty return value from `test_dir` appears to be some combination of documentation error or bug.

My thought is that a data.frame with relevant columns describing the test (e.g. "filename", "context", "info", "result") would cover my needs.  I think I can get this from `test_file` but not currently from `test_dir`.

[...]

PR #556, opened 2016-12-16
Correct documentation of test_dir return value (fixes hadley/testthat#527)

billdenney:
Pull request to correct the documentation as discussed in #527


hadley:
Thanks!

Issue #528, opened 2016-09-15
expect_equivalent() for categorical data


AmeliaMN:
I know you are using base functions like `all.equal()` to do the underlying checks, but it would be nice to have a way to check levels in factor variables without getting the ordering precisely right. See https://github.com/dsscollection/factor-mgmt/blob/master/analysis/working_with_factors.Rnw#L394 for an example of what I'm talking about. In essence, I wish that I could run

```
expect_equivalent(c("Male", "Female"), c("Female", "Male")) 
```

and get a TRUE. 
Maybe this is something more appropriate for https://github.com/hadley/forcats ?

cc @nicholasjhorton

nicholasjhorton:
I would think that:

expect_equivalent(c("Male", "Female"), c("Female", "Male")) 

should return TRUE but

expect_equal(c("Male", "Female"), c("Female", "Male")) 

should return FALSE


> On Sep 15, 2016, at 2:45 PM, Amelia McNamara <notifications@github.com> wrote:
> 
> I know you are using base functions like all.equal() to do the underlying checks, but it would be nice to have a way to check levels in factor variables without getting the ordering precisely right. See https://github.com/dsscollection/factor-mgmt/blob/master/analysis/working_with_factors.Rnw#L394 for an example of what I'm talking about. In essence, I wish that I could run
> 
> expect_equivalent(c("Male", "Female"), c("Female", "Male")) 
> 
> and get a TRUE. 
> Maybe this is something more appropriate for https://github.com/hadley/forcats ?
> 
> cc @nicholasjhorton
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Nicholas Horton 
Professor of Statistics
Department of Mathematics and Statistics, Amherst College
PO Box 5000, AC #2239
Amherst, MA 01002-5000

hadley:
Maybe `expect_setequal()` that would use `setequal()` under the hood?

I'm unlikely to implement this myself, but I'd be happy to review a PR.

PR #557, opened 2016-12-16
expect_setequal

carlganz:
Apropos of #528 

This is an attempt to implement `expect_setequal`. This version appears to work correctly, but I suspect there is still plenty of room for improvement so all feedback is appreciated. 

I tried to mimic the code for the other `expect_*` functions, but I had to stray a little bit to make it work. 

Kind Regards,
Carl


R/expect-equality.R
```diff
@@ -105,3 +105,21 @@ expect_identical <- function(object, expected, info = NULL, label = NULL,
...
 
+#' @export
+#' @rdname equality-expectations
+expect_setequal <- function(object, expected, info = NULL, label = NULL,
+                            expected.label = NULL) {
+
+  lab_act <- make_label(object, label)
+  lab_exp <- make_label(expected, expected.label)
+
+  steq <- compare_set(object, expected)
```
hadley: I think you could considerably simplify the implementation by just comparing sorted unique `object` and `expected`

PR #564, opened 2017-01-24
Extend expect_length to use an S4 object's length method, if exists

nealrichardson:
`expect_length` is a clean, convenient shortcut for `expect_equal(length(x), n)`, but in its current implementation, only objects that are one of R's internal array types are whitelisted. An S4 object that doesn't inherit from one of those types is forbidden, even if it has a `length` method defined. Aside from the inconvenience of having to fall back to `expect_equal`, this limitation increases the chances of writing a misleading test because, as `?length` says, "[a]ll other objects (including functions) have length one." 

This pull request extends `expect_length` to work with S4 objects that have a length method. The additional check examines whether `selectMethod` selects the default `.Primitive("length")` (which always returns 1) for an S4 object. If a length method other than the primitive is set or inherited, `expect_length` will proceed. Other seemingly more natural ways of checking for the existence of a method don't work: `existsMethod` doesn't follow inheritance and `hasMethod` always returns `TRUE` for S4 objects because the primitive method is found.

This pull request also contains a [commit](https://github.com/nealrichardson/testthat/commit/ebbb5208b73ea2d9a5000cb8ff0d7c3796daa881) that updates the junit test fixture for what I am guessing is a change in the 'xml2' package. When I reset my fork to upstream/master, it failed on Travis. I suspect an 'xml2' change caused the failure because the most recent release of that package was after the most recent commit to the master branch of 'testthat'.


R/expect-length.R
```diff
@@ -15,8 +15,8 @@ expect_length <- function(object, n) {
   stopifnot(is.numeric(n), length(n) == 1)
   lab <- label(object)
 
-  if (!is_vector(object)) {
-    fail(sprintf("%s is not a vector.", lab))
+  if (!has_length(object)) {
```
hadley: I'd rather simply remove this check all together. Would you mind doing that?

nealrichardson:
Check removed.

tests/testthat/test-expect-length.R
```diff
@@ -1,12 +1,31 @@
...
+  expect_success(expect_length(letters[1:5], 5))
+})
+
+test_that("uses S4 length method, if exists", {
+  # A has no length method defined
+  A <- setClass("ExpectLengthA", slots=c(x="numeric", y="numeric"))
+  # Default for S4 objects that don't inherit a length method: always length 1
+  expect_success(expect_length(A(x=1:5, y=3), 1))
+
+  # B does has a length method defined
```
hadley: This mostly tests of the implementation of `length()`. I think you only need the test for `ExpectLengthB`

nealrichardson: Tests pruned.

hadley:
Thanks!

nealrichardson:
No, thank you!

PR #566, opened 2017-01-27
Location reporter tweaks

krlmlr:
- The reporter now prints only the base name of the file
- Entry and exit of each test is also reported

https://github.com/hadley/testthat/issues/551#issuecomment-275613863


R/reporter-location.R
```diff
@@ -11,16 +11,24 @@ NULL
...
+        location <- paste0(basename(attr(ref, "srcfile")$filename), "#", ref[1], ":1")
       }
 
       status <- expectation_type(result)
-      self$cat_line(location, " [", status, "] ", test)
+      self$cat_line("  ", location, " [", status, "]")
+    },
+
+    end_test = function(context, test) {
+      self$cat_line("End test: ", test)
```
hadley: Maybe add a new line here? Seems like it would make the output slightly easier to read

PR #568, opened 2017-02-01
Added special compare for the boot object

wligtenberg:
Boot objects gave issues when running covr as mentioned here:
https://github.com/jimhester/covr/issues/244

This is not an issue that is easy to fix in covr, so I made sure that we can easily create a test using testthat that does work when running covr. By default it would behave the same as before, however, with an easy switch we can make sure that it passes when using covr. 

I think it is best to integrate here, however, if you disagree, I can release it separately.


wligtenberg:
I see issues in the build systems because of the file missing from collate.
I didn't want to commit all changes after running document(), because that would just update a lot of .Rd files.

hadley:
This doesn't feel like the right fix to me. I think it'd be easier if you just compared individual parts of the boot objects rather than the whole thing.

PR #569, opened 2017-02-05
accommodate packages without dynamic registration

kevinushey:
This PR helps accommodate packages which don't enable dynamic symbol lookup and want to use Catch. Note that development versions of R are going to encourage this, and CRAN will soon be warning about packages that don't register their routines.

Now, `testthat` will look for a symbol called `C_run_testthat_tests`, or `run_testthat_tests`, and use that as an entrypoint to calling Catch when available.

The noisiness in other `.Rd` files is just whitespace changes.


kevinushey:
Test failures appear to be unrelated.

hadley:
Do we still need this? Or does the approach in your other PR supersede it?

kevinushey:
The other PR should supercede this.

samssann:
This issue still persists. I think there is a typo:

the line
https://github.com/r-lib/testthat/blob/bba270b21381d2fce590cbf03b8e5fc3c10fbf51/R/test-compiled-code.R#L248

should be 
`return(resolved)`

kevinushey:
Ack, thanks for noticing that. I'll fix it up.

kevinushey:
https://github.com/r-lib/testthat/commit/04e4e2f8a0015b39f68bf2aef59e8b3fe6e508b2 should fix it up -- please file a separate issue if you encounter any more problems!

Issue #575, opened 2017-02-16
junit output is missing name attribute in <testcase> element


akbertram:
Currently the junit reporter writes a <testcase> element for each expectation, with a classname attribute but no name attribute, leading to broken links in reporting interfaces like Jenkins:
 
![jenkins](https://cloud.githubusercontent.com/assets/454339/23045243/33f300f2-f4a4-11e6-94d7-98b108f475fb.png)

PR #576, opened 2017-02-16
Adds name attribute to <testcase> elements in junit output (#575)

akbertram:
Improved output: 
http://build.renjin.org/job/Replacement-Packages/job/testthat/11/testReport/

### DIFF
--- a/R/reporter-junit.R
+++ b/R/reporter-junit.R
@@ -89,7 +89,8 @@ JunitReporter <- R6::R6Class("JunitReporter", inherit = Reporter,
       name <- test %||% "(unnamed)"
       testcase <- xml2::xml_add_child(self$suite, "testcase",
        time = toString(time),
-       classname = paste0(classnameOK(context), '.', classnameOK(name))
+       classname = classnameOK(context),
+       name = classnameOK(name)
       )
 
       # message - if failure or error
--- a/R/reporter-junit.R
+++ b/R/reporter-junit.R
@@ -88,9 +88,9 @@ JunitReporter <- R6::R6Class("JunitReporter", inherit = Reporter,
       # XML node for test case
       name <- test %||% "(unnamed)"
       testcase <- xml2::xml_add_child(self$suite, "testcase",
-       time = toString(time),
-       classname = classnameOK(context),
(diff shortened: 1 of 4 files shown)


akbertram:
Fixes Issue #575

akbertram:
@hadley I believe I've fixed the test failures but it looks like neither AppVeyor nor Travis are rebuilding.

R/reporter-junit.R
```diff
@@ -89,7 +89,8 @@ JunitReporter <- R6::R6Class("JunitReporter", inherit = Reporter,
       name <- test %||% "(unnamed)"
       testcase <- xml2::xml_add_child(self$suite, "testcase",
        time = toString(time),
-       classname = paste0(classnameOK(context), '.', classnameOK(name))
```
hadley: Would you mind fixing the indenting while you're in here?

akbertram:
@hadley When I run `test_package("testthat")` all tests pass. But if I run `devtools::check()` . This seems to be because the junit reporter includes stack depth in the error message, which can vary depending on how the test is invoked:

```.xml
 <failure type="failure" message="expectation_failure in structure(list(fail()), start_frame = 19): Failure has been forced&#10;&#10; @reporters/tests.R#8"/>
```

I would suggest removing the `start_frame` attribute from the output anyway, it makes the deparsed function call harder to read?

hadley:
Yeah, that looks like a mistake anyway - "structure(list(fail()), start_frame = 19)" looks like a deparsed object, not something deliberately included. Can you remove it?

hadley:
I can't really follow the logic of the code that currently generates the message, and it's not clear to me what it should be, so I trust your judgement as to what would be more helpful.

akbertram:
Phew. That seems to work now. I've added one more commit that allows CI systems to override the reporter used for test_check() and specify and output location for the junit reporter.

hadley:
Looks good. Now just needs a bullet in NEWS.md, and I can merge.

hadley:
Thanks!

antonytanj:
What is the process to publish a new version of this package with these changes?

PR #581, opened 2017-03-23
Add view.on.update parameter for expect_equal_to_reference

chambm:
It would be helpful if `expect_equal_to_reference` could allow test runners to view the new test results when they are being [re]created. It defaults to off, but it allows test authors to choose to force test runners to view the data. This probably needs some kind of extra checking so it doesn't run when testing is being done in a headless/batch mode, but I'm not sure what the right check to use would be.


chambm:
OK, apparently View() shouldn't be used in "examples etc." but maybe there's another way to do what I'm talking about? Something more convenient than doing this every time I use `expect_equal_to_reference`:
```
some_obj = some_function()
if (!file.exists("some_obj.rds") { View(some_obj) }
expect_equal_to_reference(some_obj, "some_obj.rds")
```

This is much nicer (for brief invocations of a function):
```
expect_equal_to_reference(some_function(), "some_obj.rds", view.on.update=TRUE)
```

hadley:
I don't love the current workflow either, but this doesn't feel like the right fix to me.

PR #585, opened 2017-04-09
clarify `skip` semantics

brodieG:
I couldn't quite tell precisely what `skip` did from documentation alone.  My initial incorrect impression was that it would skip the subsequent test only.  I think it is worth clarifying in the documentation what happens exactly in case others interpret it as I did.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -100,6 +100,8 @@
 
 * New JUnit reporter `JunitReporter`. (#481, @lbartnik)
 
+* Clarified `skip` semantics in documentation (@brodieG)
+
 # testthat 1.0.2
 
 * Ensure `std::logic_error()` constructed with `std::string()`
--- a/R/skip.R
+++ b/R/skip.R
@@ -41,7 +41,7 @@
 #' ## run with `test_file`, `test_dir`, `test_check`, etc.
 #'
 #' test_that("skip example", {
-#'   expect_equal(1, 1L)     # this expectation runs
+#'   expect_equal(1, 1L)    # this expectation runs
 #'   skip('skip')
 #'   expect_equal(1, 2)     # this one skipped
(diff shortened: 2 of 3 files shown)


brodieG:
Not sure what's up with the errors on the travis check:
```
Last 13 lines of output:
  The specified file is not readable: /tmp/watcher2d0a120e7b77/R/test1.R
  1: watch(c(code_path, test_path), test.removed) at testthat/test-watcher.R:87
  2: dir_state(path, pattern, hash = hash)
  3: vapply(files, safe_digest, character(1))
  4: FUN(X[[i]], ...)
  5: digest::digest(path, file = TRUE)
  6: .errorhandler("The specified file is not readable: ", object, mode = errormode)
  7: stop(txt, obj, call. = FALSE)
  
  testthat results ================================================================
  OK: 442 SKIPPED: 0 FAILED: 1
  1. Error: watcher works correctly (@test-watcher.R#87) 
```
I only changed some docs, and the tests pass on my system.  From looking at the code it looks like the type of error that could happen if for whatever reason the file is removed before the `watch` is engaged, but I'm not sure.

brodieG:
Ok, so looks like the test failure was due to the not completely deterministic nature of the "test-watch.R" test and had nothing to do with my documentation changes.

krlmlr:
We do support Markdown now, btw.

brodieG:
@krlmlr so do you want me to change the docs to use markdown format?

krlmlr:
Yes, please. I think this would improve consistency with the rest of the documentation.

brodieG:
Test failures seem to be the 3.4.0 induced failures as in #590.  Tests passed before my minor doc change here.

hadley:
Thanks!

PR #594, opened 2017-05-26
Make broken expectations catchable by try()

jcheng5:
Currently, broken expectations break rmarkdown document rendering
even if knitr chunk option error=TRUE. It also breaks the evaluate
package's ability to capture errors.

This PR is just a starting point for discussion, if you like it I can add
tests.

### DIFF
--- a/R/expectation.R
+++ b/R/expectation.R
@@ -22,6 +22,8 @@ expectation <- function(type, message, srcref = NULL) {
     class = c(
       paste0("expectation_", type),
       "expectation",
+      # Make broken expectations catchable by try()
+      if (type %in% c("failure", "error")) "error",
       "condition"
     )
   )


hadley:
This seems right to me,

hadley:
Oh I bet this is what breaks knitr too.

hadley:
Hmmmm, this does break a test:

```R
test_that("multiple failures captured even when looking for errors", {
  f <- function() {
    expect_true(FALSE)
    expect_false(TRUE)
  }

  reporter <- with_reporter("silent", {
    test_that("", expect_error(f(), NA))
  })
  expect_equal(length(reporter$expectations()), 3)
})
```

Presumably because `expect_error()` shouldn't terminate.

hadley:
Probably because of this `tryCatch()`:

```R
  error <- tryCatch(
    {
      object
      NULL
    },
    error = function(e) {
      e
    }
  )
```

hadley:
So I here I think a condition of class `exception` (but not error) should resume.  I think this means we need some unholy hybrid of `withCallingHandlers()` and `tryCatch()` or @lionel-'s `rlang::with_handlers()`

hadley:
Same basic problem in #618

PR #597, opened 2017-06-06
Add `wrap` arg to `test_dir` for consistency with `test_file` (#586)

### LINKED ISSUE #586: Adding `wrap` Parameter to `test_dir`, `test_check`, and `run_tests`
@krlmlr, Since 37cc0d04 it is no longer possible to run code with `test_dir` that involves `withCallingHandlers` error handling.  This is because the default source action now sets [`wrap=TRUE`](https://github.com/hadley/testthat/blob/master/R/source.R#L15), which causes all sourced code to be run within [`test_code`](https://github.com/hadley/testthat/blob/master/R/source.R#L31).

Since `wrap` is not exposed in [`test_dir`](https://github.com/hadley/testthat/blob/master/R/source.R#L31) there is no way to disable this behavior when using `test_dir` (or `test_check`).

A simple example of the problem (here we just use `test_file` since `test_file` has the argument so we can show both cases):
```
f <- tempfile()
cat(
  "
  withCallingHandlers(
    stop('This should be handled'),
    error=function(e) cat('handled\n')
  )
  test_that('truth', {expect_true(TRUE)})
  ",
  file=f
)
test_file(f)              # fails
test_file(f, wrap=FALSE)  # works
```
For reference, more details on the running [`withCallingHandlers` inside `try` block issue](http://stackoverflow.com/questions/20572288/capture-arbitrary-conditions-with-withcallinghandlers).

I'm happy to submit a PR that addresses this if you'd like.

brodieG:
I hope it is not too presumptuous of me to submit this PR despite negative feedback in #586.  I am just hoping that because the change:

1. is minimal
2. restores parameter symmetry between `test_file` and `test_dir` broken by 37cc0d0

you will consider it.  Unfortunately for me I'm probably one of the very few people affected by 37cc0d0 since I happen to maintain a package that uses `withCallingHandlers` in `testthat` tests, outside of `test_that` blocks.  If 37cc0d0 gets pushed to CRAN I have no way to workaround this anymore.

A possible alternative to this would be to set the `test_file` `wrap` parameter default value via `getOption`, which would provide another possible work-around.

Sorry for being obnoxious about this, but submitting this as a PR in the hopes you will consider it is a lot easier than figuring out how to take the ~800 tests in my package out of `testthat`. 

Fix #586.


brodieG:
Re test failures, these seem to be the same ones in #594 and #590, so I'll let them be pending additional feedback.  It also looks like Appveyor upgraded to 3.4 between the time these tests were run and the tests in #594 and #590.

tests/testthat/test-test_dir.R
```diff
@@ -33,3 +33,27 @@ test_that('filter_test_scripts() with tricky names', {
...
+  )
+  # With wrap should display two dots to screen
+
+  expect_true(sum(grepl("^Bare: \\.\\.", crayon::strip_style(w.wrap))) == 1L)
+
+  ## Without wrap should not display any dots; however, it is not possible to
+  ## test this because this test itself is in a `test_that` block, and further
+  ## since the whole test suite is effectively run with `wrap=TRUE` it is not
+  ## possible to test this at all.  I ran this outside of the `test_that` call
+  ## and it worked as expected.
```
gaborcsardi: Yes, you can do that, e.g. https://github.com/gaborcsardi/filelock/blob/master/tests/testthat/test.R#L27-L32

But it is kind of a big step, and brings in some dependencies as well, so I am not sure if it is worth it.

R/test-files.R
```diff
@@ -27,17 +27,20 @@ test_env <- function() {
 #' @return the results as a "testthat_results" (list)
 #' @export
 test_dir <- function(path, filter = NULL, reporter = default_reporter(),
-                     env = test_env(), ..., encoding = "unknown", load_helpers = TRUE) {
+                     env = test_env(), ..., encoding = "unknown", 
+                     load_helpers = TRUE, wrap = TRUE) {
```
brodieG: No because it is inherited from `test_file`  (@inheritParams).  You can see it in the Rd file: https://github.com/hadley/testthat/pull/597/files#diff-f540ef3582dc78a36153378451f7b979.

tests/testthat/test-test_dir.R
```diff
@@ -33,3 +33,27 @@ test_that('filter_test_scripts() with tricky names', {
     files[1]
   )
 })
+
+test_that('test_dir and `wrap`', {
+  ## forced to test screen output because tests outside of `testthat` blocks are
+  ## not recorded in return value, even when `wrap=TRUE` (see #596")
+
+  w.wrap <- capture.output(
+    test_dir('test_dir', filter='bare-expectations', wrap=TRUE)
```
brodieG: Sure, will address this evening.  I've tried to do it as much as possible (e.g. in the actual function definition), but old habits die hard.

tests/testthat/test-test_dir.R
brodieG: @krlmlr let me know if you want me to do this.

R/test-files.R
krlmlr: Oh, I did miss the `@inheritParams`...

tests/testthat/test-test_dir.R
krlmlr: Thanks -- and double quotes `"`, please ;-)

tests/testthat/test-test_dir.R
krlmlr: I guess we don't need it, because you will be using it all the time ;-)

brodieG:
Alright, done.  Also, merged in your master branch changes so the tests now pass!

Let me know if there is anything else I need to do here.

tests/testthat/test-test_dir.R
```diff
@@ -33,3 +33,27 @@ test_that('filter_test_scripts() with tricky names', {
...
+
+  expect_true(sum(grepl("^Bare: \\.\\.", crayon::strip_style(w.wrap))) == 1L)
+
+  ## Without wrap should not display any dots; however, it is not possible to
+  ## test this because this test itself is in a `test_that` block, and further
+  ## since the whole test suite is effectively run with `wrap=TRUE` it is not
+  ## possible to test this at all.  I ran this outside of the `test_that` call
+  ## and it worked as expected.
+
+  skip("untestable within test suite")
```
krlmlr: Forgot to mention that we don't need the `skip()` if we're not planning to implement a test. The rest looks good to me.

brodieG: Okay, so I'll just delete the test altogether then.  I'll do that this evening.

krlmlr: Oh, the test before `skip()` looks fine to me.

brodieG: Sorry, that's what I meant.

brodieG:
@krlmlr, I removed the skipped test.  Let me know if there is anything else that needs doing.

hadley:
I'd rather fully understand your problem before we consider a PR.

hadley:
Actually I think it'll be easier for me to finish this off.

PR #607, opened 2017-07-05
update to Catch 1.9.6

kevinushey:
This PR upgrades the bundled version of Catch to v1.9.6.


kevinushey:
Also updated to help make sure that `testthat` + Catch plays nicely with R routine registration.

hadley:
Can you please add a bullet to NEWS?

kevinushey:
Done!

kevinushey:
Test failures appear unrelated.

hadley:
Doh. I didn't realise you fixed the native registration here too 😭

R/catch-routine-registration.R
```diff
@@ -0,0 +1,6 @@
+# This dummy function definition is included with the package to ensure that
+# 'tools::package_native_routine_registration_skeleton()' generates the required
```
hadley: It's slightly easier to do `tools::package_native_routine_registration_skeleton(".", character_only = FALSE)` which is what I did

hadley: Hmmmm, overall I'm not sure automating this is the right approach because there is considerable variety in how people do symbol registration.  How does Rcpp do it?

kevinushey: `Rcpp::compileAttributes()` effectively calls `tools::package_native_routine_registration_skeleton()` and then runs the regular code that generates RcppExports.

Adding this function definition ensures that this works transparently for existing workflows.

hadley: Ah ok, perfect.

PR #612, opened 2017-08-07
Write TAP report to a file specified via options()

nealrichardson:
https://github.com/hadley/testthat/pull/467 proposed a way for the TAP reporter to write its output to a file, but its approach was seen as inconsistent with the other reporters, and it was closed. Since then, a JUnit reporter has been added to `testthat` (https://github.com/hadley/testthat/pull/481), and it allows specifying a "testthat.junit.output_file" option to direct the reporter to write to a file rather than stdout (https://github.com/hadley/testthat/pull/576). 

This pull request brings the TAP reporter in line with the JUnit one by adding a "testthat.tap.output_file" option, which, if set, similarly writes the TAP output to the specified file. The PR also adds a test for the JUnit reporter when "testthat.junit.output_file" is set.

### DIFF
--- a/R/reporter-tap.R
+++ b/R/reporter-tap.R
@@ -30,6 +30,13 @@ TapReporter <- R6::R6Class("TapReporter", inherit = Reporter,
       if (!self$has_tests)
         return()
 
+      # Check to see if an output_file path has been specified
+      self$out <- getOption("testthat.tap.output_file", self$out)
+      if (is.character(self$out) && file.exists(self$out)) {
+        # If writing to a file, overwrite it if it exists
+        file.remove(self$out)
+      }
+
       self$cat_line("1..", self$n)
       for (i in 1:self$n) {
         if (!is.na(self$contexts[i])) {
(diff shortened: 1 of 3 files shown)


hadley:
I'd rather do this globally by providing an option that controls where all reporters write to. Would you be interested in contributing a PR along those lines?  (I suspect the code will be pretty similar)

nealrichardson:
Sure, happy to prepare a PR that generalizes the option. My only concern about having a single `"testthat.output_file"` option governing all reporters is that it doesn't make sense for the `MultiReporter`--you'd have multiple streams going to the same file. 

One use I have for the MultiReporter is when I run tests on Jenkins. I like to use a MultiReporter with TAP (or now JUnit) writing to a file that Jenkins can parse and also use the `SummaryReporter` printing to stdout so that I can monitor progress in the job console logs, see when/if a build gets hung, etc. 

How would you feel about an approach like 

```
self$out <- getOption("testthat.junit.output_file",
                default=getOption("testthat.output_file",
                    default=stdout()))
```

(appropriately varied by reporter)? That is, allow a single "testthat.output_file" option, but allow you to override that with different files/connections for different reporters?

hadley:
At some point you're going to need to move away from options and call the reporter directly. I think having multiple levels of options makes things too complicated.  If you disagree, could you please discuss why in the related issue?

PR #633, opened 2017-10-06
testthat/catch/Rcpp compatibility workaround

zer0hedge:
If used, Rcpp disables dynamic loading of native routines and  'monopolizes' the registration of native routines in the package. This workaround intstructs Rcpp (if present)  to register 'run_testthat_tests' routine needed to run Catch tests and forces get_routine() to look for Rcpp-altered version of run_testthat_tests instead of failing if it can't find plain 'run_testthat_tests'. It is not very elegant, but testhat/catch now works with or without Rcpp and remains independent of Rcpp.

### DIFF
--- a/R/test-compiled-code.R
+++ b/R/test-compiled-code.R
@@ -196,8 +196,15 @@ get_routine <- function(package, routine) {
     error = function(e) NULL
   )
 
-  if (is.null(resolved))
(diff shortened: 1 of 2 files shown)


kevinushey:
FYI, https://github.com/hadley/testthat/pull/607 is going to ensure testthat properly handles the native routine registration stuff here.

PR #640, opened 2017-10-11
Add file and line number to test failures and errors in JunitReporter

nealrichardson:
The Junit report doesn't tell you where in the code the failures/errors are, which makes it difficult to track down the source of the problems. This change appends the `src_loc()`, which some of the other reporters use, to the message.

For what it's worth, TAP and teamcity reporters also appear not to include the source location in their output, and in general it appears that there is not a single standard for expressing this location. I explored a more generic solution as well (https://github.com/r-lib/testthat/compare/master...nealrichardson:format-expectation-failure) but thought it was too disruptive.

### DIFF
--- a/R/reporter-junit.R
+++ b/R/reporter-junit.R
@@ -98,19 +98,19 @@ JunitReporter <- R6::R6Class("JunitReporter", inherit = Reporter,
       )
 
       first_line <- function(x) {
-        strsplit(x, split = "\n")[[1]][1]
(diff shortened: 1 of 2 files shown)


hadley:
Thanks - can you please add a bullet to NEWS too?

If you have any PRs in the works, I'm about to re-style testthat to use the tidyverse style which is likely to cause merge conflicts.

hadley:
Actually let me add the news bullet - I'm about to rewrite the news file too.

PR #641, opened 2017-10-12
Remove deprecated function from examples

ruaridhw:
Updated `expect_less_than` as per [NEWS update from v1.0.0](../blob/70c4d53813d4191b370849412846c97daef3d867/NEWS.md#new-expectations)

Is it also necessary to keep the file [test-not.R](../blob/70c4d53813d4191b370849412846c97daef3d867/tests/testthat/test-not.R) if `not()` is deprecated as well?


hadley:
Thanks!  

Yeah, there's no need to keep the tests for `not()` around. I'll remove them

PR #663, opened 2017-10-24
Code style

krlmlr:
Each commit represents a styling level of styler. Some pieces are converted incorrectly, I have added issues to the styler repo and left the code untouched in these cases. One commit is a manual change, all other changes are automatic.

Reference:

- https://github.com/krlmlr/styler/issues/253
- https://github.com/krlmlr/styler/issues/254
- https://github.com/krlmlr/styler/issues/255
- https://github.com/krlmlr/styler/issues/256
- https://github.com/krlmlr/styler/issues/257

CC @lorenzwalthert.

### DIFF
--- a/R/expect-equality.R
+++ b/R/expect-equality.R
@@ -48,7 +48,6 @@ NULL
 #' @param ... other values passed to [all.equal()]
 expect_equal <- function(object, expected, ..., info = NULL, label = NULL,
                          expected.label = NULL) {
-
   act <- quasi_label(enquo(object), label)
   exp <- quasi_label(enquo(expected), expected.label)
 
@@ -66,7 +65,6 @@ expect_equal <- function(object, expected, ..., info = NULL, label = NULL,
 #' @export
 #' @rdname equality-expectations
 expect_setequal <- function(object, expected) {
-
   act <- quasi_label(enquo(object))
   exp <- quasi_label(enquo(expected))
 
@@ -103,7 +101,6 @@ expect_equivalent <- function(object, expected, ..., info = NULL, label = NULL,
 #' @rdname equality-expectations
 expect_identical <- function(object, expected, info = NULL, label = NULL,
                              expected.label = NULL) {
-
   act <- quasi_label(enquo(object), label)
   exp <- quasi_label(enquo(expected), expected.label)
 
@@ -131,7 +128,6 @@ expect_identical <- function(object, expected, info = NULL, label = NULL,
 #' @rdname equality-expectations
 expect_identical <- function(object, expected, info = NULL, label = NULL,
                              expected.label = NULL) {
-
   act <- quasi_label(enquo(object), label)
   exp <- quasi_label(enquo(expected), expected.label)
 
@@ -159,7 +155,6 @@ expect_identical <- function(object, expected, info = NULL, label = NULL,
 #' @rdname equality-expectations
 expect_reference <- function(object, expected, info = NULL, label = NULL,
                              expected.label = NULL) {
-
   act <- quasi_label(enquo(object), label)
   exp <- quasi_label(enquo(expected), expected.label)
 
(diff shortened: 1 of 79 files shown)


R/expect-equality.R
```diff
@@ -48,7 +48,6 @@ NULL
 #' @param ... other values passed to [all.equal()]
 expect_equal <- function(object, expected, ..., info = NULL, label = NULL,
                          expected.label = NULL) {
-
```
hadley: Ideally styler wouldn't strip this new line. It doesn't matter here, but in general it seems like a bad idea

R/expect-self-test.R
```diff
@@ -67,7 +67,8 @@ show_failure <- function(expr) {
 #' @rdname expect_success
 #' @param path Path to save failure output
 expect_known_failure <- function(path, expr) {
-  FailureReporter <- R6::R6Class("FailureReporter", inherit = CheckReporter,
+  FailureReporter <- R6::R6Class(
+    "FailureReporter", inherit = CheckReporter,
```
hadley: Why doesn't `"FailureReporter"` stay on the first line? (Given that it's an unnamed argument that fits)

R/reporter-junit.R
```diff
@@ -26,17 +26,18 @@ classnameOK <- function(text) {
...
 #' @export
-JunitReporter <- R6::R6Class("JunitReporter", inherit = Reporter,
+JunitReporter <- R6::R6Class(
+  "JunitReporter", inherit = Reporter,
   public = list(
-    results  = NULL,
-    timer    = NULL,
-    doc      = NULL,
-    errors   = NULL,
+    results = NULL,
```
hadley: I wonder if there's a way that styler could detect if the `=` were aligned, and if so, leave as is

R/stack.R
```diff
@@ -43,8 +42,8 @@ Stack <- R6Class(
   ),
 
   private = list(
-    stack = NULL,   # A list that holds the items
```
hadley: Preserving this alignment would also be nice

tests/testthat/test-test-that.R
```diff
@@ -89,46 +89,46 @@ expectation_lines <- function(code) {
 
 test_that("line numbers captured in simple case", {
   lines <- expectation_lines({
-    context('testing testFile')  # line 1
-    test_that('simple', {        # line 2
-      expect_true(FALSE)         # line 3
-    })                           # line 4
+    context("testing testFile") # line 1
```
hadley: Can you manually revert the changes to this file? The indenting of the comments is particularly important here

R/expect-self-test.R
lorenzwalthert: Because that is not yet implemented in styler. See https://github.com/krlmlr/styler/issues/246. For now, the line is broken after the first parenthesis with `strict = TRUE` if the whole call does not fit on one line.

R/expect-equality.R
krlmlr: Can we define a rule when to add/keep a newline after an opening brace? Is this just for top-level functions or for blocks in general?

krlmlr:
Thanks. I've changed alignment and applied a few more tweaks. The code of this PR doesn't change when running `styler:::style_pkg(scope = "tokens")` except in situations for which I have created issues in the styler repo.

R/expect-equality.R
hadley: I think I generally prefer a newline if the function def span multiple lines as it more clearly distinguishes the body from the args.

lorenzwalthert: So should we

* Add a section to the tidyverse style guide to leave one line blank after a multi-line function declaration and
* make styler in general preserving line breaks 
?

Not sure whether for the second bullet, we could also make some blank line formatting optional, but I could not think of anything particularly meaningful. Maybe one can choose to either to not touch blank lines, or to reduce the amount of blank lines to one if there is at least two blank lines next to each other.

hadley: I'm not sure how consistently I adhere to this style. There's also the question of wrapping multiline formals - we should probably be consistent with other calls, and if all arguments don't fit on one line, each argument gets it's own line.

lorenzwalthert:
Reference: https://github.com/r-lib/styler/issues/240

PR #674, opened 2017-11-10
API changes

krlmlr:
Comparison between 1.0.2 and 2.0.0: 7cb6304


krlmlr:
What's your preferred alternative to an `API` file?

hadley:
None, currently.

krlmlr:
Would you review a PR in _roxygen2_ that adds comments to `NAMESPACE` ? This would avoid introducing a new file.

I think it's very important to be able to track API changes.

PR #675, opened 2017-11-11
Add two columns to data reurned by list reporter

krlmlr:
- number of passed tests
- original results object (in a list column)

New columns added at end, should not disrupt existing code.

Is there any way we could add this to the upcoming release?


hadley:
This seems like a small change but given the number of packages using testthat it does seem possible this could introduce new failures. Could you do a bit of spelunking with github search to see how many packages are using the list reporter?

krlmlr:
Difficult to say, because the data is also the return value of `testthat::test_dir()` etc.. Easier to do a revdep check with downstream packages that import or depend on testthat. Will do.

krlmlr:
All checks were OK, but it's probably not worth the hassle. We may want to support this in the future, though.

hadley:
I think I'm happy to merge this now if you want to update and add a news bullet.

krlmlr:
Done.

hadley:
Thanks!

PR #679, opened 2017-11-13
Fix encoding issues in output recording and comparison

### LINKED ISSUE #672: expect_output / capture_output is broken on non UTF-8 platforms
The problem is that even if the output is not UTF-8, `read_lines` always marks it as UTF-8. Actually, even if you try to output (`cat`) proper UTF-8 like I do below, R will convert it to the native encoding (as much as possible), so the recorded output will not be UTF-8.

The behavior is better on the current stable version (1.0.2), because there readLines is used to read back the output, without a specified encoding, so the encoding of the output will be "unknown". This is still better than the current non-UTF-8 output marked as UTF-8. It also works fine in the checks, even if the encoding is not specified.

This is old testthat (example not self-contained, sorry):

```r
o1 <- paste(capture.output(print(boxx(""))), collapse = "\n")
o2 <- capture_output(print(boxx("")))
Encoding(o1)
#> [1] "latin1"
Encoding(o2)
#> [1] "unknown"
o1 == o2
#> [1] TRUE
expect_equal(o1, o2)
expect_match(o1, o2, fixed = TRUE)
expect_match(o2, o1, fixed = TRUE)
```

And this is new testthat. The raw representation of the output is still the same, i.e. it is `latin1`, but the output is marked as UTF-8:

```r
o3 <- capture_output(print(boxx("")))
Encoding(o3)
#> [1] "UTF-8"
stringi::stri_enc_isutf8(o3)
#> [1] FALSE
o3 == o1
#> [1] FALSE
all(charToRaw(o1) == charToRaw(o3))
[1] TRUE
```

And the check results are pretty bad, you get the same for using `expect_output()` directly, of course:

```r
expect_equal(o1, o3)
#> Error in nchar(x) : invalid multibyte string, element 1
expect_match(o1, o3, fixed = TRUE)
#> Error in grepl(regexp, act$val, perl = perl, fixed = fixed, ...) : 
#>   regular expression is invalid UTF-8
expect_match(o3, o1, fixed = TRUE)
#> Error: `o3` does not match "+------+\n¦      ¦\n¦      ¦\n¦      ¦\n+------+".
#> Actual value: "+------+\n<U+00A6>      <U+00A6>\n<U+00A6>      <U+00A6>\n<U+00A6>      #> <U+00A6>\n+------+"
#> In addition: Warning message:
#> In grepl(regexp, act$val, perl = perl, fixed = fixed, ...) :
#>   input string 1 is invalid UTF-8
```

I'll be happy to do a PR, but first we need to think about the correct solution....

gaborcsardi:
Closes #672.

Please see the commit messages for the details of the changes.

Maybe we also want to document it that both the code files, and the reference files are expected to be UTF-8 encoded now.

### DIFF
--- a/R/capture-output.R
+++ b/R/capture-output.R
@@ -51,6 +51,6 @@ eval_with_output <- function(code, print = FALSE, width = 80) {
   list(
     val = result$value,
     vis = result$visible,
-    out = read_lines(temp)
+    out = read_lines(temp, encoding = "unknown")
   )
 }
--- a/R/expect-known.R
+++ b/R/expect-known.R
@@ -57,6 +57,12 @@ expect_known_output <- function(object, file,
     ref_out <- read_lines(file)
     if (update) {
       write_lines(act$out, file)
+      if (!all_utf8(act$out)) {
+        warning("New reference output is not UTF-8 encoded", call. = FALSE)
+      }
+    }
+    if (!all_utf8(ref_out)) {
+      warning("Reference output is not UTF-8 encoded", call. = FALSE)
     }
 
     comp <- compare(act$out, enc2native(ref_out), ...)
--- a/R/expect-known.R
+++ b/R/expect-known.R
@@ -59,7 +59,7 @@ expect_known_output <- function(object, file,
       write_lines(act$out, file)
     }
 
-    comp <- compare(act$out, ref_out, ...)
+    comp <- compare(act$out, enc2native(ref_out), ...)
     expect(
       comp$equal,
       sprintf(
--- a/R/expect-output.R
+++ b/R/expect-output.R
@@ -102,7 +102,7 @@ expect_output <- function(object, regexp = NULL, ..., info = NULL, label = NULL)
       info = info
     )
   } else {
-    expect_match(act$cap, regexp, ..., info = info, label = act$lab)
+    expect_match(act$cap, enc2native(regexp), ..., info = info, label = act$lab)
   }
 
   invisible(act$val)
--- a/R/utils-io.R
+++ b/R/utils-io.R
@@ -1,8 +1,8 @@
 readLines <- function(...) stop("Use read_lines!")
 writeLines <- function(...) stop("Use write_lines!")
 
-read_lines <- function(path, n = -1L) {
-  base::readLines(path, n = n, encoding = "UTF-8", warn = FALSE)
+read_lines <- function(path, n = -1L, encoding = "UTF-8") {
+  base::readLines(path, n = n, encoding = encoding, warn = FALSE)
 }
 
 write_lines <- function(text, path) {
--- a/R/utils-io.R
+++ b/R/utils-io.R
@@ -8,3 +8,7 @@ read_lines <- function(path, n = -1L, encoding = "UTF-8") {
 write_lines <- function(text, path) {
   base::writeLines(enc2utf8(text), path, useBytes = TRUE)
 }
+
+all_utf8 <- function(x) {
+  ! any(is.na(iconv(x, "UTF-8", "UTF-8")))
+}
--- a/tests/testthat/test-expect-known-output.R
+++ b/tests/testthat/test-expect-known-output.R
@@ -38,3 +38,16 @@ test_that("updates by default", {
   expect_failure(expect_known_output(cat("oops"), file, update = TRUE))
   expect_success(expect_known_output(cat("oops"), file))
 })
+
+test_that("Warning for non-UTF-8 reference files", {
+  x <- "\xe9\xe1\xed\xf6\xfc"
+  Encoding(x) <- "latin1"
+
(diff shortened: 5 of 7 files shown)


krlmlr:
I think we should be reading as UTF-8 and never work in the native encoding. Would you consider using enc? Releasing today.

hadley:
I'd rather not add any more dependencies.  

But I think we'll still need to re-run the revdeps after this change, and delay release at needed. (At least until tomorrow)

krlmlr:
Happy to copy `read_lines_enc()` and `write_lines_enc()` to testthat. These functions read and write UTF-8 reliably, regardless of the native encoding.

gaborcsardi:
> I think we should be reading as UTF-8 and never work in the native encoding.

Is it possible to capture output in UTF-8 on Windows? I don't think so.

> But I think we'll still need to re-run the revdeps after this change, and delay release at needed. (At least until tomorrow)

I am not sure it is necessary, actually. These are the changes to user visible code:
* adding some `enc2native` calls. These do not change anything in UTF-8 locales, and the previous behavior was broken, anyway.
* reading back output in the native encoding. Again, this do not change anything in UTF-8 locales, and fixes the previous broken behavior.
* adding two warnings

The rest are changes in the test cases.

gaborcsardi:
> Happy to copy read_lines_enc() and write_lines_enc() to testthat. These functions read and write UTF-8 reliably, regardless of the native encoding.

The problem is not about reading in UTF-8. It is that you cannot capture output in UTF-8. The captured output is always in the native encoding. So we cannot read it as UTF-8.

hadley:
@gaborcsardi ok, if you're confident, we won't re-run the revdeps! (which I guess is pointless anyway since they're all on linux so won't illuminate the problem anyway)

gaborcsardi:
@hadley you could set a non-UTF-8 locale on Linux.... but I am confident that 
* this only affects printing non-ASCII characters on non-UTF-8 locales
* this probably won't break anything
* even if it does, without this `testthat` is still broken in non-UTF-8 platforms,
* this is still the correct thing to do, wrt circumstances.

krlmlr:
Can we convert to UTF-8 as soon as we have captured the output? Otherwise our code would suffer from native encoding problems even if this problem is fixed upstream.

I'm happy to rerun the revdep checks in `LC_ALL=C` tonight just to be sure. This is the better thing to do anyway, because on my system the new processx fails to report check failures to revdepcheck. I only noticed this when running tibble checks, because we still should have two failing downstream packages there. Issue in processx follows.

gaborcsardi:
On 13 Nov 2017 15:06, "Kirill Müller" <notifications@github.com> wrote:

Can we convert to UTF-8 as soon as we have captured the output?

UTF-8 -> native -> UTF-8 loses information on windows, so no.

Otherwise our code would suffer from native encoding problems even if this
problem is fixed upstream.

What kind of problems?

krlmlr:
Our code should use UTF-8 internally, but currently we call `enc2native()`. I'm confident we shouldn't do it. If we get output in the native encoding (as we currently do on Windows), we convert to UTF-8 -- this is lossless. If `capture.output()` is fixed in R on Windows, and we still call `enc2native()`, our code will be responsible for losing information.

We still can read and write UTF-8 files on Windows with `enc:::read_lines_enc()` and `enc:::write_lines_enc()`, so no roundtrip to native here.

gaborcsardi:
> Our code should use UTF-8 internally, but currently we call enc2native(). I'm confident we shouldn't do it. If we get output in the native encoding (as we currently do on Windows), we convert to UTF-8 -- this is lossless.

The UTF-8 -> native step, performed by `cat`, etc. is lossy.

> If capture.output() is fixed in R on Windows, and we still call enc2native(), our code will be responsible for losing information.

Currently we use the default (native) encoding on the file we are saving the output to. So if `cat` is fixed, nothing changes for us.

krlmlr:
Your solution does look like the lesser of all evils to me. I still think it's cleaner to use http://utf8everywhere.org/, but it doesn't seem to be worth the effort in this particular situation.

gaborcsardi:
>  I still think it's cleaner to use http://utf8everywhere.org/

It is just not possible. R converts to native encoding and currently we can do nothing about it.

> it doesn't seem to be worth the effort in this particular situation.

I am not even sure what that effort would entail. 

Converting back to UTF-8? Then, to compare to the reference, you would need to do the same to the UTF-8 reference output, i.e. have `enc2utf8(enc2native(ref_out))`. I don't see how that is any better than the current solution.

When we fix base R, so that writing a declared UTF-8 string to a UTF-8 connection is not re-encoded twice, then we can do better.

krlmlr:
Well, we could convert each string we see to UTF-8 as soon as possible. I attempted an `enc2utf8(enc2native(ref_out))` approach but, as you said, it's not much better.

PR #701, opened 2018-01-10
Print progress only if enough time has passed.

jimhester:
Currently progress is printed after every test, which can cause a significant portion of the total test time taken to be used by printing if each test runs quickly. You also get a distracting flashing cursor when the updates occur too frequently.

As a real world example https://github.com/r-lib/fs/compare/master...sanitize adds a bunch of tests which happen very fast, the total time taken for all tests in the fs package with the current master of testthat is ~5.973 seconds. With this PR the total time drops to is 4.254 seconds, a considerable change considering it is just test printing overhead.

This PR currently reuses the min_time parameter, which seems to work ok, 1/10 of a second is a reasonable amount of time between updates. But if necessary we could add an additional option or hardcode it to some other value.

### DIFF
--- a/R/reporter-progress.R
+++ b/R/reporter-progress.R
@@ -18,6 +18,8 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
     show_praise = TRUE,
     min_time = 0.1,
     start_time = NULL,
+    last_update = NULL,
+    update_interval = NULL,
 
     max_fail = NULL,
     n_ok = 0,
@@ -37,11 +39,13 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
     initialize = function(show_praise = TRUE,
                           max_failures = getOption("testthat.progress.max_fails", 10L),
                           min_time = 0.1,
+                          update_interval = 0.1,
                           ...) {
       super$initialize(...)
       self$max_fail <- max_failures
       self$show_praise <- show_praise
       self$min_time <- min_time
+      self$update_interval <- update_interval
     },
 
     is_full = function() {
@@ -85,6 +89,11 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
           status <- crayon::green(cli::symbol$tick)
         }
       } else {
+
+        # Do not print if not enough time has passed since we last printed.
+        if (!self$should_update()) {
+          return()
+        }
         status <- spinner(self$ctxt_n)
       }
 
@@ -109,6 +118,7 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
     end_context = function(context) {
       time <- proc.time() - self$ctxt_start_time
 
+      self$last_update <- NULL
       self$show_status(complete = TRUE)
 
       if (time[[3]] > self$min_time) {
@@ -187,6 +197,20 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
       } else {
         self$cat_line(colourise(encourage(), "error"))
       }
+    },
+
+    should_update = function() {
+      if (self$update_interval == 0) {
+        return(TRUE)
+      }
+
+      time <- proc.time()[[3]]
+      if (!is.null(self$last_update) &&
+        (time - self$last_update) < self$update_interval) {
+        return(FALSE)
+      }
+      self$last_update <- time
+      TRUE
     }
   )
 )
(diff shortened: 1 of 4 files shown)


tests/testthat/reporters/progress.txt
```diff
@@ -1,12 +1,5 @@
 ✔ | OK F W S | Context
 
-⠏ |  0       | Expectations
-⠋ |  1       | Expectations
-⠙ |  1 1     | Expectations
-⠹ |  1 2     | Expectations
-⠸ |  1 3     | Expectations
-⠼ |  1 4     | Expectations
-⠴ |  2 4     | Expectations
 ✖ |  2 4     | Expectations
```
gaborcsardi: Do you think it is possible to keep these lines? Because otherwise the tests are not really testing the output. Maybe by setting `min_time`?

Timing based tests are not too great in general, but of course in this case we can't really avoid them. We can have another set of (the same?) tests that set `min_time` to something very large, so that the output is omitted on very slow systems as well.

R/reporter-progress.R
```diff
@@ -78,15 +80,22 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
...
           status <- crayon::red(cli::symbol$cross)
         } else {
           status <- crayon::green(cli::symbol$tick)
         }
       } else {
+
+        # Do not print if not enough time has passed since we last printed.
+        if ((time - self$last_time)[[3]] < self$min_time) {
+          return()
+        }
```
gaborcsardi: I would always print a line at the beginning of a new context/file, so that you at least know which test file is running. I.e. this:
```
⠏ |  0       | Expectations
```

Otherwise, if the first test case takes very long, you'll have no output from the file for a long time.

jimhester:
Ok I changed how this was implemented slightly, there is now a separate `update_interval` argument which when set to `0` uses the old behavior of updating every time a test completes, which we set when testing. I set `update_interval` to `0.1`, but we can change it to something else if we find that doesn't update often enough.

jimhester:
It also now should always update when there is a new context.

R/reporter-progress.R
```diff
@@ -37,11 +39,13 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
     initialize = function(show_praise = TRUE,
                           max_failures = getOption("testthat.progress.max_fails", 10L),
                           min_time = 0.1,
+                          update_interval = 0,
```
gaborcsardi: Should the default be 0.1 here? Otherwise it is never throttled.

R/reporter-progress.R
```diff
@@ -18,6 +18,8 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
     show_praise = TRUE,
     min_time = 0.1,
     start_time = NULL,
+    last_update = NULL,
+    update_interval = 0.1,
```
gaborcsardi: It is a small thing, but I would set this to `NULL` here. It is set by `initialize` anyway, and then this magic number is misleading.

R/reporter-progress.R
jimhester: Yeah, I changed the wrong value.

gaborcsardi:
Looks good, but the CI is failing. :(

jimhester:
I missed setting `update_interval = 0` for one test instance, should pass now 🤞

R/reporter-progress.R
```diff
@@ -78,13 +82,19 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
     },
 
     show_status = function(complete = FALSE) {
+      time <- proc.time()
```
hadley: Does this get used somewhere?

R/reporter-progress.R
```diff
@@ -187,6 +198,20 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
...
+    },
+
+    should_update = function() {
+      if (self$update_interval == 0) {
+        return(TRUE)
+      }
+
+      time <- proc.time()
+      if (!is.null(self$last_update) &&
+        (time - self$last_update)[[3]] < self$update_interval) {
```
hadley: Maybe just save `proc.time()[[3]]` to make a little simpler?

R/reporter-progress.R
jimhester: No it is vestigial, thanks for catching it.

jimhester:
Thank you both for the reviews!

PR #704, opened 2018-01-15
Update praise text

jrosen48:
Super minor-and I loved the praise when I got it-but think that no one should be spelled without a dash, i.e. [here](http://www.grammar-monster.com/easily_confused/no-one_no_one.htm). This changes "no-one" to "no one" in the two statements that include them.


hadley:
Thanks!

PR #707, opened 2018-01-25
changed the expect_equal_to_reference function to hand over the updat…

ktrask:
The news.md states that expect_equal_to_reference is equivalent to expect_known_value, but with a default of update=FALSE. Unfortunately, the function did not hand over these update-variable to the expect_known_value function. This commit fixes that.


jimhester:
Thanks! this was fixed by https://github.com/r-lib/testthat/pull/687/

PR #721, opened 2018-02-24
Do not handle warnings when options("warn") >= 2

yutannihilation:
When `options("warn")` >= 2, a warning is converted into an error. So, in that case, we should not handle it by the warning handler, but do nothing and let it handled by the error handler.

This usually won't be a problem, except when a package has some tricky tests that change the value of `options("warn")`, but actually evaluate package suffers from this: https://github.com/r-lib/evaluate/pull/81#issuecomment-367685196

I believe this PR will fix the problem, but I'm not fully sure there's no downside... Should this be treated as a known issue?


jimhester:
Thanks! I think passing the warning on in this case makes sense, it also allows you set `options(warn = 2)` in your tests if you want warnings to fail the build ala https://github.com/r-lib/testthat/issues/731, https://community.rstudio.com/t/fail-ci-on-warnings-in-tests/6187/9

yutannihilation:
Oh, I didn't come up with the usage. Thanks!

PR #736, opened 2018-03-29
Add skip_if_offline

### LINKED ISSUE #685: Add skip_if_offline()
Would a PR to add something like this be accepted? I suppose the ping-ee should be settable, possibly defaulting to google.

``` r
skip_if_offline <- (function() {
  offline <- NA
  function() {
    if (is.na(offline)) {
      offline <<- tryCatch(
        is.na(pingr::ping_port("google.com", count = 1, timeout = 1)),
        error = function(e) TRUE
      )
    }
    if (offline) testthat::skip("Offline")
  }
})()
```

I'm using it in googledrive and adapted it from something @gaborcsardi did in gh.

It means Importing or Suggesting [pingr](https://cran.r-project.org/package=pingr), which doesn't have any dependencies but does require compilation.

jimhester:
This is useful for packages which test against online APIs. We use
`curl::nslookup()` rather than `utils::nsl()` because the latter is not
available on Windows.

Fixes #685


jennybc:
This will be great!

Reading between the lines, I gather you may not be keen on the idea of determining offline status for a test run vs individual tests, which I championed in https://github.com/r-lib/testthat/issues/685#issuecomment-353150775.

Are you pro, against, neutral?

Any technical thoughts about how to implement it, whether it's built into testthat or achieved via an *ad hoc* test helper wrapping the function in this PR?

jimhester:
Mostly against, if the connection is flaky as you suggest you may get spurious errors rather than spurious skips, which seems worse.

Looking up the namespace is quite fast, so having this skip multiple times in your tests is not really an issue.

``` r
system.time({ i <- 1; while((i <- i + 1) <= 10000) !is.null(curl::nslookup("r-project.org", error = FALSE))})
#>    user  system elapsed 
#>   1.062   0.756   3.514
```
> Created on 2018-03-29 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

If you do want to do the test once per run I would put
```r
has_internet <- !is.null(curl::nslookup(host, error = FALSE))
``` 
In a helper or setup file and have `skip_if_not(has_internet)` in my test files.

jimhester:
Thanks for the review!

PR #737, opened 2018-03-30
Use 3 character wide columns for successful tests

### LINKED ISSUE #712: Report is misaligned on 3digit tests
A tiny issue. Bars are misaligned when more than 100 tests per context are involved:

```
Testing lubridate
✔ | OK F W S | Context
✔ | 144       | Accessors [0.4 s]
✔ |  2       | Test days_in_month
✔ | 13       | AM/PM
✔ | 24       | Dates [0.1 s]
✔ | 39       | Daylight savings times [0.1 s]
✔ | 12       | Decimal_date
✔ | 10       | Difftimes
✔ | 74       | Durations [0.2 s]
✔ |  1       | Guessing format
✔ | 16       | Instants [0.1 s]
✔ | 199       | Intervals [1.7 s]
✔ |  4       | Namespace [2.4 s]
✔ | 91       | Addition operations [0.4 s]
✔ | 17       | Comparisons operations
✔ | 75       | Division operators [0.4 s]
✔ | 32       | Integer division operations [0.2 s]
✔ | 36       | Modulo operations [0.2 s]
✔ | 12       | Multiplication operations
✔ | 44       | Subtraction operations [0.2 s]
✔ | 372       | Parsing [1.7 s]
✔ | 157       | Periods [0.8 s]
✔ | 32       | POSIXt [0.2 s]
✔ |  2       | Pretty formatting of dates
✔ |  5       | Rollback
✔ | 238       | Rounding [0.9 s]
✔ | 404       | Settors [1.2 s]
✔ | 69       | Stamp [0.7 s]
✔ | 38       | Timespans [0.3 s]
✔ | 27       | Time zones [0.1 s]
✔ | 250       | Updating dates [0.7 s]
✔ | 12       | Utilities
```

jimhester:
This is a simple solution to the problem of more than 100 successful tests messing up the column alignment. I don't know if we want to try and do something more sophisticated, or if this is sufficient to handle the majority of cases.

FIxes #712

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,9 @@
 * Pass through warnings when `options(warn = 2)` is set (#721, @yutannihilation).
 * Progress reporter now generates a context from the filename and no longer
   errors if tests occur before a context (#700, #705).
+* The progress reporter now uses a 3 character wide column to display test
+  successes, so up to 999 successful tests can be displayed without changing
+  the alignment (#712).
 
 * `expect_lt()`, `expect_lte()`, `expect_gt()` `expect_gte()` now handle `Inf`
   and `NA` arguments appropriately (#732).
--- a/R/reporter-progress.R
+++ b/R/reporter-progress.R
@@ -90,7 +90,7 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
 
     show_header = function() {
       self$cat_line(
-        cli::symbol$tick, " | OK ",
+        cli::symbol$tick, " |  OK ",
         colourise("F", "failure"), " ",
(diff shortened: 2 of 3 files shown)


gaborcsardi:
An alternative would be to use 2 characters by default, but when it is wider, make the other column(s) narrower. They are probably empty, or at least do not need two characters,  anyway. 

But I don't have a strong preference for this.

jimhester:
Yeah I think making the other columns smaller will just push the problem there, so am just going to make this one bigger and call it a day.

Thanks for the review!

PR #742, opened 2018-04-11
Set an environment variable with the package being tested

### LINKED ISSUE #699: Inspecting the name of the package being tested
This would be useful for encapsulating the effects of verbose retirement to the package being tested.

jimhester:
@hadley Do you think setting an environment variable is the right thing to do here? If so we would also need to set it in `devtools::test()`, which I  can do, e.g. https://github.com/r-lib/devtools/compare/master...jimhester:testthat-pkg-name

testthat actually already assigns the package name to a `package` object in the test environment https://github.com/r-lib/testthat/blob/9066e991c7a8aec485c0b5d9bcbf6399ae866047/R/test-directory.R#L115
but this seems somewhat cumbersome to retrieve, doesn't work when running interactively and also won't work if you have nested R processes.

Fixes #699


NEWS.md
```diff
@@ -12,6 +12,9 @@
 
 ## Minor improvements and bug fixes
 
+* testing_package() function is available to retrieve the name of the package
```
hadley: Missing `` ` ``

PR #761, opened 2018-05-18
Naive fix for auto_test_package

### LINKED ISSUE #759: BROKEN: auto_test_package no longer works
Took a lot of debugging, but the core issue is in testthat...

### Observed behavior

```
~/dev/pds3(master*)$ Rscript -e "testthat::auto_test_package()"
Error in path.expand(path) : invalid 'path' argument
Calls: <Anonymous> ... rprojroot_find_root -> get_start_path -> normalizePath -> path.expand
Execution halted
```

### Root Cause

I'm not sure where the regression was introduced, but the problem is that currently pkgload::load_all requires a path and testthat is passing a package here https://github.com/r-lib/testthat/blob/master/R/auto-test.R#L90
```
pkg <- devtools::as.package(pkg)
...
env <- devtools::load_all(pkg)$env
```

As a result, the eventual pathNormalize is being passed a 'package' rather than a string

Relevant code in devtools and pkgload: 

 * https://github.com/r-lib/devtools/blob/master/R/pkgload.R#L4
 * https://github.com/r-lib/pkgload/blob/master/R/load.r#L95

### Fix

I believe that changing 
```
env <- devtools::load_all(pkg)$env
```
to 
```
env <- devtools::load_all(pkg$path)$env
```
should do the trick, but this part of the ecosystem is such a maze of packages, I'm hesitant to know exactly the correct place to make a change or where the original regression occurred as testthat hasn't actually changed to break this behavior

mwaldstein:
This fixes #759 for me

### DIFF
--- a/R/auto-test.R
+++ b/R/auto-test.R
@@ -87,7 +87,7 @@ auto_test_package <- function(pkg = ".", reporter = default_reporter(), hash = T
   test_path <- normalizePath(file.path(pkg$path, "tests", "testthat"))
 
   # Start by loading all code and running all tests
-  env <- devtools::load_all(pkg)$env
+  env <- devtools::load_all(pkg$path)$env
   withr::with_envvar(
     devtools::r_env_vars(),
     test_dir(test_path, env = env, reporter = reporter$clone(deep = TRUE))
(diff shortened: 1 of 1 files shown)


jimhester:
Thanks! Please add a note to news.md describing the change and mentioning the issue number and your GitHub user name, then we can merge this

mbojan:
@jimhester , can this be merged?

jimhester:
This is fixed by https://github.com/r-lib/devtools/commit/bcdde3985c8220accc975e38236e5f802a0036b2, so it does not require changes to testthat.

Issue #754, opened 2018-05-10
`encoding` is deprecated; all files now assumed to be UTF-8


vincentvanhees:
I am trying to address:

**Warning message:
"`encoding` is deprecated; all files now assumed to be UTF-8"**

as produced by testthat functions [test-directory](https://github.com/r-lib/testthat/blob/master/R/test-directory.R) and [source](https://github.com/r-lib/testthat/blob/master/R/source.R) for package [GGIR](https://github.com/wadpac/GGIR).

I only get this warning message when testing the package as a whole, and not when running the individual tests with test_file().

What did NOT help was:
- Adding `options(encoding = "UTF-8")`  to the top of each test file one line below library(mypackagename).
- Adding `options(encoding = "UTF-8")`  to testthat.R one line below library(mypackagename).
- Adding argument `fileEncoding="UTF-8"` to all my calls to `write.table`.

Could it be that the warning message is printed regardless of whether I use UTF-8 encoding or not and that I am trying to fix an issue that does not really exist? If yes, should the warning message then be a note rather than an warning? If not, could you advise me on what might cause this warning message and how to address it?

thanks, Vincent (and thanks for the great package!)

My sessionInfo():
```
R version 3.4.4 (2018-03-15)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 16.04.4 LTS
Matrix products: default
BLAS: /usr/lib/openblas-base/libblas.so.3
LAPACK: /usr/lib/libopenblasp-r0.2.18.so
locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=nl_NL.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=nl_NL.UTF-8   
 [6] LC_MESSAGES=en_US.UTF-8    LC_PAPER=nl_NL.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=nl_NL.UTF-8 LC_IDENTIFICATION=C       
attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     
other attached packages:
[1] testthat_2.0.0 GGIR_1.5-22   
loaded via a namespace (and not attached):
[1] compiler_3.4.4      magrittr_1.5        R6_2.2.2            tools_3.4.4         yaml_2.1.18         Rcpp_0.12.16       
[...]

PR #767, opened 2018-06-16
modifying condition to get encoding deprecation message

baogorek:
Issue #754 complains about the message:
```
`encoding` is deprecated; all files now assumed to be UTF-8 
```
which has been appearing lately when using `devtools::test()`.  @kenahoo suggested checking for the encoding argument only if  `testthat` is less than a particular version.  I thought about that but noticed a few different things during experimentation with `devtools::test()`:

- If no encoding is specified in the DESCRIPTION file, encoding gets set to "unknown" in `testthat::test_dir` and `missing(encoding)` returns `FALSE`, thus triggering the message (and feels "out of the blue")
- One way to silence the message is to add "ENCODING: UTF-8" in the DESCRIPTION file, because only then is encoding both not missing and identical to "UTF-8".
- Any other valid encoding also triggers the message, and invalid encodings produce errors before this warning message is ever encountered.

If the encoding field really is meant to be deprecated, I feel that you should get the warning even if you specified an encoding of UTF-8. According to [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-DESCRIPTION-file), there is reason not to do this as it makes the package "less portable."

Thus, the very small change proposed in this pull request triggers the warning message when and only when a package's DESCRIPTION file specifies an encoding, no matter what that encoding is.

### DIFF
--- a/R/test-directory.R
+++ b/R/test-directory.R
@@ -65,7 +65,7 @@ test_dir <- function(path,
                      stop_on_failure = FALSE,
                      stop_on_warning = FALSE,
                      wrap = TRUE) {
-  if (!missing(encoding) && !identical(encoding, "UTF-8")) {
+  if (!identical(encoding, "unknown")) {
     warning("`encoding` is deprecated; all files now assumed to be UTF-8", call. = FALSE)
   }
 


henrykironde:
Any updates on this issue?

alexsanjoseph:
👍

Hong-Revo:
Note that as of R 3.5, specifying an encoding in a package's DESCRIPTION will trigger a warning, at least on Windows. So it's probably a good idea to get this fix merged ASAP.

baogorek:
Glad to hear a fix is on the way! Thanks for looking into this.

PR #779, opened 2018-07-25
mark inline functions containing static members as hidden

kevinushey:
Closes https://github.com/r-lib/devtools/issues/1832.

### DIFF
--- a/inst/include/testthat/testthat.h
+++ b/inst/include/testthat/testthat.h
@@ -12,9 +12,16 @@
  * Force 'testthat' to be disabled by defining TESTTHAT_DISABLED.
  * TESTTHAT_DISABLED takes precedence.
  * 'testthat' is disabled on Solaris by default.
+ *
+ * Hide symbols containing static members on gcc, to work around issues
+ * with DLL unload due to static members in inline functions.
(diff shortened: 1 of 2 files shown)


inst/include/testthat/testthat.h
```diff
@@ -12,9 +12,16 @@
...
  * TESTTHAT_DISABLED takes precedence.
  * 'testthat' is disabled on Solaris by default.
+ *
+ * Hide symbols containing static members on gcc, to work around issues
+ * with DLL unload due to static members in inline functions.
+ * https://github.com/r-lib/devtools/issues/1832
  */
 #if defined(__GNUC__) || defined(__clang__)
 # define TESTTHAT_ENABLED
+# define TESTTHAT_ATTRIBUTE_HIDDEN __attribute__ ((visibility("hidden")))
```
jimhester: Shouldn't this only be defined for gcc and not for clang or does this behavior happen for clang as well?

kevinushey: I was thinking that since we don't really require these symbols to be exported in any case, it's better to mark them as hidden.

jimhester:
Thanks Kevin, glad you were able to track this down!

PR #788, opened 2018-08-04
Set env vars before sourcing test helpers and setup

jennybc:
I would like to consult the TESTTHAT env var inside `setup.R` and possibly even `helper.R`, so it would be nice if it were defined a bit earlier.


jimhester:
Thanks!

PR #793, opened 2018-08-18
Support for wildcard in expect_known_output

lbartnik:
* new argument for `expect_known_output`: `wildcard` accepts a single character matched against an arbitrary string; by default it is `NULL` which means no wildcard
* new matching algorithm in `match.R` and its tests in `test-match.R`
* new tests for and examples for `expect_known_output`


hadley:
This is a very large change to how `expect_known_output()` works, so can you please open an issue to discuss the motivation?

PR #802, opened 2018-09-24
Expose chdir arguments in test functions

cfhammill:
Sometimes you don't want to run tests in the environment
where those files are. For example users may not have write
access to an R package installation directory but still
want to run tests that require file IO. This commit
exposes the `chdir` option of `source_file` to test
runner functions. Can be disabled to run tests in the
current working directory.

### DIFF
--- a/R/auto-test.R
+++ b/R/auto-test.R
@@ -27,10 +27,11 @@
 #' @param env environment in which to execute test suite.
 #' @param hash Passed on to [watch()]. When FALSE, uses less accurate
 #'   modification time stamps, but those are faster for large files.
+#' @inheritParams test_dir
 #' @keywords debugging
 auto_test <- function(code_path, test_path, reporter = default_reporter(),
                       env = test_env(),
-                      hash = TRUE) {
+                      hash = TRUE, chdir = TRUE) {
   reporter <- find_reporter(reporter)
   code_path <- normalizePath(code_path)
   test_path <- normalizePath(test_path)
@@ -55,7 +56,7 @@ auto_test <- function(code_path, test_path, reporter = default_reporter(),
     } else if (length(tests) > 0) {
       # If test changes, rerun just that test
       cat("Rerunning tests: ", paste0(basename(tests), collapse = ", "), "\n")
-      test_files(tests, env = env, reporter = reporter$clone(deep = TRUE))
+      test_files(tests, env = env, reporter = reporter$clone(deep = TRUE), chdir = chdir)
     }
 
     TRUE
@@ -70,9 +71,11 @@ auto_test <- function(code_path, test_path, reporter = default_reporter(),
 #' @param reporter test reporter to use
 #' @param hash Passed on to [watch()].  When FALSE, uses less accurate
 #'   modification time stamps, but those are faster for large files.
+#' @inheritParams test_package
 #' @keywords debugging
 #' @seealso [auto_test()] for details on how method works
-auto_test_package <- function(pkg = ".", reporter = default_reporter(), hash = TRUE) {
+auto_test_package <- function(pkg = ".", reporter = default_reporter(),
+                              hash = TRUE, chdir = chdir) {
   if (!requireNamespace("devtools", quietly = TRUE)) {
     stop(
       "devtools required to run auto_test_package(). Please install.",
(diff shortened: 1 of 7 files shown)


gaborcsardi:
Can you just change the working directory before running the tests? Or even in the test cases?

If you want file IO in test cases, one solution is to use the session temporary directory, see `?tempfile()`, this should be writeable.

cfhammill:
Nope, `source_file` will `setwd` to where the test file lives, so as a user you can't just move before running the tests. `tempfile` does mostly solve the problem, it only fails if you want to run tests on a cluster, in which case the nodes can't see `/tmp`

jimhester:
Thanks, can you add a note to NEWS.md describing the change and mentioning your GitHub username and the issue number?

gaborcsardi:
> tempfile does mostly solve the problem, it only fails if you want to run tests on a cluster, in which case the nodes can't see /tmp

 You can set the `TMPDIR` env var (before starting R), to point to a different temporary directory if you don't have `/tmp` on the machine.

cfhammill:
@jimhester thanks, absolutely can do. @gaborcsardi great suggestion, I hadn't thought of that, does require a little more expertise on the user's end though.

gaborcsardi:
@cfhammill How would you actually use this option in a package?

cfhammill:
I've already prototyped it in mine, the test bed runs `test_dir` with `chdir = FALSE`. Now users on my system can test for themselves that the package works. Before they would be trying to do IO in my directories which wouldn't work.

gaborcsardi:
So the users need to specify `test_dir(chdir = FALSE)` themselves?  I think it is just better to use `tempfile()`, hundreds of packages do that already. If the user can run R, then R must have access to a writeable temporary directory. I am pretty sure that R does not run without one.

Running package tests with `test_dir` is not the best, because you don't have access to internal (non-exported) functions of the package.

jimhester:
I mostly agree with @gaborcsardi, first tests are not typically installed by default (unless `--install-tests`) is used when the package is installed. In addition usually users would run tests via `R CMD check` or `tools::testInstalledPackages()` rather than using testthat functions directly.

cfhammill:
@gaborcsardi  `test_dir` has served us fine up until this point, but `test_package` would serve just as well, it would also suffer from this problem. As mentioned above `tempfile` requires the use to manually specify their `TMPDIR` to deal with tests on our HPC cluster, again not a big problem, but more error prone.

@jimhester our tests are installed by default via a symlink from tests to inst. We provide a test runner function which downloads example data, sets up the environment, and then runs the tests.  It's in this function that the `chdir` argument is set.

But as you both allude, this is a pretty case specific problem, I understand the resistance to changing the API.

gaborcsardi:
> As mentioned above tempfile requires the use to manually specify their TMPDIR to deal with tests on our HPC cluster, again not a big problem, but more error prone.

Last time I checked R was not able to run without a writeable temporary directory.  I am pretty sure this  still holds, and R does not even start, unless it can create one. If your users can start R, then they already have a working writeable temporary directory in `tempdir()`, where the tests will be able to write.

gaborcsardi:
> But as you both allude, this is a pretty case specific problem, I understand the resistance to changing the API.

It is not end of the world, and adding a new argument at the end of the args list(s) is not a big deal in R in general. But if there is already a way to solve your problem well, it makes sense to avoid it.

cfhammill:
You do have a writeable `TMPDIR` but the cluster nodes cannot see it. The cluster nodes will also have writeable `TMPDIR`s, but they will not contain the same files.

gaborcsardi:
I am sorry, but I don't get this. Please bear with me. 

So a cluster node runs R, right? That means that it already has a temporary directory. Basically all Unix-like systems have TMPDIR already set, actually, it is included in most Unix standards.

The cluster node must also have access to the package, including tests. (Btw. instead of the symlink, the standard way to include the test in the installed package is  `R CMD INSTALL --install-tests`.) So  the cluster node (your supplied function) can just call `tools::testInstalledPackage()`. It will work, it has to. If your test cases use `tempdir()` and/or `tempfile().

The primary node and the cluster nodes do not have to have the same TMPDIR. You do no have to  copy the test cases to TMPDIR. Just keep them at their proper place, in `/tests` in you package.

Here is an example of a test case with IO, that I just wrote: https://github.com/r-lib/sessioninfo/blob/c89a5d0ebf008030fa4a1b03c5530a7f4793e92d/tests/testthat/test-warnings.R#L14-L15

If you install this package using `--install-tests`, then running 
```
R -q -e 'tools::testInstalledPackage("sessioninfo", types = "tests")'
```
will run the test cases just fine, even if the package library is not writeable.

cfhammill:
The tests cannot share data in this setup, which is not ideal. On many clusters work nodes do not have internet access and would be unable to download the test data even if that was reasonable.

hadley:
Given that @jimhester and @gaborcsardi are ambivalent about the change, and it requires threading arguments through many calls, I unfortunately I don't think it's worthwhile to merge this PR.

PR #803, opened 2018-09-25
Fixes for devel devtools

jimhester:


jimhester:
This is a place to store the diff for a testthat patch release.

It has conflicts vs the master branch, but not based on the v2.0.0 tag. e.g. https://github.com/r-lib/testthat/compare/v2.0.0...jimhester:v2.0.1-rc

PR #806, opened 2018-10-16
Improve junit time format compatibility

comicfans:
Jenkins junit plugin requires time has at most 3 digits, or it reports 

```
junit.xml:cvc-pattern-valid: Value '5.44999999999999' is not facet-valid with respect to pattern '(([0-9]{0,3},)*[0-9]{3}|[0-9]{0,3})*(\.[0-9]{0,3})?' for type 'SUREFIRE_TIME'.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,6 +43,8 @@
 * `expect_lt()` and friends now work with any object that defines the
   appropriate comparison method. (#777)
 
+* improve junit xml compatibility for jenkins. (#806, @comicfans)
+
 # testthat 2.0.1
 
 * Fix failing tests with devtools 2.0.0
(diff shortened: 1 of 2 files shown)


hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

comicfans:
of course , but I found bullet is written as ```(#issuenumber, @username)``` so I followed this style

hadley:
Thanks, although in the future please follow the new bullet style that I requested, not the old.

comicfans:
Ok

PR #817, opened 2018-11-16
Adding 'info' parameter to comparison tests for consistency with other functions

gvwilson:
- Add `info` parameter to `expect_lt`, `expect_lte`, `expect_gt`, and `expect_gte`.
- Modify `expect_compare` to pass through.
- Add tests to ensure that value of `info` parameter shows up in output.

### DIFF
--- a/R/expect-comparison.R
+++ b/R/expect-comparison.R
@@ -20,7 +20,8 @@
 #' @name comparison-expectations
 NULL
 
-expect_compare <- function(operator = c("<", "<=", ">", ">="), act, exp) {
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks @gvwilson!

@hadley wasn't there at some point the idea to deprecate the `info` parameter in `expect_()` functions in favor of the pseudo tidy evaluation we now have, or am I miss-remembering?

hadley:
Yeah, the goal is to eventually deprecate it, which is why the newer expectations don't have it.

Issue #786, opened 2018-08-04
Remove test that mocks devtools::add_path()


jennybc:
`devtools::add_path()` does not exist in current devtools (https://github.com/r-lib/devtools/commit/dbf00241ec0a48ea1a19e7c96be2ace21d7e726d) and this test mocks it and therefore fails for me:

https://github.com/r-lib/testthat/blob/1faa32fc1afa769ac16bdb5792fd597324143c57/tests/testthat/test-mock.R#L81-L87

I can't decide if this test should be deleted or modified, so will not make a PR to fix this.

hadley:
Fixed in #803

PR #821, opened 2018-11-27
Define a generic for identical() so custom methods can be utilized

### LINKED ISSUE #820: expect_identical() doesn't dispatch on identical() methods defined in other packages
The core comparison within `expect_identical()` is carried out by calling `identical(x, y)` ([code link](https://github.com/r-lib/testthat/blob/master/R/expect-equality.R#L107)).

Currently, since `identical()` is not a generic, `expect_identical()` always calls `base::identical()`. However, user might want to define some specific `identical()` method for specific objects. For instance, `RProtoBuf` package defines an `identical()` method for comparing two `Message`s ([code link](https://github.com/eddelbuettel/rprotobuf/blob/master/R/identical.R#L3)). Those user-defined methods are not visible to `testthat` and are not used (see [this issue](https://github.com/eddelbuettel/rprotobuf/issues/55) in `RProtoBuf` repo for context).

It would be great if `testthat` can define a generic on top of `identical()` so that it can dispatch to specific methods defined by other packages.

jarodmeng:
This fixes #820.

### DIFF
--- a/R/expect-equality.R
+++ b/R/expect-equality.R
@@ -97,6 +97,8 @@ expect_equivalent <- function(object, expected, ..., info = NULL, label = NULL,
   invisible(act$val)
 }
 
+setGeneric("identical")
+
 #' @export
 #' @rdname equality-expectations
 expect_identical <- function(object, expected, info = NULL, label = NULL,


jarodmeng:
The test fails because of #786.

hadley:
See discussion on issue.

PR #822, opened 2018-11-29
Use sha1 instead of digest to overcome precision errors

Bisaloo:
Currently `expect_known_hash` does not work on different computer / OS because of floating precision error.

According to [the digest vignette](https://cran.r-project.org/web/packages/digest/vignettes/sha1.html), in this case, we should use `sha1()` instead of `digest()`. This PR performs this change.

I can revise it if necessary! Also, I'm not sure under which section this should be added to `NEWS.md`. This change will require users to update their hash values.

### DIFF
--- a/R/expect-known.R
+++ b/R/expect-known.R
@@ -131,7 +131,7 @@ expect_equal_to_reference <- function(..., update = FALSE) {
 #'   to use in the test output.
 expect_known_hash <- function(object, hash = NULL) {
   act <- quasi_label(enquo(object))
-  act_hash <- digest::digest(act$val)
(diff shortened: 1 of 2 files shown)


Bisaloo:
Or a possibly better solution would be to add an `method`/`type` argument to `expect_known_hash()` so that users can pick the hash they want.

The upside is that it wouldn't break the existing tests.

hadley:
I think it has to be a new argument because this change would break all existing tests. Do you want to prepare a new PR? (I'll be working seriously on testthat in >1 month, so there's no rush)

Issue #781, opened 2018-07-26
describe() does not catch bare conditions


yonicd:
<details>

``` r
tf <- tempfile(fileext = '.R')

cat("testthat::context('a test')

fn <- function(x){
  as.numeric(x)
}

testthat::describe('invalid input',{
  
  val <- fn(c(1,'.'))
  
  it('create warning', {
    
    testthat::expect_equal(c(1,NA),val)
  })
  
})
",file = tf)

x <- testthat::test_file(tf)
#> ✔ | OK F W S | Context
#> 
⠏ |  0       | a test
⠋ |  0   1   | a test
⠙ |  1   1   | a test
✔ |  1   1   | a test
#> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
#> file33ed32a28722.R:7: warning: (unknown)
#> NAs introduced by coercion
#> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
#> 
#> ══ Results ════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
#> Duration: 0.1 s
#> 
#> OK:       1
#> Failed:   0
#> Warnings: 1
#> Skipped:  0

# 0 warnings in object output

x
#>                 file context                          test nb failed
#> 1 file33ed32a28722.R  a test invalid input: create warning  1      0
#>   skipped error warning  user system  real
#> 1   FALSE FALSE       0 0.003      0 0.003
```

Created on 2018-07-26 by the [reprex
package](http://reprex.tidyverse.org) (v0.2.0).

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> Session info -------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.5.0 (2018-04-23)
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  tz       America/New_York            
#>  date     2018-07-26
#> Packages -----------------------------------------------------------------
#>  package    * version date       source        
#>  assertthat   0.2.0   2017-04-11 CRAN (R 3.5.0)
#>  backports    1.1.2   2017-12-13 CRAN (R 3.5.0)
#>  base       * 3.5.0   2018-04-24 local         
#>  cli          1.0.0   2017-11-05 CRAN (R 3.5.0)
[...]

PR #823, opened 2018-11-30
List reporter error bug

kforner:
This pull request solves one problem, fixes a bug and removes one useless call:

- the problem is when an exception happens in-between tests. The exception is captured by the testthat framework, but is not reported by the ListReporter and the `test_file()` function . So that if you look only at the results as output by the `test_file()` function, you can not know that your test file has not been processed properly. The proposed solution is to record those exceptions, detect them in the end_context() method, and report them.

- the bug was when a NULL reporter is given in the `test_file()` function, cf  `R/test-files.R` line 110

- the useless call is `lister$start_file(basename(path))`, cf  `R/test-files.R` line 132


kforner:
N.B: the travis builds seemed to fail for another reason : `Function add_path not found in environment devtools.`

kforner:
N.B: the travis builds seemed to fail for another reason : `Function add_path not found in environment devtools.`

hadley:
Can you please merge/rebase to fix the failing test?

hadley:
Actually, let me me do that.

tests/testthat/test-reporter-list.R
```diff
@@ -1,25 +1,59 @@
+# regression test: test_file() used to crash with a NULL reporter
+test_that("ListReporter with test_file and NULL reporter", {
+  test_file_path <- 'test-list-reporter/test-exercise-list-reporter.R'
+  expect_error(test_file(test_path(test_file_path), reporter = NULL), NA)
+})
+
 test_that("ListReporter with test_file", {
```
hadley: This is a very large test. Do you think you could break it up into smaller pieces?

coatless:
Any chance this PR could be merged in? The `ListReporter` drops the `result` information if a function explicitly calls `stop()`.

This PR seems to address a bug with:

```r
fake_fun = function(n) {
  stop("error'd")
}

describe("works-fine", {
   it("should trigger results", {
    expect_equal(1, 1)
   })
})

describe("errors", {
   it("should error and causes no results to appear", {
    expect_equal(fake_fun(2), 1)
   })
})
```

```r
test_output = testthat::test_dir("tests/", reporter = testthat::ListReporter)
test_output$results
```

kforner:
I had completely forgotten about my PR. I'll try to split the tests so that it can get merged as soon as I can.

coatless:
@kforner great! I can't wait for this to be added. One less item for me to keep track of.

tests/testthat/test-reporter-list.R
kforner: How so ? The test is ~ 24 lines, and has a single call to `test_file()`.

hadley: It's sufficiently long that I can't follow exactly what it's doing.

hadley:
Looks good. Does this fix #781?

hadley:
Nope, looks unrelated.

hadley:
Thanks!

kforner:
You're welcome!

R/reporter-list.R
```diff
@@ -31,32 +31,65 @@ ListReporter <- R6::R6Class("ListReporter",
     },
 
     add_result = function(context, test, result) {
-      if (!is.null(self$current_expectations)) {
-        self$current_expectations$push(result)
+      if (is.null(self$current_expectations)) {
+        # we received an unexpected result: could be a bare expectation or an exception/error
+        if (!inherits(result, 'error')) return()
```
hadley: What's the point of this line?

hadley: I guess the intent is to not capture bare expectations, just errors, but I'm not sure why that's desirable?

kforner: Suppose that you have some outcode outside tests, e.g. to preprocess some test data. 
If that fails, you want your test script to crash so that you know there's something wrong and you can debug. 
Otherwise this could unnoticed, among the tests that failed.

hadley: Oh I see what you mean — we need to make sure an error outside of a test case causes the code to terminate.

kforner: Yes. And that probably should be handled somewhere, but not sure the ListReporter is the best place to do this, as this should probably be reporter-independent.

hadley: Filed an https://github.com/r-lib/testthat/issues/950

Issue #825, opened 2018-12-04
Support test skipping on CircleCI builds?


mbjoseph:
For automated builds running on CircleCI, it might be convenient to have a `skip_on_circle()` function to skip tests, similar to `skip_on_travis()`. By default, CircleCI builds export an environment variable `CIRCLECI` that takes on the value "true": https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

Is this something that folks would be interested in including in testthat?

jimhester:
Most CI systems set the `CI` environment variable, e.g. both [travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) and [CircleCI](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) do.

I think if we were going to add another skip to testthat itself it would be a general`skip_on_ci()` rather than another CI specific skip function.

mbjoseph:
Thanks @jimhester! That does seem like a better solution. 

I have a PR open (https://github.com/r-lib/testthat/pull/826) and would be happy to modify it to implement `skip_on_ci()` if you'd like.

jimhester:
Modifying the existing PR sounds good to me, thanks!

mbjoseph:
Great - I have updated the PR to implement `skip_on_ci()`, which checks to see whether the `CI` environment variable is either `"true"` (as with Travis and CircleCI) or `"True"` (as with AppVeyor Windows builds): https://github.com/r-lib/testthat/pull/826/files#diff-8118c4bf22b8b996a3b1d49ba32e01feR173

PR #826, opened 2018-12-04
Add skip_on_ci() function

mbjoseph:
This PR adds a `skip_on_ci()` function that uses the `CI` environment variable to skip tests running on continuous integration systems, and would solve https://github.com/r-lib/testthat/issues/825. 

* Edit: Generalized from `skip_on_circleci()` to `skip_on_ci()`


R/skip.R
```diff
@@ -163,6 +166,17 @@ skip_on_appveyor <- function() {
   skip("On Appveyor")
 }
 
+#' @export
+#' @rdname skip
+skip_on_ci <- function() {
+  ci <- Sys.getenv("CI")
+  if (!any(identical(ci, "true"), identical(ci, "True"))) {
```
jimhester:
```suggestion
  if (!isTRUE(as.logical(Sys.getenv("CI")))) {
```

jimhester: This is a potential shorter / more robust solution.

``` r
isTRUE(as.logical("true"))
#> [1] TRUE
isTRUE(as.logical("True"))
#> [1] TRUE
isTRUE(as.logical("False"))
#> [1] FALSE
isTRUE(as.logical(""))
#> [1] FALSE
isTRUE(as.logical(NULL))
#> [1] FALSE
isTRUE(as.logical(NA))
#> [1] FALSE
```

<sup>Created on 2018-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

mbjoseph:
Thank you @jimhester! I swapped in `isTRUE` as suggested, and added a NEWS bullet in 7ff05d4

hadley:
Thanks!

PR #827, opened 2018-12-04
Add check.attributes argument to compare.POSIXt

harvey131:
POSIXct objects created by some objects, such as the index of xts objects, contain additional attributes.  For example the POSIXct objects in the index() of an xts object will contain the 'tclass' attribute.  This is make it difficult to compare POSIXct objects.  Adding the check.attributes argument allows POSIXct objects to be compared by their numeric values.


hadley:
I don't think this is the right fix — rather than ignoring attributes, I think you should be generating POSIXct objects with the correct attributes via some helper.

Issue #750, opened 2018-04-27
expect_setequal doesn't work on lists (document?)


MichaelChirico:
I'd expect the following test to pass: `expect_setequal(list(a = 1:4, b = 4:1), list(b = 4:1, a = 1:4))`, but it fails because the inputs are sorted with `sort.int`, which doesn't know how to handle a list.

I was attempting to replace a test of the form `expect_equal(my_list[order(names(my_list))], list(a = 1:4, b = 4:1))`.

If there's disagreement about whether these objects should be considered `setequal`, at the very least we need to document the acceptable types (atomic?) to be passed to `object` as `?expect_setequal` only says:

`object` | object to test
-- | --

kevinykuo:
This feels like a bug; one would expect `expect_setequal()` to pass if `setequal` is `TRUE`.

kevinushey:
I agree this would be nice to have.

hadley:
Set equality ignores names:

``` r

setequal(
  list(a = 1:4, b = 4:1), 
  list(d = 4:1, e = 1:4)
)
#> [1] TRUE
```

<sup>Created on 2019-03-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

But maybe `expect_setequal()` should use an implementation closer to `setequal()` and avoiding sorting. We could using something like `all(vec_in(x, y)) && all(vec_in(y, x))`. The error reporting will be a bit trickier though.

PR #830, opened 2018-12-14
handle named R lists in expect_setequal (#750)

kevinushey:
Useful for when you're treating the two lists as property bags and order doesn't matter.


hadley:
Let's discuss this more first — I'm not 100% sure this is the right approach.

PR #847, opened 2019-02-13
 Fix bug that auto_test_package does not work at all

masaha03:
devtools::load_all requires not package object but path string.


jimhester:
Thanks, however this is already fixed by using the development version of devtools.

Issue #824, opened 2018-12-03
problem with the travis CI build system ? 


kforner:
All the Pull requests travis CI checks seem to fail for the same reason:
```
> test_check("testthat")
1. Error: can mock if package is not loaded (@test-mock.R#86)  
Function add_path not found in environment devtools.
1: with_mock(`devtools::add_path` = identity, expect_identical(devtools::add_path, identity)) at testthat/test-mock.R:86
```
which seems entirely unrelated to the PR themselves. 
So please do not disregard the PRs based on their travis checks...

hadley:
Fixed in #803

PR #849, opened 2019-02-16
Fix and add new tests (#786, #824)

### LINKED ISSUE #786: Remove test that mocks devtools::add_path()

patr1ckm:
This adds a few more tests, addressing some of the low coverage areas and untested branches from the codecoverage report.

Would more contributions in this vein be appreciated? It would be fun to tackle more of the low coverage areas, but it will probably require more testing code. 

Fixes #786, #824


patr1ckm:
The tests are now compatible with the new version of devtools.

## Fixed Errors
There was one new error, as noted in #786. This was fixed by mocking a different function from devtools instead of `devtools::add_path` which was removed. I chose `devtools::document` because it was unlikely to be removed.

## Fixed Warnings
There were two new warnings in `testthat/test_catch.R`. To fix I needed to do the following:

- Set `R_TESTS=''`. This fix may be a result in a change in behavior upstream but I'm not sure.
- Replace `devtools::create` with `usethis::create_package`. I added `usethis` to `Suggests` as a result.

Because I was unable to debug these warnings when the `test_catch` messages were suppressed, I left them on in case errors in `test_catch` surface again. However, I can make `test_catch` run quietly again if you prefer.

kevinushey:
The Catch-related test changes look fine to me.

hadley:
Thanks - can you please merge with changes to master? I think the conflicts come because we fixing the mocking failure in a different way to you.

hadley:
Thanks @patr1ckm !

PR #853, opened 2019-03-15
Clarify error message when hashes are not equal

chasemc:
Example of error message before:
`Value hashes to f6719f8ec8c does not hash not 2f971e94c2f`
Example of error message after:
`Value hashes to f6719f8ec8c, not 2f971e94c2f`


hadley:
Thanks!

PR #860, opened 2019-04-03
reverse dep checks

topepo:
We checked 4224 reverse dependencies (3753 from CRAN + 471 from BioConductor), comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 27 new problems
 * We failed to check 384 packages

Issues with CRAN packages are summarised below.

[`data.sqlite` file](https://www.dropbox.com/s/9lcc2tmkptdrcss/data.sqlite?dl=0)


hadley:
Thanks! I'm going to merge by hand since I'm also working on tweaks to revdepcheck

PR #866, opened 2019-04-17
Fixes for expect_mapequal

wch:
This fixes a few things in `expect_mapequal`. It now:

* Fails if there are any duplicated names.
* Fails if there are any unnamed values.
* Succeeds when comparing any length-0 vectors, whether they are named or unnamed.

### DIFF
--- a/R/expect-setequal.R
+++ b/R/expect-setequal.R
@@ -78,8 +78,31 @@ expect_mapequal <- function(object, expected) {
     abort("`object` and `expected` must both be vectors")
   }
 
+  # Length-0 vectors are OK whether named or unnamed.
+  if (length(act$val) == 0 && length(exp$val) == 0) {
+    succeed()
+  }
+
   act_nms <- names(act$val)
   exp_nms <- names(exp$val)
+
+  if (anyDuplicated(act_nms)) {
+    fail(
+      paste0("Duplicate names in `object`: ", unique(act_nms[duplicated(act_nms)]))
+    )
+  }
+  if (anyDuplicated(exp_nms)) {
+    fail(
+      paste0("Duplicate names in `expected`: ", unique(exp_nms[duplicated(exp_nms)]))
+    )
+  }
+  if (any(act_nms == "")) {
+    fail(paste0("All elements in `object` must be named"))
+  }
+  if (any(exp_nms == "")) {
+    fail(paste0("All elements in `expected` must be named"))
+  }
+
 
   if (!setequal(act_nms, exp_nms)) {
     act_miss <- setdiff(exp_nms, act_nms)
--- a/R/expect-setequal.R
+++ b/R/expect-setequal.R
@@ -82,28 +82,15 @@ expect_mapequal <- function(object, expected) {
   if (length(act$val) == 0 && length(exp$val) == 0) {
     warn("`object` and `expected` are empty lists")
     succeed()
+    return(invisible(act$val))
   }
 
   act_nms <- names(act$val)
   exp_nms <- names(exp$val)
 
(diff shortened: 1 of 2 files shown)


R/expect-setequal.R
```diff
@@ -78,9 +78,32 @@ expect_mapequal <- function(object, expected) {
     abort("`object` and `expected` must both be vectors")
   }
 
+  # Length-0 vectors are OK whether named or unnamed.
+  if (length(act$val) == 0 && length(exp$val) == 0) {
+    succeed()
```
hadley: `succeed()` doesn't return so you also need `return(invisible(act$val))` here.

But are you sure this shouldn't be an error? It seems unlikely to be what you actually want.

R/expect-setequal.R
```diff
@@ -78,9 +78,32 @@ expect_mapequal <- function(object, expected) {
...
+  }
+
   act_nms <- names(act$val)
   exp_nms <- names(exp$val)
 
+  if (anyDuplicated(act_nms)) {
```
hadley: Can you wrap this up into an `names_ok()` function? Also note that names can be `NA`, not just `""`.

R/expect-setequal.R
```diff
@@ -78,9 +78,32 @@ expect_mapequal <- function(object, expected) {
...
+    fail(
```
hadley: `fail()` doesn't return either, so you'll need to make sure that the code below still makes sense.

R/expect-setequal.R
wch: I'd expect this comparison to work (it's in the tests below):

```
x <- list()
x
#> list()

y <- list(a=1)[0]
y
#> named list()

expect_setequal(x, y)
```

And also:

```R
expect_setequal(list(), list())
```

When using a list as simple map or key-value store, I sometimes start with a `list()` and then populate it, and R automatically upgrades it to a named list. If that named list is then reduced to length 0, I'd expect this expectation to treat it as the same as an empty unnamed list.

kevinushey: I think it make sense -- we'd expect `expect_setequal(character(), character())` to succeed; this is a similar analogue I think. (I also don't think the distinction between an unnamed vector and an empty vector with 0-length names is meaningful in this context)

hadley: Yeah, I think `mapequal(list(), list())` should be `TRUE`, but it's not clear to me whether the same reasoning applied to `expect_mapequal()`. Because if you are comparing two empty lists, you really should be using something else. I think this would be at least worth a warning.

wch: Oh, I meant to write `expect_mapequal()`, not `expect_setequal()`. At any rate, I'm OK with a warning here.

R/expect-setequal.R
```diff
@@ -78,9 +78,20 @@ expect_mapequal <- function(object, expected) {
...
+    warn("`object` and `expected` are empty lists")
+    succeed()
+    return(invisible(act$val))
+  }
+
   act_nms <- names(act$val)
   exp_nms <- names(exp$val)
 
+  if (!names_ok(act_nms, "object") || !names_ok(exp_nms, "expected")) {
+    return(act$val)
```
hadley: Shouldn't this be a failure or error?

hadley: Oh you put the `fail` in `names_ok()` — I think you need to bring it back into this function

hadley: Rename `names_ok()` to `check_names_ok()` and have it call `stop()` rather than `fail()`. Then you can discard this `if` block.

hadley:
Thanks!

Issue #851, opened 2019-02-24
StopReporter doesn't stop


HughParsonage:
How can I run the test suite of a package so that any failure immediately halts so that the following tests do not run? 

My use-case is I have a package with a moderately long-running test suite with some 1000 tests. Often a failure occurs in the first dozen or so, but if I just run `devtools::test()` I have to wait till the other tests succeed before I can debug. 

I note that there is an argument `stop_on_failure` but this doesn't seem to do what I want. I'm guessing I could achieve this via a reporter but not sure if there is a better way.

jimhester:
I generally just interrupt the tests once a failure happens, but understand this is not ideal.

There is a `stop` reporter, e.g. `devtools::test(reporter = "stop")` but with limited testing it seems broken...

hadley:
It works ok interactively:

``` r
library(testthat)

with_reporter("stop", {
  test_that("this should fail", {
    expect_equal(1, 2)
  })
})
#> Error: Test failed: 'this should fail'
#> * 1 not equal to 2.
#> 1/1 mismatches
#> [1] 1 - 2 == -1
```

<sup>Created on 2019-03-31 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

But ` devtools::test(reporter = "stop")` doesn't fail.

hadley:
Hmmm, a little interactive browsing reveals that the `StopReporter` is throwing the error, so something must be silently eating it.

hadley:
I thought this might be an easy fix but it's something to do with the error being swallowed by the tooling that detects errors _inside_ tests. I'll need to come back to it when I'm deeper into testthat.

hadley:
Good reprex:

```R
testthat::test_file("tests/testthat/test_dir/test-failures.R", reporter = "stop")
```

hadley:
Ok, I have a hacky way to fix this, which is to throw a special type of error that terminates the test run early. But I'd rather not add even more complication to the code before we've reviewed it to make sure the current process mostly makes sense. That will happen as part of #729

PR #876, opened 2019-05-07
Tweaking skip messages

### LINKED ISSUE #868: tweak default messages for skip_if, skip_if_not
E.g.

```
> skip_if_not(FALSE)
Error: FALSE
```

It'd be nice if this showed something more like `Error: FALSE is not TRUE` or something with slightly more text.

As an example, the reported skip condition in `devtools::test()` is something like:

```
test-file.R:33: skip: we can do the thing
check_thing()
```

It would be a little more descriptive to instead write something like:

```
test-file.R:33: skip: we can do the thing
check_thing() is not TRUE
```

patr1ckm:
This addresses #868 by changes to the report messages in `skip`, `skip_if`, and `skip_if_not`. 

Examples of the new format are:

```
Reason: empty

tests.R:77: skip: Skip:4
Reason: condition is TRUE
```

The reference diff is a little noisier since I added a new test for `skip_if`.

EDIT: I'm skipping the `DebugReporter` output test unless the test is run interactively. The DebugReporter uses the package source to print relative debugging output. Since package source isn't available in R CMD CHECK, the test fails. 

Fixes #868 .


tests/testthat/test-reporter.R
```diff
@@ -29,7 +29,7 @@ test_that("reporters produce consistent output", {
   expect_report_unchanged("teamcity")
   expect_report_unchanged("silent")
   expect_report_unchanged("rstudio")
-  expect_report_unchanged("junit", reporter = createJunitReporterMock())
```
patr1ckm: This was refactored out earlier.

tests/testthat/reporters/debug.txt
```diff
@@ -50,10 +50,13 @@
 1: tests.R#72: test_that("Skip:3", {
 })
 
-1: tests.R#78: warning("abc")
+1: tests.R#77: skip_if(condition)
+2: skip.R#86: skip(message)
```
patr1ckm: This is the line that produced different output in `R CMD CHECK`. That showed only

`2: skip(message)`
rather than
`2: skip.R#86: skip(message)`

I'm still not sure what the bug is, but I think it may be an issue in `recover2`, which seems out of scope for this PR.

hadley: I don't see this failure, either locally or on travis?

hadley: Oh hmmm, seems like there's some problem with the reporter tests. I'll look into it

hadley: Maybe this is #851?

hadley: Ok, now I see the problem

hadley: For now the easiest thing is just to remove that new skip. There are easier ways to test the output.

PR #887, opened 2019-06-18
Unify as.data.frame structure for empty testthat_results

jozefhajnala:
This PR proposes to unify the "return type" of `as.data.frame.testthat_results()` to a data frame with 13 columns with predefined names, regardless of whether the input object is of length 0.

**Context and motivation**

I often use testthat results in a CI context, where combining (i.e. rbinding) the results of tests for multiple packages into 1 data frame is quite useful. It happens sometimes that one of those objects is empty, in which case `as.data.frame(testResults)` returns a `0 x 0` data frame, requiring further manipulation if we want to apply unified methods.


hadley:
Thanks!

PR #890, opened 2019-06-25
Fixes the end_context() call in test_file()

### LINKED ISSUE #883: Crash with "Error in private$report_event"
With versions 2.1.0 and 2.1.1 I get a crash when invoking `devtools::test(reporter = testthat::TeamcityReporter)`:

```
Error in private$report_event("testSuiteFinished", context) :
  argument "context" is missing, with no default
Calls: <Anonymous> ... o_apply -> lapply -> FUN -> <Anonymous> -> <Anonymous>
```

It didn't happen before.
It doesn't depend on R version (it's both with R 3.4 and 3.6), both on FreeBSD and Linux.

atheriel:
Commit 25e7d930 replaced the call to `end_context()` in `test_file()` with `reporter$end_context()`. However, the `end_context()` function in fact calls the equivalent of `reporter$.end_context()` (note the dot).

The `.end_context()` version, it turns out, is responsible for handling the stack of contexts, and without it `reporter$end_context()` will have a missing `context` argument.

As I mentioned in #883, the only reporter that actually uses the `context` parameter is the TeamCity one, which is perhaps why it took a while to notice the problem.

This commit simply switches back to using `reporter$.end_context()`.

Fixes #883.

### DIFF
--- a/R/test-files.R
+++ b/R/test-files.R
@@ -152,7 +152,7 @@ test_file <- function(path,
         chdir = TRUE, wrap = wrap
       )
 
-      reporter$end_context()
+      reporter$.end_context()
       reporter$end_file()
     }
   )
--- a/tests/testthat/test-reporter.R
+++ b/tests/testthat/test-reporter.R
@@ -62,7 +62,6 @@ test_that("reporters accept a 'file' argument and write to that location", {
   expect_report_to_file(TapReporter)
   expect_report_to_file(TeamcityReporter)
   expect_report_to_file(RstudioReporter)
-  expect_report_to_file(SilentReporter)
 })
 
 test_that("reporters write to 'testthat.output_file', if specified", {


atheriel:
The failing test is that the `SilentReporter` no longer produces an output file.

atheriel:
Something I don't quite understand is going on. The test suite prior to this patch reports:

> (snip)
> ✔ |   9       | reporter [0.6 s]
> (snip)

and after this patch I see:

> (snip)
> ✖ |  11 1     | reporter [0.6 s]
> (snip)

In other words, **more tests are running (and passing)** than before. Is it possible that errors in `expect_report_to_file()` at

```r
test_file(test_path("reporters/tests.R"), reporter, wrap = FALSE)
```

are actually being swallowed?

atheriel:
I have fixed what I believe to be the offending test issues.

jimhester:
Calling a private method does not seem correct? maybe using the `end_context()` free function is a better approach? I believe it fixes the problem is mostly the same way.

```diff
diff --git a/R/test-files.R b/R/test-files.R
index 2a9f21a..3ca93a4 100644
--- a/R/test-files.R
+++ b/R/test-files.R
@@ -152,7 +152,7 @@ test_file <- function(path,
         chdir = TRUE, wrap = wrap
       )

-      reporter$end_context()
+      end_context()
       reporter$end_file()
     }
   )
```

atheriel:
Happy to do that, too. Alternatively, you could add a public method to retrieve the current context so that the reporter could be called directly, e.g.

```diff
diff --git a/R/reporter.R b/R/reporter.R
index 14d00b0..a018f54 100644
--- a/R/reporter.R
+++ b/R/reporter.R
@@ -83,6 +83,11 @@ Reporter <- R6::R6Class("Reporter",
         self$.context <- NULL
       }
       invisible()
+    },
+
+    # Public accessor for the current context.
+    context = function() {
+      self$.context
     }
   )
 )
diff --git a/R/test-files.R b/R/test-files.R
index 45ae59d..daa4544 100644
--- a/R/test-files.R
+++ b/R/test-files.R
@@ -152,7 +152,7 @@ test_file <- function(path,
         chdir = TRUE, wrap = wrap
       )

-      reporter$end_context()
+      reporter$end_context(reporter$context())
       reporter$end_file()
     }
   )
```

It's up to you. I'll rewrite the commit as needed.

atheriel:
@jimhester Anything I can do to move this forward?

atheriel:
@hadley Thanks!

PR #892, opened 2019-06-29
Background continuous automated test for package + addin

CorradoLanera:
For better TDD run test continuously is a great option. I have just discovered `testthat::auto_test_package()`. Include an addin could be useful. On the other hand, a busy section while developing a package can hurt (especially using **usethis**' functions to create scripts, tests, etc.). I thought that run continuously the test in a background job can be a better solution.

A function to stop the job is also provided. I did not found a way to do it on the open-source version of RStudio. Anyway, the closing function offers some useful suggestions on how to close it manually (if it cannot close the job automatically).

I tried to replicate a little of the color of `usethis::ui_*` without **usethis**, because it is a suggested package, and I prefer not to change it as imported.

RStudio 1.2+ is required. I have put **rstudioapi** under suggests requiring it in the function definition because I think these are "addin" functions, and **rstudioapi** is not needed for all the users of **testthat** (as `auto_test_package()` did for *devtools*).


CorradoLanera:
I did some refactoring and added more tests.
Anyway, I do not know how to reach the coverage required.

CorradoLanera:
I have also refactored `auto_test` and `auto_test_package`, and write some test for them. Now the coverage thresholds required are both reached too :-)

CorradoLanera:
I have just noticed these three "last" commits (ffa4e88, a7e6c74, and 68630f2) which I had committed before the d445640. I had an issue with my system clock in the meanwhile. So, I suppose they have attached the wrong time label. Anyway, the last commit is correctly the green checked one: eabde1d :+1:

hadley:
I'm now generally sceptical about the utility of `auto_test()` so I'd prefer to not include more work on it in testthat. You've obviously done a bunch of work on it, so I'd suggest that it's probably enough to make a package in its own right.

CorradoLanera:
Perfect. I did it [here](https://github.com/CorradoLanera/autotestthat). Thank you for the suggestion.

PR #895, opened 2019-07-09
Implement skip_on_covr()

ianmcook:
This implements `skip_on_covr()` which skips tests when covr is running by inspecting the `R_COVR` environment variable. Thanks!


jimhester:
Thanks, I think the code looks good! I am just curious as to the use cases when you need to skip tests when running with covr? Is it because the tests fail with covr and not without or is it some other reason?

ianmcook:
Thanks Jim! Yes, I ran into a case where tests failed with covr immediately after succeeding with testthat. I was using a live Impala instance to run the test so I don’t have a self-contained reprex handy but I’ll see if I can create one, and I’ll create a covr issue with the details if I can.

gaborcsardi:
@jimhester FWIW I am skipping these in covr, but it is not a typical use case:
https://github.com/r-lib/callr/blob/b2d8a58a6409af3c6c3bfb91eb021c2048c6b3f8/tests/testthat/test-clean-subprocess.R

ianmcook:
@jimhester here's the underlying covr issue that caused me to need to skip some tests: https://github.com/r-lib/covr/issues/379

hadley:
Thanks!

Issue #896, opened 2019-07-12
auto_test_package do not update helpers when a test is changed


CorradoLanera:
If you modify a script, all tests (and helpers) are run.
If you modify a test, only that test is run, correctly.
The problem rises if you create a helper file to extract some preprocessing from a test.

When you create the helper it is run, it contains no test, and the results are all green zeros, correctly. Next, when you return to the test function, and you remove the no more necessary code, only this test is run, but not the helper, and it rises a (wrong) error.

On the other hand, the first script you modify later provide a run of the whole test folder, helper included, and from that time on the objects are stored in the test environment, and so everything return working. The problem occurs only the first time.

example:

1. Create `foo.R` in `R/`:
```
foo <- function(x) x
```

2. Create a `test-foo.R`
```
a <- 1
expect_equal(foo(a), 1)
```
All the test pass!

3. Create `helper-foo.R` 
```
a <- 1
```
The "test" `helper-foo.R` is runned reporting no test and no erorr :-)

4. Remove assignment from `test-foo.R` which becomes simply
```
expect_equal(foo(a), 1)
```
And now the test report an error
> object 'a' not found

5. Make any modification to any script, for example, add curly braces to `foo.R`, just to save it and trig the call to the tests
```
foo <- function(x) {
    x
}
```

And everything is evaluated, and all the test passed.

6. Even if you modify `test-foo.R`, e.g.
```
expect_equal(foo(a), 1)
expect_equal(foo(a), 1)
```
And only this test is called and run, from now it passes correctly. Until you change the helper:

7. Modify `helper-foo.R`
```
a <- 2
```
The "test" `helper-foo.R` is run reporting no test and no errors

8. Modify `test-foo.R` accordingly
```
expect_equal(foo(a), 2)
```
And the test is run reporting an error:
> 'a' not equal to 2
> 1/1 mismatches
> [1] 1 - 2 == -1"

Adopting a solution like the one purposed in #376 could help.

Issue #376, opened 2016-02-23
if a helper file is changed, auto_test_package should re-run tests


jackwasey:
Currently, auto_test_package notices that the helper-* files have changed, and appears to source them, but doesn't run any tests. It would be clever if the dependencies between the contents of the helper files and the tests could be inferred, but at a first approximation, all tests should be re-run.

hadley:
I'd be happy to review a PR for this, but since I don't use the autotest stuff, I'm not going to implement it

PR #897, opened 2019-07-12
fixes autotest when helpers change (solve #376 and #896)

CorradoLanera:
I separated the cases when changes in test folder come from files starting with "helper-" and replicate the same behavior and code-structure used for changes in the code folder (modifying the message printed accordingly).

I do not know how to automatize some test for the auto_test functions. I tried some interactive test (like the one described in #896) and it works fine.

### DIFF
--- a/R/auto-test.R
+++ b/R/auto-test.R
@@ -103,21 +103,16 @@ auto_test_package <- function(pkg = ".", reporter = default_reporter(), hash = T
 
     tests <- changed[starts_with(changed, test_path)]
     helper <- tests[starts_with(basename(tests), "helper-")]
-    tests <- setdiff(tests, helper)
     code <- changed[starts_with(changed, code_path)]
+
+    # Remove helper from test and add it to code (if a helper changed,
+    # like for code, reload all and rerun all tests)
+    tests <- setdiff(tests, helper)
+    code <- c(code, helper)
 
     if (length(code) > 0) {
       # Reload code and rerun all tests
       cat("Changed code: ", paste0(basename(code), collapse = ", "), "\n")
-      cat("Rerunning all tests\n")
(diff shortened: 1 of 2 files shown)


R/auto-test.R
```diff
@@ -113,6 +115,15 @@ auto_test_package <- function(pkg = ".", reporter = default_reporter(), hash = T
         devtools::r_env_vars(),
         test_dir(test_path, env = env, reporter = reporter$clone(deep = TRUE))
       )
+    } else if (length(helper) > 0) {
```
hadley: Is this this same code as the block above? If so, I think a simpler solution would be to add `helper` to the `code` vector.

CorradoLanera: Almost all: the "active" code is exactly the same; the message changed, reporting "Changed helper: " instead of "Changed code: ". If that is not a relevant "feature", your solution seems better for me too.

I have implemented your purpose, so you can choose what you prefer to commit.
(I have also added a little "why" above the update of `test` and `code`)

R/auto-test.R
```diff
@@ -103,9 +103,13 @@ auto_test_package <- function(pkg = ".", reporter = default_reporter(), hash = T
 
     tests <- changed[starts_with(changed, test_path)]
     helper <- tests[starts_with(basename(tests), "helper-")]
-    tests <- setdiff(tests, helper)
     code <- changed[starts_with(changed, code_path)]
 
+    # Remove helper from test and add it to code (if a helper changed,
+    # like for code, reload all and rerun all tests)
+    tests <- setdiff(tests, helper)
```
hadley: Can you please move `helper <- ...` into this block as well? I think that makes it all hang together better

CorradoLanera: Sure, just did it!

hadley:
Thanks!!

Issue #904, opened 2019-07-17
What license does this library have?


nalzok:
The current LICENSE file isn't clear, and it's outdated. 

https://github.com/r-lib/testthat/blob/de6a87033021326a521c8db0483d0654e33e5eed/LICENSE#L1-L2

I notice some GPL-licensed files in this repository, and since the GPL is infectious, I suppose this library is licensed under a GPL too?

gaborcsardi:
https://github.com/r-lib/testthat/blob/de6a87033021326a521c8db0483d0654e33e5eed/DESCRIPTION#L13

nalzok:
Cool, thanks for the reply!

oxr463:
@nalzok, other than [`R/stack.R`](https://github.com/r-lib/testthat/commit/53b5b635f298de1b52e12fe24265b07c0f689466#diff-a3dd38952aa234e2fb229f7484ebf704), what files are licensed under the GPL?

hadley:
@oxr463 the original is GPL-3 licensed, but it's relicensed as MIT with permission of the original authors.

PR #905, opened 2019-07-18
WIP: License audit

oxr463:
Bug: https://github.com/r-lib/testthat/issues/904
See also: https://opensource.stackexchange.com/q/8537/8868


jimhester:
The license is correct as written as required for MIT licensed packages included in the [Comprehensive R Archive Network](https://cran.r-project.org/web/licenses/MIT). This is not something we can change.

hadley:
@jimhester but we can add a `.md`

PR #906, opened 2019-07-18
Implement verify_output()

### LINKED ISSUE #782: expect_known_results
I've noticed that I've started writing output regression tests like this:

```R
  expect_known_output(
    file = test_path("test-type-df.txt"),
    {
      cat("mtcars:\n")
      print(vec_type(mtcars))
      cat("\n")
      cat("iris:\n")
      print(vec_type(iris))
    }
  )
```

This is starting to feel like a poor man's knitr, which made me realise we need `expect_known_something()` that uses evaluate to show both the input code and the output results.

This would subsume #681 (assuming we added crayon options) and #768.

### LINKED ISSUE #834: Regression versus failure
This is about better support for tests that should not cause R CMD check failures because they are testing output that is not under the developer's control, at least not entirely. For instance, rlang has a lot of `expect_known_output()` tests for the backtraces. These outputs sometimes change across R versions, e.g. because `eval()` produces a different backtrace. Another case is vdiffr tests, because the appearance of plots is sensitive to upstream code such as computation of margins or spacing between elements.

Such tests should fail on platforms where the `CI` envvar is set (Travis, Appveyor), or where the `NOT_CRAN` envvar is set (tests run locally). This allows the developer to monitor and assess regressions during development.

@hadley suggested calling these tests regression tests. They could be implemented with a new expectation class `expectation_regression`.

hadley:
Goal is to create a new regression test focussed on human readable output

Fixes #782. Fixes #834.


DESCRIPTION
```diff
@@ -33,7 +34,8 @@ Suggests:
     rmarkdown,
     usethis,
     vctrs (>= 0.1.0),
-    xml2
+    xml2,
+    testthat (>= 2.1.0)
```
jimhester: This may not hurt anything, but it is kind of weird for testthat to Suggest itself.

tests/testthat/test-verify-output.R
```diff
@@ -0,0 +1,27 @@
...
+    stop("Error")
+
+    "Without calls"
+    warning("Warning", call. = FALSE)
+    stop("Error", call. = FALSE)
+  })
+})
+
+test_that("can't record plots", {
+  expect_error(verify_output("path", plot(1:10)), "Plots")
```
jimhester: This test opens a plot window, not sure the best way to avoid this and still run this test.

tests/testthat/test-verify-output.R
```diff
@@ -0,0 +1,27 @@
+test_that("can record all types of output", {
+  verify_output(test_path("test-verify-output.txt"), {
+    "Output"
+    1 + 2
+    invisible(1:10)
+    12345678 + 12345678 + 12345678 + 12345678 + 12345678 + 12345678 +
```
jimhester: Setting the width option must need to be moved slightly, as with a larger width option this output is different.

DESCRIPTION
hadley: Ooops, I guess we need to special case `use_test()`

PR #921, opened 2019-08-07
Add backtraces

### LINKED ISSUE #729: Consider traceback of expectation failures
And minimise impact of quasiquotation as much as possible.

lionel-:
`expectation()` and `expect()` gain a `trace` argument. If set to an rlang backtrace created with `trace_back()`, the expectation is printed with a simplified backtrace (through the `format()` method). This way any expectation can easily elect to print a backtrace as part of the results.

A backtrace is captured when:

* Unexpected errors are caught in `test_code()`.
* Errors and conditions captured by `expect_error()` and `expect_condition()` do not conform to expectations (closes #729).

The uninteresting testthat stack frames are trimmed via three mechanisms:

* From the top, the `rlang_trace_top_env` option is set before evaluating unit tests to prevent inclusion of the deep testthat call stack in the backtrace.

* From the bottom, the signalling context created by R (frames like `.handleSimpleError()`) is trimmed by calling the newly exported `rlang::cnd_entrace()` function from a calling handler.

* There are still many testthat frames included in backtraces (those belonging to `expect_is()`, `expect_error()`, etc), but these are not part of the terminal branch of backtrace trees. These frames are simpliy removed by formatting backtraces with `simplify = "branch"`.

The trimmed backtraces are currently printed in full, but we could omit the middle part by supplying `max_frames` to `format(trace)`. Edit: Now done.

`expect_warning()` and `expect_message()` currently don't display backtraces because there might be multiple non-conforming conditions and this doesn't fit well with the current design.


lionel-:
Live demo: https://travis-ci.org/r-lib/testthat/jobs/568871290#L812

```
── 1. Error: watcher works correctly (@test-watcher.R#87)  ─────────────────────
The specified file is not readable: /tmp/watcher2ddf140c7177/R/test1.R
Backtrace:
 1. testthat::watch(c(code_path, test_path), test.removed)
 2. testthat:::dir_state(path, pattern, hash = hash)
 3. base::vapply(files, safe_digest, character(1))
 4. testthat:::FUN(X[[i]], ...)
 5. digest::digest(path, file = TRUE)
 6. digest:::.errorhandler(...)
```

Not sure what this is about.

R/capture-condition.R
```diff
@@ -1,3 +1,25 @@
...
+}
+new_capture <- function(class) {
+  exiting_handlers <- rep_named(class, list(identity))
+
+  calling_handlers <- rep_named(class, alist(function(cnd) {
+    cnd <- cnd_entrace(cnd)
+    return_from(env, cnd)
+  }))
+
+  bang(function(code, entrace = FALSE) {
```
hadley: I think this might be simpler with `new_function()`, since I think the environment of the generated function should be the package env, not an execution environment of `new_capture()`.

R/capture-condition.R
```diff
@@ -1,3 +1,25 @@
...
+    if (!entrace) {
+      return(tryCatch({ code; NULL }, !!!exiting_handlers))
+    }
+
+    env <- environment()
```
hadley: I think this would be simpler to follow if you required `class` to be length one so you could use `!!class := function(cnd) ... `

R/expectation.R
```diff
@@ -86,12 +90,25 @@ format.expectation_success <- function(x, ...) {
 
 #' @export
 format.expectation_error <- function(x, ...) {
-  paste(c(x$message, create_traceback(x$call)), collapse = "\n")
+  format_with_trace(x)
```
hadley: Can `create_traceback()` now be deleted?

R/test-that.R
```diff
@@ -152,6 +155,9 @@ test_code <- function(test, code, env = test_env(), skip_on_empty = TRUE) {
   }
 
   test_env <- new.env(parent = env)
+  old <- options(rlang_trace_top_env = test_env)[[1]]
```
hadley: Can the `frame_calls()` code above now be deleted?

tests/testthat/reporters/check.txt
```diff
@@ -14,38 +14,46 @@ FALSE isn't true.
 
 ── 5. Error: Error:1 (@tests.R#33)  ────────────────────────────────────────────
 stop
-1: stop("stop") at reporters/tests.R:33
+Backtrace:
+ 1. NULL
```
hadley: I'm surprised to see a `NULL` in a backtrace

tests/testthat/reporters/check.txt
```diff
@@ -14,38 +14,46 @@ FALSE isn't true.
...
-20: f(x - 1) at reporters/tests.R:54
-21: f(x - 1) at reporters/tests.R:54
-22: f(x - 1) at reporters/tests.R:54
-23: f(x - 1) at reporters/tests.R:54
-24: f(x - 1) at reporters/tests.R:54
-25: f(x - 1) at reporters/tests.R:54
-26: f(x - 1) at reporters/tests.R:54
-27: stop("This is deep") at reporters/tests.R:54
+Backtrace:
+  1. f(25)
```
hadley: I think we should trim this. Let's keep it at 20 frames?

tests/testthat/test-reporter.R
```diff
@@ -55,3 +61,8 @@ test_that("reporters write to 'testthat.output_file', if specified", {
   test_file(test_path("reporters/tests.R"), MinimalReporter$new(), wrap = FALSE)
   expect_true(file.exists(path))
 })
+
+test_that("backtraces are reported", {
```
hadley: Does this really need to be tested with multiple reporters?

tests/testthat/test-reporter.R
```diff
@@ -1,3 +1,9 @@
+
+# Disable srcrefs because they differ across systems
+withr::local_options(list(rlang_trace_format_srcrefs = FALSE))
```
hadley: Could you take a look at trimming the srcref path when it's a child of the working directory? (i.e. at least file an issue in rlang)

R/capture-condition.R
lionel-: We need to allow deep name-unquoting in rlang first.

R/expectation.R
lionel-: There's still some work on that front because the current mechanism of capturing calls is tied to a reporter that gets you into `recover()`.

tests/testthat/test-reporter.R
lionel-: Sure, but in any case we need to disable the srcrefs in unit tests, otherwise we'll get failures when the pointed code is updated, and that can be in other packages like rlang.

lionel-: Also Travis will typically not have srcrefs for packages like rlang, while they may be present locally.

tests/testthat/reporters/check.txt
lionel-: Fixed by using `seq2()` instead of `:` https://github.com/r-lib/rlang/commit/a334055b

tests/testthat/test-reporter.R
lionel-: We now format srcrefs relative to `TESTTHAT_DIR`, see https://github.com/r-lib/testthat/pull/921/commits/7eb3640a654ebaf0871135668b8b5fee4cd139b4.

lionel-:
@hadley I've added a `skip_on_cran()` in reporter tests so we don't get failures when the formatting of backtraces change upstream. However that's really a case where it'd be better to let the code actually run so we catch unexpected errors. Having `verify_that()` blocks, or perhaps `local_allow_failures()` would be useful here.

R/capture-condition.R
hadley: Ooooh, bummer.

R/test-that.R
```diff
@@ -38,30 +38,26 @@ test_code <- function(test, code, env = test_env(), skip_on_empty = TRUE) {
   }
 
   ok <- TRUE
-  register_expectation <- function(e) {
-    calls <- e$expectation_calls
+  register_expectation <- function(e, debug_end) {
```
hadley: Could you add a comment explaining what `debug_end` is?

lionel-: :white_check_mark:

R/expectation.R
lionel-: Now removed, as well as `frame_calls()`.

tests/testthat/reporters/check.txt
```diff
@@ -14,38 +14,38 @@ FALSE isn't true.
...
```
jimhester: Why are we losing the location of the call? I realize these are noisy, but it is often useful information, and I typically do have the source references for all packages I compile on my machine (I have `R_KEEP_PKG_SOURCE=true` in my `.Renviron`).

tests/testthat/test-reporter.R
```diff
@@ -1,3 +1,12 @@
...
+
+skip_if(getRversion() < "3.4", "Fails because of new `eval()` call structure")
```
jimhester: Does this mean testthat now depends on R 3.4+? What happens if this code is run on R 3.2 or 3.3?

tests/testthat/test-reporter.R
jimhester: I think we should probably figure out a way to enable source references in at least a limited set of tests, maybe constructing them by hand or similar, even if most of the tests need it to be disabled due to the above issues.

lionel-:
Sorry Jim I missed your comments, looking at them now.

R/capture-condition.R
```diff
@@ -6,6 +32,8 @@
 #' [expect_warning()], and [expect_silent()].
 #'
 #' @param code Code to evaluate
+#' @param entrace Whether to add a [backtrace][rlang::trace_back] to
```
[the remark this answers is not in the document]
lionel-: I don't think we should refrain from neologisms. Maybe `rlang::add_trace()` would have been better, but now that we have `rlang::entrace()` we should probably be consistent?

tests/testthat/reporters/check.txt
lionel-: Only in the tests because foreign srcrefs now leak into the backtraces (e.g. from `rlang::eval_bare()`. The srcrefs are disabled with:

```r
withr::local_options(list(rlang_trace_format_srcrefs = FALSE))
```

tests/testthat/test-reporter.R
lionel-: Nothing bad happens, the output is just different. I would prefer to still run the code to detect unexpected errors, see comment in https://github.com/r-lib/testthat/pull/921#issuecomment-519494083.

lionel-: (oops the comment is about `skip_on_cran()`, not the skip on earlier versions)

PR #924, opened 2019-08-16
Output Junit timestamps in ISO 8601 and use all three decimal places 

jimhester:
There was a bug previously where the times were rounded to 2 rather than 3 decimal places.


jimhester:
lol, nice

PR #928, opened 2019-08-19
Abort when `Stop` reporter encounters a failure

### LINKED ISSUE #851: StopReporter doesn't stop

lionel-:
Closes #851

* Adds srcref to bullet list so location of failures can be traced.
* Invokes abort restart to shut down testthat.

```r
testthat::test_file("tests/testthat/test_dir/test-failures.R", reporter = "stop")
#> Test failed: 'just one failure'
#> * test-failures.R:4: FALSE isn't true. 
```

### DIFF
--- a/R/reporter-stop.R
+++ b/R/reporter-stop.R
@@ -33,9 +33,15 @@ StopReporter <- R6::R6Class("StopReporter",
       self$failures$initialize()
 
       messages <- vapply(failures, format, character(1))
-      messages <- paste0("* ", messages, collapse = "\n")
+      locations <- vapply(failures, exp_location, character(1))
+      messages <- paste0("* ", locations, ": ", messages, collapse = "\n")
+      message <- paste_line(
+        paste0("Test failed: '", test, "'"),
+        !!!messages
+      )
 
-      stop("Test failed: '", test, "'\n", messages, call. = FALSE)
+      cat(message, "\n", file = stderr())
+      invokeRestart("abort")
     },
 
     add_result = function(context, test, result) {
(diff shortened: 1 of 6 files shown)


R/reporter-stop.R
```diff
@@ -33,9 +33,15 @@ StopReporter <- R6::R6Class("StopReporter",
...
+      locations <- vapply(failures, exp_location, character(1))
+      messages <- paste0("* ", locations, ": ", messages, collapse = "\n")
+      message <- paste_line(
+        paste0("Test failed: '", test, "'"),
+        !!!messages
+      )
 
-      stop("Test failed: '", test, "'\n", messages, call. = FALSE)
+      cat(message, "\n", file = stderr())
+      invokeRestart("abort")
```
hadley: Doesn't this immediately jump to the top level? Is that too aggressive?

lionel-: I don't know the usage of the `Stop` reporter but for interactive applications this seemed appropriate. We can also throw a special class that we'd rethrow from `handle_fatal()`, this way it could be caught again from outside testthat.

hadley: I think throwing a special class might be safer. Otherwise how do we test it?

lionel-: We can add another abort restart on the stack to test it. No worries for a special class.

lionel-: `with_reporter()` now installs a `testthat_abort_reporter` restart that is invoked by the stop reporter.

PR #932, opened 2019-09-01
Added documentation on stop_on_failure default (fixes #912)

### LINKED ISSUE #912: question on stop_on_failure default
I see the following in `?testthat::test_dir`

image

Why is the default for `stop_on_failure` in `test_dir()` `FALSE` when it is `TRUE` for `test_package()` and `test_check()`? Is that intentional or an oversight?

In my opinion, the default behavior of a testing framework should be to fail (cause a non-zero exit code) if any tests fail. If you agree with me and this is an oversight, please let me know and I'd be happy to make a PR to change it.

Thanks!

jameslamb:
See #912 for background on this PR. This PR is intended to implement the suggestion in [this comment](https://github.com/r-lib/testthat/issues/912#issuecomment-524836631).

Thanks for considering it.

### DIFF
--- a/R/test-directory.R
+++ b/R/test-directory.R
@@ -41,6 +41,19 @@
 #' test_check("yourpackage")
 #' ```
 #'
+#' @section A note on function defaults:
+#' The behavior of `test_dir()`, `test_package()`, and `test_check()` when failing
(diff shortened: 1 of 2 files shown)


hadley:
Thanks! I condensed it a bit and moved it to the argument, where I think it's more likely to be seen.

PR #937, opened 2019-09-23
Fix abort restart on interactive invokation

lionel-:


jimhester:
Can you provide any context for this, someone looking at this PR or the code in the future has no way of knowing the context here.

In particular a test for `expect_true(FALSE)` deserves some comments I think.

tests/testthat.R
```diff
@@ -1,3 +1,9 @@
 library(testthat)
 options(testthat.use_colours = FALSE)
+
+test_that("can run test interactively", {
```
hadley: Could replace interactively with "outside of test reporter"

lionel-:
I have restored the behaviour back to what it was before the stop-reporter changes. Running a failing testthat block is an error again to avoid a breaking change.

@jimhester The test documentation now mentions "outside of test reporter", following Hadley's suggestion.

PR #955, opened 2019-11-02
Tweak documentation of verify_output()

krlmlr:
Examples and details.


hadley:
This doesn't feel quite right because I think the code should (almost) always be a braced expression

PR #957, opened 2019-11-04
Export `is_informative_error()` generic

lionel-:


R/expect-condition.R
```diff
@@ -174,22 +174,56 @@ cnd_message <- function(x) {
...
+#' developing an error hierarchy for your package. Override
+#' `is_informative_error()` to return `FALSE` to avoid encouraging
+#' users to depend on the experimental class in their tests.
+#'
+#' Since testthat should be a `Suggest` dependency, methods for
+#' `is_informative_error()` should typically be lazily registered,
+#' e.g. with `vctrs::s3_register()`.
+#'
+#' @param x An error object.
+#' @inheritParams ellipsis::dots_empty
```
hadley: `@keywords internal`

lionel-: Now internal.

PR #966, opened 2019-11-18
Fix rJava conditions

lionel-:
The `$` and `$<-` methods of rJava's `Throwable` condition class prevent read/write access to unknown condition fields. To fix this:

- Prevent `rJava` conditions from being entraced.
- Use `[[` to access fields.

Unit tests cover captured and uncaptured errors with a fake Throwable class. This way we won't reintroduce `$` or `$<-` calls in future updates.

Fixes R CMD check with the corehunter and SqlRender packages. I couldn't test tabulizer because of incompatible rJava versions but I assume it has been fixed as well.


hadley:
Can you add a bullet to NEWS?

PR #975, opened 2019-12-13
adds skip_on_gh_actions (resolves #970)

### LINKED ISSUE #970: Add a skip on GitHub actions function
Apologies if you've already implemented this and I missed it but I think testthat needs a `skip_on_actions` function like this:

```
skip_on_actions <- function() {
  if (!identical(Sys.getenv("GITHUB_ACTIONS"), "true")) {
    return(invisible(TRUE))
  }
  skip("On GitHub Actions")
}

```

I can submit a PR if that would be desired.

njtierney:
Completely stolen the code from https://github.com/r-lib/testthat/issues/970, but renamed function to `skip_on_gh_actions()` instead of `skip_on_actions()` - please feel free to change!

Thanks!


jimhester:
LGTM

tests/testthat/test-skip.R
```diff
@@ -25,6 +25,7 @@ test_that("Skip env vars", {
   expect_skip_with_env(c("APPVEYOR" = "True"), skip_on_appveyor)
   expect_skip_with_env(c("R_COVR" = "true"), skip_on_covr)
   expect_skip_with_env(c("BBS_HOME" = "asdf"), skip_on_bioc)
+  expect_skip_with_env(c("GITHUB_ACTIONS" = "asdf"), skip_on_gh_actions)
```
jimhester: I think this needs to be "true", or else the test will fail. In the skip you are checking the value of the environment variable is `"true"` This is why it is currently failing on the CI.

NEWS.md
```diff
@@ -9,6 +9,7 @@
 * Fixed issue where error and warning messages were not retrieved with
   `conditionMessage()` under certain circumstances.
 
+* Add `skip_on_gh_actions()` helper, to skip tests on GitHub Actions (#970) (@boshek, @njtieney).
```
hadley: Can you please merge/rebase and move the NEWS bullet up?

coatless:
@njtierney outside of rebasing, could you add the `skip_on_gh_actions()` function to the `details` section of the documentation so that it is alongside the other skip variants? (I stumbled across this looking for the relevant `skip_on_*()`. 

https://github.com/r-lib/testthat/pull/975/files#diff-8118c4bf22b8b996a3b1d49ba32e01feR23-R30

hadley:
Closing since you can just use `skip_on_ci()`, which is more general.

kevinushey:
Does `skip_on_ci()` need to check `GITHUB_ACTIONS` as well? It seems like `CI` is not defined on GitHub Actions; at least as documented at https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables.

hadley:
We set it in our CI actions, which I think will cover 99% of R users.

PR #990, opened 2020-03-09
Change defaults for SummaryReporter

krlmlr:
so that dots are omitted, results are limitless, and praise is never shown.

This makes it composable with the new default `ProgressReporter`: `reporter = c("summary", "progress")` gives great progress output and a summary of all skips, warnings and errors.

I also have a version that deprecates the optional arguments and simplifies the code; happy to add that here.


gaborcsardi:
I think some of these changes would make me happy, but not all. E.g. I like the dots and use the summary reporter for this purpose in `test_check()`, because then I have a change to guess which test case has crashed.

I am also wondering if we should "fix" most of these issues with a new reporter that behaves differently on a CI, and produces more verbose output.

krlmlr:
I'm tending towards a new reporter too. I suspect it will live in a package of my own though...

Mildly related: is there a way to set an option so that it is active when running "Test package" in RStudio? IIUC, RStudio is stubbornly running `R --vanilla` (for good reasons) for tool commands, maybe I have missed a hack? If not, should we support setting the default reporter via an environment variable?

gaborcsardi:
> is there a way to set an option so that it is active when running "Test package" in RStudio?

It is tricky to detect this, because env vars are inherited, here is my best effort:
https://github.com/r-lib/cli/blob/master/R/rstudio-detect.R#L2

krlmlr:
I want Ctrl + Shift + T to use e.g. the summary reporter. I have hacked helper files in packages to do that; I'd rather set a global environment var, as in `TESTTHAT_DEFAULT_REPORTER=summary,fail`.

krlmlr:
...and I'm happy if this applies only to my system, not to everyone's.

hadley:
See `default_reporter()` to change the default that `devtools::testthat()` uses.

Otherwise, it seems like this is probably best as a separate package.

krlmlr:
Does it work for you in RStudio in the "Build" pane? AFAICT it's running with `R --vanilla`.

Otherwise agree to move elsewhere.

hadley:
Oh that's a good question; you'd need to take it up with the IDE team.

PR #992, opened 2020-03-09
strip crayon style from errors in junit reporter's xml fixes #991

### LINKED ISSUE #991: ANSI codes in backtrace cause JunitReporter to create invalid xml
On our build machine, when error messages have backtraces they print with ANSI escape codes (for color, I believe). This causes problems when we use `JunitReporter` to create an xml report because the ANSI codes are left in the xml and they are not valid xml.

I don't want to disable ANSI codes everywhere because elsewhere we check the output of objects that take advantage of ANSI codes.

I've created an example package here: https://github.com/gergness/ttjtest

And here's the offending xml file: https://github.com/gergness/ttjtest/blob/master/tests/ttjtest.xml

``` r
xml2::read_xml("ttjtest.xml")
#> Error in doc_parse_file(con, encoding = encoding, as_html = as_html, options = options): PCDATA invalid Char value 27 [9]
```

Thank you!

gergness:
A simple way to fix #991 where invalid xml are created from the backtraces in error messages.


R/reporter-junit.R
```diff
@@ -122,12 +122,12 @@ JunitReporter <- R6::R6Class("JunitReporter",
...
         # "type" in Java is the exception class
         error <- xml2::xml_add_child(testcase, "error", type = "error", message = first_line(result))
-        xml2::xml_text(error) <- format(result)
+        xml2::xml_text(error) <- crayon::strip_style(format(result))
         self$errors <- self$errors + 1
       } else if (expectation_failure(result)) {
         # "type" in Java is the type of assertion that failed
         failure <- xml2::xml_add_child(testcase, "failure", type = "failure", message = first_line(result))
-        xml2::xml_text(failure) <- format(result)
+        xml2::xml_text(failure) <- strip_style(format(result))
```
jimhester: You need to qualify this call as well
```suggestion
        xml2::xml_text(failure) <- crayon::strip_style(format(result))
```

hadley:
Thanks!

PR #997, opened 2020-03-25
Show error class for rlang errors

### LINKED ISSUE #994: verify_output() should print error class
when catching an error. It already does when printing the condition object.

``` r
testthat::verify_output("test.txt", {
  tibble::tibble(a = lm)
  tryCatch(tibble::tibble(a = lm), error = identity)
})
#> Warning: Creating reference output
cat(readLines("test.txt"), sep = "\n")
#> > tibble::tibble(a = lm)
#> Error: All columns in a tibble must be vectors:
#> * Column `a` is function
#> 
#> > tryCatch(tibble::tibble(a = lm), error = identity)
#> <error/tibble_error_column_must_be_vector>
#> All columns in a tibble must be vectors:
#> * Column `a` is function
#> Backtrace:
#>  1. base::tryCatch(tibble::tibble(a = lm), error = identity)
#>  5. tibble::tibble(a = lm)
#>  6. tibble:::tibble_quos(xs[!is_null], .rows, .name_repair) /home/kirill/git/R/tibble/R/tibble.R:147:2
#>  7. tibble:::check_valid_col(res, col_names[[j]], j) /home/kirill/git/R/tibble/R/tibble.R:230:8
#>  8. tibble:::check_valid_cols(set_names(list(x), name)) /home/kirill/git/R/tibble/R/tibble.R:265:4
```

<sup>Created on 2020-03-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Closes #994.


hadley:
I think we should do this for every condition

hadley:
Tweaked it a bit more to unify code across all conditions. What do you think of this display?

krlmlr:
Looks great, thanks for taking this on.

krlmlr:
rlang shows classes reversed and omits `condition`:

```
<error/tibble_error_need_rhs_vector>
```

Do we want to be consistent here?

hadley:
@lionel- my main concern with removing some components of the class is that it makes it much harder to match the `verify_output()` back up to the class.

I also realised that if we merge this, it's going to cause a lot of failures once it hits cran. It won't cause failures on CRAN (because `verify_output()` is always skipped there), but it'll still be annoying because it'll cause failures on CI. I think that means it needs to get wrapped up into the 3rd edition settings.

hadley:
Closing, because I expect we'll wrap this up in to `expect_snapshot_error()`

PR #1008, opened 2020-04-09
Provide detailed output from Catch

jimhester:


src/test-catch.cpp
```diff
@@ -13,14 +13,14 @@ void ouch() {
 
 } // anonymous namespace
 
-context("Example Unit Test") {
+context("Catch: Example Unit Test") {
```
jimhester, line 16: Since these are now showing up in the main testthat results I added a prefix

inst/include/testthat/testthat.h
```diff
@@ -123,7 +123,8 @@ inline Catch::Session& catchSession()
 
 inline bool run_tests()
 {
-  return catchSession().run() == 0;
+  const char* argv[] = {"catch", "-r", "xml"};
+  return catchSession().run(3, argv) == 0;
```
jimhester, line 127: Surely there must be a better way to do this, but the docs related to this only seem to reference the command line interface (https://github.com/catchorg/Catch2/blob/master/docs/command-line.md#choosing-a-reporter-to-use)

tests/testthat/test-cpp.R
```diff
@@ -1,3 +1,3 @@
-test_that("Catch unit tests pass", {
-  expect_cpp_tests_pass("testthat")
-})
+#test_that("Catch unit tests pass", {
```
jimhester, line 1: With this new style we would get duplicate results if we run it within a `test_that()` block

R/test-compiled-code.R
```diff
@@ -21,10 +26,66 @@ expect_cpp_tests_pass <- function(package) {
...
+        org_text <- xml2::xml_text(org, trim = TRUE)
+
+        filename <- xml2::xml_attr(failure, "filename")
 
-  expect(tests_passed, paste("C++ unit tests:", info, sep = "\n"))
+        line <- xml2::xml_attr(failure, "line")
+        failure_srcref <- srcref(srcfile(file.path("src", filename)), c(line, line, 1, 1))
+
+        exp <- expectation("failure", org_text, srcref = failure_srcref)
+        exp$test <- test_name
```
jimhester, line 65: This is slightly awkward, but the reporters expect a test file as a list element. It is normally added by `test_code()` at https://github.com/r-lib/testthat/blob/c5daf5ba21692a38de2c3068c5b8b8bb741ffab6/R/test-that.R#L59

jimhester:
@kevinushey it would be good to have a second set of eyes look at this, also maybe you know of a way to set the reporter apart from the command line interface.

inst/include/testthat/testthat.h
kevinushey: To the best of my knowledge this is indeed the only way to go :(

R/test-compiled-code.R
```diff
@@ -21,10 +26,66 @@ expect_cpp_tests_pass <- function(package) {
     }
   )
 
-  # Drop first line of output (it's jut a '####' delimiter)
-  info <- paste(output[-1], collapse = "\n")
+  report <- xml2::read_xml(paste(output, collapse = "\n"))
```
kevinushey, line 29: Should this be wrapped in `tryCatch()` in case Catch produces invalid XML (e.g. maybe something goes wrong during a segfault)?

jimhester: sure, that is probably safest

jimhester: Actually, I guess if the XML is invalid then xml2 should fail, which is what we want in this case I think?

R/test-compiled-code.R
```diff
@@ -21,10 +26,66 @@ expect_cpp_tests_pass <- function(package) {
...
+  xml2::write_xml(report, "test.xml")
```
jimhester, line 30: This is inadvertent, was here when testing, I will remove it

R/test-compiled-code.R
jimhester: So this happened to me yesterday, and I am not sure of the best way to handle it. Even if we throw a normal error here it ends up being caught in the testthat machinery, but not shown, so you see there was a warning and an error, but not what they were.

I guess we could throw a registered error like we do below...

PR #1019, opened 2020-04-24
fix example code

kos59125:


R/teardown.R
```diff
@@ -16,7 +16,7 @@ teardown_env$queue <- list()
 #' \dontrun{
 #'
 #' tmp <- tempfile()
-#' setup(writeLines(tmp, "some test data"))
+#' setup(writeLines(con = tmp, "some test data"))
```
hadley, line 19: I think it would be better to just flip the argument order here. I always forget which comes first 😞

kos59125: I follow your comment.

hadley:
Thanks!

PR #1022, opened 2020-04-28
Export reassign_function C code and provide a header file

jonkeane:
First of all, thanks for all of the work that has gone in to `testthat` 

Would it be possible to export and provide a header file for `reassign_function` and `duplicate_` for use in other packages? I'm working on a [database testing package](https://github.com/jonkeane/dittodb) that would benefit from importing them directly rather than copying the code over.

I've made the necessary changes in this PR following [R packages](http://r-pkgs.had.co.nz/src.html#c-export) but am happy to make any adjustments or additions requested.

### DIFF
--- a/inst/include/testthatAPI.h
+++ b/inst/include/testthatAPI.h
@@ -1,0 +1,17 @@
+#include <R_ext/Rdynload.h>
+#include <R.h>
+#include <Rinternals.h>
+
+SEXP duplicate_(SEXP x) {
+  static SEXP(*fun)(SEXP) = NULL;
+  if (fun == NULL)
+    fun = (SEXP(*)(SEXP)) R_GetCCallable("testthat", "duplicate_");
+  return fun(x);
+}
+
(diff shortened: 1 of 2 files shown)


jonkeane:
I investigated the test failures on CI and each of them looks to be a slight change in formatting for testing numerics:

```
  "`actual` (`i`) not equal to `expected` (2)."
230
  ""
231
+ "  `actual`: 1"
232
- "`actual`:   1"
233
  "`expected`: 2"
```

I checked on my fork and I see the same failures on the master branch as well. I suspect the change to justification in https://github.com/r-lib/waldo/commit/8b64a4e11635cce18553931ca17fab6c1ed504c9#diff-73d15c2b4a5d43c10e33474095609e65R39 is what's causing this and not anything in my PR, but   please let me know if there is something in my changes that I'm missing and that is causing this.

hadley:
I would feel extremely uncomfortable exporting these functions, or even really directly encouraging people to use them, because they're fundamentally not things you should do.

PR #1032, opened 2020-05-11
Concurrent testthat

gaborcsardi:
PoC. See vignette and code comments for details.

TODO list:
- [ ] Auto-detect number of processors. (Needs fix in ps for Windows.)
- [ ] Fix the reporters that always need the trace, e.g. the `Location` reporter. The reporter will be able to declare that it needs the trace, and then we'll copy them over.
- [ ] Write new reporter that shows a proper progress bar.
- [x] Handle crashes. E.g. replay partial result. **We just error for now, that might be good enough.**
- [ ] Handle the `env` argument properly.
- [ ] Handle failures in the helper, startup and teardown files.
- [ ] Handle process cleanup on OSes that ps does not support.
- [x] Fix testthat's own tests when running them in parallel.
- [ ] Make it more efficient with many expectations. Collect the expectations at the client side and copy them over at once, a regular time intervals, or at the end of the file.
- [ ] Support timeouts.
- [ ] Make sure covr handles parallel tests. Right now it fails for testthat's own test suite.
- [ ] Fix testthat's own tests on Windows. There are a bunch of warnings now, and one test freezes.


R/test-directory-c.R
```diff
@@ -90,7 +88,7 @@ test_dir_parallel <- function(path,
     TESTTHAT_DIR = testthat_dir
   ))
 
-  paths <- find_test_scripts(path, filter, ...)
+  paths <- sort(find_test_scripts(path, filter, ...))
```
jimhester, line 91: One way to 'solve' your slowest tests is the last one problem would be to randomize the order the test files are run in.

This would also more quickly smoke out hidden dependencies between test files.

I think the big drawback would be the order of test outputs would then no longer be deterministic, which probably makes it a non-starter.

gaborcsardi: Yeah, I thought about that. Maybe eventually we'll have a config parameter for this, but I think in general it would be nice if testthat remembered the last test results. E.g. you could re-run the failing tests easily. And if we remember the previous results, then we can just order the files according the previous results.

gaborcsardi: Btw. I don't think a random order is a non-starter. We are used to seeing the lines of successful test files, but honestly, while this is heart-warming, it is not very informative. I'd rather see a very short summary of all successes, and then the execution order does not matter that much.

gaborcsardi: I guess we'll need to experiment with this a bit....

gaborcsardi: Actually, maybe we should randomize the order by default. Because otherwise the interdependence might not come up until much later, potentially on CRAN. Just like it came up for testthat itself on GHA...

tests/testthat/test-test_dir.R
```diff
@@ -17,13 +25,17 @@ test_that("test_dir()", {
   df <- as.data.frame(res)
   df$user <- df$system <- df$real <- df$result <- NULL
 
+  # We order this here, because the order of parallel tests is
+  # not fully predictable.
+  df <- df[order(df$file, df$context,  df$test), ]
```
hadley, line 30: Don't you really only need to order by file? The order within a file is still deterministic.

R/test-directory-c.R
```diff
@@ -0,0 +1,356 @@
...
+#                                  |--| Subprocess 2                  |
+#                                  |  +-------------------------------+
+#                                  |  +-------------------------------+
+#                                  +--| Subprocess 3                  |
+#                                     +-------------------------------+
+#                                       ...
+#
+# ## Notes
+#
+# * Subprocesses run `callr::r_session` R sessions. They are re-used,
```
hadley, line 29: Maybe integrate these notes into the vignette so they don't even get out of sync? (I think mostly they're there already)

R/test-directory-c.R
```diff
@@ -0,0 +1,356 @@
...
+}
+
+start_test_processes <- function(dir, paths, pkg_name, testthat_dir,
+                                 load_helpers, pkgload) {
+
+  # TODO: detect number of CPUs
+  num_proc <- min(getOption("Ncpus", 4), length(paths))
+  proc <- replicate(num_proc, NULL)
+
+  load_hook <- rlang::expr(asNamespace("testthat")$process_setup(
```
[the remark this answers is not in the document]
gaborcsardi: Yeah, because this runs in the subprocess.

vignettes/parallel.Rmd
```diff
@@ -0,0 +1,125 @@
...
+user cannot make any assumptions about the order of the files, and
+which subprocess they'll be executed on.
+
+Test files do not start in a clean R process currently. The first *n*
+files, where *n* is the number of subprocesses, do start in a clean R
+process, but users should not rely on that. In particular, options
+(set via `options()` or otherwise) are *not* reset, loaded packages
+are *not* unloaded, the global environment is *not* cleared, etc. In
+the future the user will be able to request a clean R process for a
+test file, see the TODO at
```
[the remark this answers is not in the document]
gaborcsardi: Yeah, it is not a good default. But it would be convenient to be able to request it for some test files, that are special. E.g. you want to run your tests with a package (say tibble), but you also want to have a test file for the case where tibble is not loaded. For the latter you would need a clean session.

Btw. running everything in a clean session might be a good default for a CI, actually. Then even if setting up a session is a second, for 60 test files you only lose a minute.

vignettes/parallel.Rmd
```diff
@@ -0,0 +1,125 @@
...
+<https://github.com/r-lib/testthat/pull/1032> .
+
+Since files run in alphabetical order, ideally the files that take the
+longest to run, should be started first. This way, we can avoid
+starting a long running test file on a subprocess, when the other
+files have finished already and the other subprocesses have nothing to
+do. To achive this, you can name your test files like this:
```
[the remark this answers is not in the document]
gaborcsardi: That's OK, I think. Initially the names assigned by `use_test()` are fine, and if you need to optimize later, you can just rename the files.

vignettes/parallel.Rmd
```diff
@@ -0,0 +1,125 @@
...
+    relies on stack traces being present in all test results, will
+    fail.
+-   The current reporters are not ideal for parallel execution, as
+    they cannot handle running multiple files at the same time.
+    Parallel testthat tricks the reporters by collecting all results
+    for a test file, and then "replaying" them at once. Reporters can
+    not show the progress within files. This will be fixed later.
+-   Testthat does not handle crashes currently. This will be fixed
+    later.
+-   You cannot currently use the `env` argument of `test_dir()`, etc.
```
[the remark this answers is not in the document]
gaborcsardi: Yeah, I agree. But we don't need to do that in this PR, I guess.

vignettes/parallel.Rmd
```diff
@@ -0,0 +1,125 @@
...
+    Do not set this argument if you use parallel testthat. This will
+    be fixed later.
+-   Test files that perform many quick expectations might be slow.
+    This will be fixed later.
```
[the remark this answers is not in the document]
gaborcsardi: The expectations will be collected in the subprocess and sent to the main process in batches, e.g. once every second or so.

R/test-directory-c.R
gaborcsardi: I am leaning towards a randomized order on the CI now, and alphabetical order locally, for a better interactive experience. But it should be all configurable.

R/test-directory-c.R
```diff
@@ -0,0 +1,356 @@
...
+      msgs <- queue$poll(Inf)
+      for (x in msgs) {
+        if (x$code == 301) {
+          m <- x$message
+          # TODO: should we keep other messages in the subprocess? Probably.
+          if (!inherits(m, "testthat_message")) next
+          files[[m$filename]] <- append(files[[m$filename]], list(m))
+          if (m$cmd == "end_file") replay(m$filename)
+
+        } else if (x$code == 200) {
```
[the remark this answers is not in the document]
gaborcsardi: Yeah. Ideally in callr, but we can do it here as well as a temporary solution.

R/test-directory-c.R
```diff
@@ -0,0 +1,356 @@
...
+          # File is done, nothing to do here, we'll get an end_file
+
+        } else if (x$code > 500) {
+          # a subprocess has crashed
```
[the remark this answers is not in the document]
gaborcsardi: OK.

R/test-directory-c.R
```diff
@@ -0,0 +1,356 @@
...
+    stop("no 'Package' field in DESCRIPTION file '", desc_path, "'", call. = FALSE)
+  }
+  pkg
+}
+
+start_test_processes <- function(dir, paths, pkg_name, testthat_dir,
+                                 load_helpers, pkgload) {
+
+  # TODO: detect number of CPUs
+  num_proc <- min(getOption("Ncpus", 4), length(paths))
```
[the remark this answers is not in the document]
gaborcsardi: Yeah, this is temporary, see TODO list in the PR. 

The `Config/testthat/processes` is a bit weird, because the right number depends more on the machine than on the package. Well, unless you want to fix it to one, because you tests concurrently. I think such an option makes sense, but we would need to combine it with an option set by the user, I think.

tests/testthat/test-test_dir.R
gaborcsardi: Yeah, but `order()` might not be stable. The man page suggests that it is, but links to `sort()` and there they say it is only stable if the vector has names. So it is better to be on the safe side.

R/test-directory-c.R
gaborcsardi: I think ideally we would have two vignettes, one for the users, and one with the internals. The current parallel.Rmd is the former, and these notes are really about the implementation details. I can certainly create a `parallel-internals.Rmd` or actually maybe just an `internals.Rmd` with all the testthat internals?

tests/testthat/test-test_dir.R
```diff
@@ -1,3 +1,11 @@
+
+# We cannot currently run test_dir reliably within R CMD check, in parallel
```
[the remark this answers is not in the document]
gaborcsardi: It works if we unset `TESTTHAT_PARALLEL`, i.e. if the `test_dir()` in the subprocess does not create more subprocesses. IDK why it fails when running in parallel within parallel, it didn't seem worth while investigating so far. :)

Should we actually just unset `TESTTHAT_PARALLEL` in `test_dir_parallel()`? Or in the test case?

gaborcsardi:
@hadley Please take another look. Thanks!

tests/testthat/test-test_dir.R
hadley: I think you should just assume it's stable; otherwise sorting by context and test are going to cause a lot of ordering changes. (If it later turns out not to be the case, we can add an integer index to tests to track the exact order)

gaborcsardi: Yes, for the actual results, I agree. But this is a test case, so I just want to make sure that we always get the same results, on all platforms. Might be overly cautious, but it also does not hurt too much...

gaborcsardi: But of course I also don't mind, I can change it of course.

hadley:
Ok, done. @gaborcsardi feel free to play around with this more; I won't look again until next week.

gaborcsardi:
> Ok, done. @gaborcsardi feel free to play around with this more; I won't look again until next week.

Great, thanks!

PR #1047, opened 2020-06-16
Use pdf() device instead of png()

### LINKED ISSUE #1011: verify_output fails on Linux if there is no X11
In general it fails if there is on png() device.

TBH I am not sure if it is testthat's business to deal with graphic output at all, but at this point we might not be able to change this.

We could still use a more minimal device, e.g. the devoid package could be handy here?

hadley:
I experimented with getting rid of the device all together, but I couldn't get it to work, likely because of the hoops that evaluate jumps through to capture graphics device state. I think pdf() should work everywhere.

Fixes #1011

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `verify_output()` now uses the `pdf()` device instead of `png()`; that makes
+  it work on systems without X11 (#1011).
+
 * `expect_identical()`, `expect_equal()`, `expect_equivalent()` and
   `verify_output()` now use `waldo::compare()` to compare actual and expected
   results. The chief difference should be that you now get much more 
--- a/R/verify-output.R
+++ b/R/verify-output.R
@@ -117,7 +117,7 @@ verify_exec <- function(exprs,
   source <- unlist(exprs, recursive = FALSE)
 
   # Open temporary new device
-  grDevices::png(filename = tempfile())
+  grDevices::pdf(tempfile())
   grDevices::dev.control(displaylist = "enable")
   dev <- grDevices::dev.cur()
   on.exit(grDevices::dev.off(dev), add = TRUE)
--- a/R/verify-output.R
+++ b/R/verify-output.R
@@ -103,6 +103,9 @@ verify_exec <- function(exprs,
                         unicode = FALSE,
                         env = caller_env()) {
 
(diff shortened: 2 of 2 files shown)


gaborcsardi:
Weird. Anyway, I agree that `pdf()` should be fine.

Btw. the current `dev.off()` does not actually restore the active device, because the devices are not in a stack, but in a circular list (!!!), and `dev.off()` switches to the **next** device. Only really an issue if there are multiple (non-null) devices open while running the tests, so not very important. :) E.g. in

```r
> png("1.png")
> png("2.png")
> png("3.png")
> dev.off()
png
  2
> plot(1:10)
> dev.off()
png
  3
> dev.off()
null device
          1
```

the plot will go to `1.png`, and not to `2.png`.

hadley:
But it doesn't just call `dev.off()`; it records the old device too?

```R
  dev <- grDevices::dev.cur()
  on.exit(grDevices::dev.off(dev), add = TRUE)
```

gaborcsardi:
> But it doesn't just call dev.off(); it records the old device too?

Right, totally missed that, sry.

gaborcsardi:
Actually, that's still not good. It'll close the right device, yes, but it activate the wrong one. E.g.

```r
> png("2.png")
> png("3.png")
> dev.cur()
png
  3
> testthat::verify_output("output.txt", 1+1)
Warning message:
Creating reference output
> dev.cur()
png
  2
```

gaborcsardi:
I think you also need to look up the current device before creating the temporary pdf device, and then restore that in `on.exit()`.

hadley:
Oh duh, yeah. Not that it really matters here, but I'll fix it.

hadley:
Actually I should just use `withr::local_pdf()` here.

gaborcsardi:
Except that that seem to be broken as well:

```r
> f <- function() { withr::local_pdf(tempfile()) }
> png("2.png")
> png("3.png")
> dev.cur()
png
  3
> f()
> dev.cur()
png
  2
```

hadley:
I'll still use it, so t'll eventually get fixed in withr. Given that no one has complained about the currently broken behaviour, I don't think it's a big worry.

gaborcsardi:
Makes sense. Let me report it in withr. It is more important to fix it there.

PR #1054, opened 2020-06-23
Set reproducibility options in test_that()

### LINKED ISSUE #971: set TESTTHAT envvar within test_file()?
Currently, `testthat` sets a number of environment variables within `test_dir()`:

https://github.com/r-lib/testthat/blob/3633dfeb30fe89aac17b43e5f919447c2a1185da/R/test-directory.R#L82-L86

However, it's possible to run test files individually; e.g. with `testthat::test_file()`. (RStudio uses this when running a single testthat test file.)

Would it make sense to set TESTTHAT (or maybe TESTTHAT_FILE) within `test_file()`?

### LINKED ISSUE #1044: Reconsider local options for test_that/test_file/etc
e.g.

```R
local_test_lifecycle <- function(.local_envir = parent.frame()) {
  withr::local_options(
    list(
      lifecycle_verbosity = "warning",
      # Promote retirement stages except on CRAN
      lifecycle_verbose_retirement = if (on_cran()) TRUE
    ),
    .local_envir = .local_envir
  )
}
```

(partly from #1000)

hadley:
Fixes #1044. Fixes #971.


R/test-directory.R
```diff
@@ -91,11 +86,6 @@ test_dir <- function(path,
   source_test_setup(path, env)
   on.exit(source_test_teardown(path, env), add = TRUE)
 
-  # Promote retirement stages except on CRAN
-  if (identical(Sys.getenv("NOT_CRAN"), "true")) {
-    withr::local_options(list(lifecycle_verbose_retirement = TRUE))
```
hadley, line 96: This doesn't appear to do anything — there's no mention of "lifecycle_verbose_retirement" in the lifecycle package.

R/reporter-progress.R
```diff
@@ -47,6 +48,7 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
       self$show_praise <- show_praise
       self$min_time <- min_time
       self$update_interval <- update_interval
+      self$frames <- cli::get_spinner()$frames
```
hadley, line 51: Need to store spinner frames on init, since that's what should determine the unicode status.

R/local.R
```diff
@@ -1,36 +1,68 @@
...
+#' detected as running), and `TESTTHAT = "true"` (which ensures that
+#' [is_testing()] returns `TRUE)`, and sets collation locale to "C"
+#' (which ensures that character sorting the same regardless of current
+#' locale).
+#'
+#' @export
+#' @keywords internal
+#' @param .env Environment to use for scoping; expert use only.
+local_test_context <- function(.env = parent.frame()) {
+  withr::local_envvar(list(TESTTHAT = "true"), .local_envir = .env)
```
jennybc, line 40: Nitpick: I always use a named character vector with `withr::local_envvar()`, but clearly named list works too.

NAMESPACE
```diff
@@ -124,6 +124,7 @@ export(is_null)
 export(is_testing)
 export(is_true)
 export(local_mock)
+export(local_test_context)
```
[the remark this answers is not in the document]
hadley: `local_test_environment()`?

R/reporter-progress.R
gaborcsardi: This is enough for the progress reporter, but in general might be not. With the current setup the `add_result` handler runs in the reproducible context, which is correct. We might be to get away with it, so we don't necessarily need to do anything now, just don't forget this. :) Maybe add a comment somewhere?

gaborcsardi:
> Is there any way to register your own options or env vars to set and restore?

testthat cannot provide an API for this, because you would not want to import testthat. So yeah, one solution would be to set options with a specific name, and a setup and a restore function. E.g. usethis would do something like this in `.onLoad()`:

```r
options(
  testthat_context_usethis = list(
    setup = function(.env) {
      options(usethis.quiet = TRUE)
    },
    restore = function(.env, old) {
      options(old)
    }
  )
)
```

testthat would get all the `testthat_context_*` options, and call their `setup()` functions at the beginning of `test_that()` and then call their `restore()` functions at the end, passing the return value of `setup()` to the latter, in addition to the env.

NAMESPACE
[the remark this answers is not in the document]
jennybc: I thought about that too, but “environment” also has very special meaning.

hadley:
Packages can already uses the TESTTHAT env var to condition behaviour based on whether or not testthat is running. Do we need something more?

jennybc:
> Packages can already uses the TESTTHAT env var to condition behaviour based on whether or not testthat is running. Do we need something more?

I guess it's a question of where to put the instructions to do THIS not THAT when testing.

If there were a way to register package-specific test state instructions, then they are all centralized in, e.g., `onLoad()`, following @gaborcsardi's proposal.

If we're supposed to consult the env var`TESTTHAT`, these conditionals potentially end up sprinkled in function bodies throughout the package.

I feel like centralizing them makes them much more visible and maintainable.

NAMESPACE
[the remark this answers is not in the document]
hadley: Given the paucity of good names and this is going to be used by relatively few people, I think `context()` is probably fine.

hadley:
I don't see how we can make a registry work when you are running the tests interactively. That would force you to also run `local_test_context()` so that the behaviour matches up inside and outside of the test (i.e. in more cases than you do now). 
Overall, making this hookable at the package level, feels to me like a step backward; we generally want to reduce the amount of implicit context in a test, not increase it.

gaborcsardi:
> Packages can already uses the TESTTHAT env var to condition behaviour based on whether or not testthat is running. Do we need something more?

Maybe. For the other package, I think just setting an option is easier. But we don't have to rush this now I think.

R/reporter-progress.R
hadley: Hmmm, to fix this probably I think we'd need to split `test_code()` into two pieces — one piece would run the code and catch the expectations (with `local_test_context()`), and then the other would send the expectations to the reporter. I'll turn this into an issue and think about more later on.

gaborcsardi:
> I don't see how we can make a registry work when you are running the tests interactively. That would force you to also run local_test_context() so that the behaviour matches up inside and outside of the test (i.e. in more cases than you do now).

Yes, when you run parts of a `test_that()` block interactively, you already have to run `local_text_context()` I believe. No?

> Overall, making this hookable at the package level, feels to me like a step backward; we generally want to reduce the amount of implicit context in a test, not increase it.

I agree that using well this needs some discipline from the packages. But if we don't have a hook, then we need to collect all such information in testthat, which also does not seem right. (Although the `TESTTHAT` env var is already a hook as you say.)

gaborcsardi:
> I feel like centralizing them makes them much more visible and maintainable.

Yeah, I agree. The alternative is to introduce a helper function into your package (usethis), that queries `TESTTHAT` first, and only consults the option (`usethis.quiet`) if it is not set. That is not too bad, either, I think.

hadley:
Hmmm I just noticed that `rlang::is_interactive()` already checks for the testthat envvar, so I don't need to add it here.

gaborcsardi:
Yeah, I think just having a helper that checks for `TESTTHAT` before the user-defined option is fine and much simpler.

hadley:
One more point in favour of doing this in one place — it's nice to know that you can load at `local_test_context()` to see all the options that are set, rather than having to carefully read the docs for all the functions that you call

gaborcsardi:
Yeah, we want to have it in one place as well, I think. That's actually the motivation for the API in https://github.com/r-lib/testthat/pull/1054#issuecomment-649536274

With that, for a given `test_that()` block, you can list _all_ options/env vars/etc. that are changed, by simply looking at `options()`.

If packages will change their behavior based on `TESTTHAT`, then you cannot really do this.

PR #1067, opened 2020-06-29
Display details of errors/failures outside of tests

### LINKED ISSUE #1004: Error in setup code triggers silent failure
E.g. 

```
setup(1+"a")
```

```
✔ |  OK F W S | Context
⠋ |   0 1     |
══ Results ═════════════════════════════════════════════════════════════════════
OK:       0
Failed:   1
Warnings: 0
Skipped:  0
```

Would be great to print that the setup code failed.

hadley:
Fixes #1004


R/reporter-progress.R
```diff
@@ -62,18 +62,11 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
...
       self$ctxt_start_time <- proc.time()
-    },
 
-    start_test = function(context, test) {
-      if (is.null(context)) {
-        # Have to go via get_reporter() in order to trigger multi-reporter
-        name <- context_name(self$file_name)
-        get_reporter()$.start_context(name)
-      }
+      name <- context_name(self$file_name)
```
hadley, line 68: I don't understand why this was previously only done in the first test, which reset any errors/failures captured earlier.

I guess there's a subtle change in behaviour if you had an empty test file? Previously it wouldn't have been shown; now it will be.

hadley: Oooh, this is to ensure that if you do have a context, you don't get two context messages printed out

hadley: Hmmm, not sure how to work around this. I only want to set the context in `start_file()` if it's not set elsewhere; maybe I should just grep the file for `context()` in the 2nd edition and earlier?

hadley: Ok, I've just tweaked the display so the first context from the file will be over-written by the  context from `context()`. 

It's not quite right, because I'm getting an extra new line whenever the context takes more than `min_time` but I can't figure out where it's coming from.

R/test-that.R
```diff
@@ -58,7 +58,7 @@ test_code <- function(test, code, env = test_env(), skip_on_empty = TRUE) {
       e$end_frame <- sys.nframe() - debug_end - 1L
     }
 
-    e$test <- test %||% "(unknown)"
+    e$test <- test %||% "(code run outside of a test)"
```
gaborcsardi, line 61:
```suggestion
    e$test <- test %||% "(code run outside of `test_that()`)"
```
Maybe? "outside of a test" might be taken differently, as everything in `test-foo.R` might be considered "inside of a test".

R/reporter-progress.R
gaborcsardi: Yeah, I think the new code is cleaner.

tests/testthat/reporters/progress.txt
```diff
@@ -1,16 +1,16 @@
 ✔ |  OK F W S | Context
 
-⠏ |   0       | tests
-⠋ |   1       | tests
+⠏ |   0       | tests                                                           
+⠋ |   1       | tests
```
gaborcsardi, line 4: Do you know why the whitespace differences happen? Is it because of changes in testthat itself?

hadley: That's the new `strpad()` I added in case the first context is shorter than the second

PR #1069, opened 2020-06-30
Check max fails after each expectation

### LINKED ISSUE #967: max_fail is checked after `test_that` blocks, not after each expectation
I see that the default to 10 `max_failures` is no longer respected. The `max_fail` seems to no longer be used in the code https://github.com/r-lib/testthat/search?q=max_fail&unscoped_q=max_fail

hadley:
Fixes #967

@lionel- it's pretty simple, but can you please double check that I've used the restart correctly, and my wrapper has an appropriate name?

@gaborcsardi this is a more drastic example of the printing problem — `$end_context()` is usually called from outside of the tests, but now it's being called inside a test, and hence all the nice formatting is lost because of `local_reproducible_output()`.  Maybe `ProgressReport` should capture `width`, `crayon`, and `unicode` at start up so it can restore for these situations?


gaborcsardi:
> Maybe ProgressReport should capture width, crayon, and unicode at start up so it can restore for these situations?

Yes, I think that pattern would make sense. Ideally we could use the same function to (partially) switch the context back in the reporter, but it's not super important.

hadley:
Just pushed the minimum fix for this PR. The only method that is run within `test_code()` is `$add_result()` which calls `$show_status()`, which only uses colour when `complete = TRUE` which is usually only called from outside the test block. So I don't _think_ I need to do anything else, but we have a technique prepared for future use, if needed.

R/reporter-progress.R
```diff
@@ -53,6 +56,8 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
       # Capture at init so not affected by test settings
       self$frames <- cli::get_spinner()$frames
       self$width <- cli::console_width()
+      self$unicode <- getOption("cli.unicode")
+      self$colour <- getOption("crayon.enabled")
```
gaborcsardi, line 60: I think this does not work if you have unicode and/or color because they were auto-detected. `cli::is_utf8_output()` and `crayon::has_color()` are probably better.

hadley: Oh good catch, thanks!

hadley:
`stop_reporter()` also just sounds like a better name. Thanks for the review!

PR #1070, opened 2020-06-30
Check class in expect_messages() and expect_warnings()

### LINKED ISSUE #1052: Classed warning expectation doesn't fail
Should fail:

```r
testthat::expect_warning(warning("foo"), class = "bar")
#>
```

Reported in r-lib/lifecycle#50

hadley:
This unifies the output across expect_message()/expect_warning() and expect_condition()/expect_error(). The new messages are possibly a little less informative when you have multiple messages/warnings that don't match, but I think this is relatively rare and it's a relatively small decrease in utility so I haven't put more time into fixing it.

I also added calls to ellispis::check_dots_used() to make sure other arguments aren't silently swallowed in the same way that `class` was here.

Fixes #1052

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # testthat (development version)
 
+* `expect_message()` and `expect_warning()` gain a `class` argument so that you
+  can assert that message or warnings have a specific class, rather than
+  checking for matching text (#1052).
+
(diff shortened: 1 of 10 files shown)


hadley:
Closing in favour of revamped #1068

PR #1074, opened 2020-06-30
Breakdown skipped tests by type

### LINKED ISSUE #1028: Capture skip "type"
So skips can later be reported broken down by type (as in #1026)

hadley:
Fixes #1028

@jennybc the implementation is pretty straightforward so I'd particularly like feedback on how the skipped tests are display in the reports.


jennybc:
I think this looks good. Caveat: I haven't actually run any of this code, I'm just looking at the PR.

PR #1084, opened 2020-07-05
Implement test_local()

### LINKED ISSUE #1030: Have a `test()` function, similar to `devtools::test()`
testthat already uses pkgload for package loading, so there is not much benefit using devtools.

hadley:
Also eliminates need for devtools.

* [x] Add news bullet
* [x] Figure out why `test_local(".")` gives different results to `devtools::test()` on testthat itself

Fixes #1030


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # testthat (development version)
 
+* New `test_local()` tests a local source package. It's equivalent to
```
gaborcsardi, line 3: Maybe clarify that it is not a package file, but a directory? Not terribly important in the NEWS, though.

R/test-directory.R
```diff
@@ -181,12 +178,53 @@ test_check <- function(package,
...
+test_local <- function(path = ".",
+                       filter = NULL,
+                       reporter = default_reporter(),
+                       ...,
+                       stop_on_failure = TRUE,
+                       stop_on_warning = FALSE
+                       ) {
+  package <- pkgload::pkg_name(path)
+
+  test_path <- file.path(path, "tests", "testthat")
```
gaborcsardi, line 190: Should `test_local()` work from a subdirectory of the package?

NEWS.md
hadley: :+1:

R/test-directory.R
hadley: Good idea, done.

PR #1088, opened 2020-07-08
Set wd to tempdir() in test_that()

### LINKED ISSUE #1059: Change working directory
Either change working directory to the package root, and consequently deprecating test_path(); or possibly setting the working directory to `tempdir()` and requiring `test_path()`.

hadley:
Fixes #1059

Now that I've implemented this I'm not sure it's worth it — yes, it forces you to use `test_path()`, but forcing you to use that only ensures that your tests work when you run them interactively; they already work in R CMD check. So when we turn this on, you'll have to run down a bunch of unimportant changes without any immediate payoff.

I'm also not sure which working directory the setup and teardown files should use. Currently I'm setting the working directory inside `test_that()` so anything outside of a `test_that()` will be run in the default working directory (since this is the best we can do for interactive usage). I'm not even sure that setting `test_file()` level will be worth it because `source()` will get absolute urls, rather than relative, which will make the srcrefs in error messages that much harder to parse.

So overall, seems like a very minor gain for quite a lot of additional complexity, but I'd love to get your thoughts.


R/test-files.R
```diff
@@ -148,9 +148,9 @@ test_files <- function(test_dir,
 
 test_one_file <- function(path, env = test_env()) {
   reporter <- get_reporter()
-  on.exit(teardown_run(), add = TRUE)
+  on.exit(teardown_run(tempdir()), add = TRUE)
```
hadley, line 151: This feels iffy to me, but it's what preserved the existing behaviour in the tests.

tests/testthat/reporters/check.txt
```diff
@@ -19,40 +19,40 @@ Error: stop
 Error: !
 Backtrace:
     █
- 1. └─f()
- 2.   └─g()
- 3.     └─h()
+ 1. └─f() reporters/tests.R:47:2
```
hadley, line 22: I'm not sure why this now generates srcrefs. But I'm pretty sure it's just an artefact of the testing setup of testthat; it shouldn't affect other packages.

gaborcsardi:
I agree that it is ok not to change the wd, and basically keep the current behavior. 

Maybe we could add more docs to stress that you need to use `test_path()`, otherwise people might get stumped when only `R CMD check` has issues and `test()` does not.

R/utils.R
```diff
@@ -56,3 +56,15 @@ can_entrace <- function(cnd) {
 }
 
 isNA <- function(x) length(x) == 1 && is.na(x)
+
+relish <- function(x, dir = testing_directory()) {
```
jimhester, line 60: Can we come up with a different name for this function

hadley:
If we only set it in `test_dir()`, then this has a neutral impact on consistency because it'll still be different for interactive usage. (i.e. it's no more consistent than the current version)

R/utils.R
hadley: Chutney? Condiment? Sauce?

jennybc:
I think we have two main goals. First this one articulated by @jimhester 

> I think we should do whatever we can improve consistency when running tests interactively and in R CMD check. Differences are often hard to debug and difficult to explain to people new to package development or testing.

and then a related one: help people develop file path discipline in their tests, so: they're developing good general habits and won't create trouble for themselves at CRAN submission time.

I find everyone's arguments compelling that the wd change explored here does not have a great payoff in terms of achieving those things. I'm happy with the *status quo*.

hadley:
So it looks like we're all in agreement that this was worth exploring, but it doesn't seem like we can do significantly better than currently, so it's best to leave as is.

PR #1091, opened 2020-07-09
Start on 3e vignette

### LINKED ISSUE #1061: Vignette for 3e
Mostly focussed on converting a package, and the problems that you're likely to encounter.

* https://github.com/r-lib/waldo/issues/25
* delete all uses of `context()`
* convert any old `expect_that()` calls
* replace `expect_equivalent()`

hadley:
Fixes #1061


jennybc:
I used `pr_push()` to make that one change and it worked as expected.

PR #1097, opened 2020-07-14
Write test fixtures vignette

### LINKED ISSUE #1042: Document text fixtures pattern
https://www.tidyverse.org/blog/2020/04/self-cleaning-test-fixtures/

Per #1021, show how to use for a whole file, but talk about the downsides — non-locality makes it harder to debug. Emphasise fixtures over `setup()` and `teardown()`

hadley:
Fixes #1042


vignettes/test-fixtures.Rmd
```diff
@@ -0,0 +1,317 @@
...
+options(op)
+```
+
+Notice how `pi` prints differently before and after the call to `sloppy()`. Calling `sloppy()` has a side effect: it changes the `digits` option globally, not just within its own scope of operations. This is what we want to avoid[^1].
+
+[^1]: Don't worry, I'm restoring global state (specifically, the `digits` option) behind the scenes here.
+
+### `on.exit()`
+
+The first function you need to know about is base R's `on.exit()`. `on.exit()` calls the code to suppled to its first argument when the current function exits, regardless of whether it returns a value or errors. You can use `on.exit()` to clean up after yourself by ensuring that every mess-making function call is paired with an `on.exit()` call that cleans up.
```
jennybc, line 75:
```suggestion
The first function you need to know about is base R's `on.exit()`. `on.exit()` calls the code to supplied to its first argument when the current function exits, regardless of whether it returns a value or errors. You can use `on.exit()` to clean up after yourself by ensuring that every mess-making function call is paired with an `on.exit()` call that cleans up.
```

vignettes/test-fixtures.Rmd
```diff
@@ -0,0 +1,317 @@
...
+
+We can use this idea to turn `sloppy()` in `neat()`:
```
jennybc, line 77:
```suggestion
We can use this idea to turn `sloppy()` into `neat()`:
```

vignettes/test-fixtures.Rmd
```diff
@@ -0,0 +1,317 @@
...
+  
+  # switch to new usethis project
+  proj_set(dir) # C
+  withr::defer(proj_set(old_project, force = TRUE), envir = env) # -C
+  
+  dir
+}
+```
+
+(If we wrote this today, would call it `local_create_package()` to emphasise the local-ness of it.)
```
jennybc, line 284: I'm happy to rename it in usethis. I don't have any attachment to the name.

vignettes/test-fixtures.Rmd
```diff
@@ -0,0 +1,317 @@
...
+
+### Scope
+
+So far we have applied our test fixture to individual tests, but it's also possible to apply them to an entire file, just by calling the test fixture at the top of the file. This has both advantages and disadvantages:
+
+-   If you would otherwise have called the fixture in every test, you've saved yourself a bunch of work and duplicate code.
+
+-   But on the downside, if you a test fails and you want to recreate the failure in an interactive environment so you can debug, you need to remember to run all the setup code at the top of the file first.
+
+Generally, I think it's to copy and paste test fixtures across many tests --- sure, it adds some duplication to your code, but it makes debugging test failures so much easier.
```
jennybc, line 309: I think you mean:

```suggestion
Generally, I think it's better to copy and paste test fixtures across many tests --- sure, it adds some duplication to your code, but it makes debugging test failures so much easier.
```

vignettes/test-fixtures.Rmd
hadley: Ok, renamed. You can change in usethis at your leisure; I don't think there's any rush.

vignettes/test-fixtures.Rmd
```diff
@@ -0,0 +1,333 @@
...
+
+Generally, I think it's better to copy and paste test fixtures across many tests --- sure, it adds some duplication to your code, but it makes debugging test failures so much easier.
+
+### Package
+
+To run code before any test is run, you can create a file called `test/testthat/setup.R`. If the code in this file needs clean up, you can use the special `teardown_env()`:
+
+```{r, eval = FALSE}
+# Run before any test
+write.csv("mtcars.csv", mtcars)
```
jennybc, line 319: Do you even want to show writing a file that's not in the session temp dir 😬?

vignettes/test-fixtures.Rmd
```diff
@@ -0,0 +1,333 @@
...
+
+# Run after all tests
+withr::local(unlink("mtcars.csv"), teardown_env())
+```
+
+Setup code is typically best used to create external resources that are needed by many tests. It's best kept to a minimum because you will have to manually run it before interactively debugging tests.
```
jennybc, line 325: Maybe a good place to touch on what's same/different about `setup.R` vs `helper.R`.

vignettes/test-fixtures.Rmd
hadley: I think it's ok to write into the test directory; I don't really want to write into the session temp dir because there isn't really much point in cleaning it up since the tests are usually run in a fresh session.

vignettes/test-fixtures.Rmd
hadley: Did you see elsewhere that I'm no longer recommending the use of `helper.R`? Instead just put the code in `R/`

hadley:
(I've merged but will keep working on the vignette and docs)

PR #1098, opened 2020-07-15
Fix typo in test-fixtures vignette

Bisaloo:
I *think* this is what was intended. There is no exported function called `local()` in withr.


hadley:
Ooops, yes, thanks!

PR #1105, opened 2020-07-20
Fix typo

jimhester:


hadley:
https://github.com/r-lib/testthat/runs/891056589?check_suite_focus=true#step:7:11

jimhester:
Ah, need to fix that one as well now I guess...

PR #1107, opened 2020-07-21
Added where to put custom `expect_` functions

### LINKED ISSUE #1106: custom expectations
I'm running into a problem that I think probably has a simple solution, but this solution should probably be **made explicit in the vignette for Custom Expectation**.

When we have defined one, where do we put the code for it? Should it go in a file in the `R/` directory, in a file in the `tests/testthat/` directory, or somewhere else?

Intuitively, I originally put them in my `testthat.R` file, but that doesn't seem to be working. I also have a few other lines of code (e.g. defining dataset) that I would like to run before my tests.

MyKo101:
Fixes #1106 by giving advice on where to save custom expectation functions

### DIFF
--- a/vignettes/custom-expectation.Rmd
+++ b/vignettes/custom-expectation.Rmd
@@ -12,7 +12,7 @@ library(testthat)
 knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
 ```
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1108, opened 2020-07-22
Add vignette on skipping

### LINKED ISSUE #1060: Write vignette about skipping
* When/why to skip tests
* Writing helpers
* How to test your helpers (e.g. with env var)

hadley:
Fixes #1060


vignettes/skipping.Rmd
```diff
@@ -0,0 +1,61 @@
...
+
+```{r}
+skip_if_Tuesday <- function() {
+  if (as.POSIXlt(Sys.Date())$wday == 2) {
+    skip("Not run on Tuesday")
+  }
+}
+```
+
+It's also very important to test your skip helpers because it's easy to miss if accidentally made them always skip, and the test code is never run. This is unlikely to happen locally (since you'll see the skipped tests in the summary), but is quite possible that it may happen in continuous integration.
```
jennybc, line 53: > easy to miss if accidentally made them always skip

Wording problem

vignettes/skipping.Rmd
```diff
@@ -0,0 +1,61 @@
...
+
+For that reason, it's a good idea to add a test that you skip is activated when you expect. skips are a special type of error, so you can test that they occur with `expect_error()`:
+
+```{r}
+test_that("not skipped on CI", {
+  expect_error(skip("Hi"), class = "skip")  
+})
+```
```
jennybc, line 61: I think I need more words or a more fleshed out example here.

hadley:
You automatically get more visibility into skips as soon as you use 3.0.0 so I didn't say anything, but maybe it's worth a comment just to reinforce where to look.

jennybc:
> You automatically get more visibility into skips as soon as you use 3.0.0 so I didn't say anything, but maybe it's worth a comment just to reinforce where to look.

Yeah I feel like 3e comes with better skip handling AND there's this new insight / advice to express your expectation around a skip condition in an actual test. Together, this seems like a substantial upgrade in thinking and tooling re: skipping.

PR #1122, opened 2020-07-28
Switch to brio

### LINKED ISSUE #1120: Use brio everywhere
Since file snapshotting needs it

hadley:
Fixes #1120.

Needs https://github.com/r-lib/brio/issues/4


R/reporter-junit.R
```diff
@@ -137,7 +137,7 @@ JunitReporter <- R6::R6Class("JunitReporter",
       } else if (inherits(self$out, "connection")) {
         file <- tempfile()
         xml2::write_xml(self$doc, file, format = TRUE)
-        write_lines(read_lines(file), self$out)
+        self$cat_line(brio::read_lines(file))
```
gaborcsardi, line 140: This should still go to `self$out`, no?

gaborcsardi (approved):
See a comment about `self$out`. 

The big question with this is what we are going to do with windows, if anything. It is not really possible to `sink()` UTF-8 on Windows, you'll get a transliterated output, so non-ASCII snapshots are not portable.

I think we could maybe
* skip the two failing UTF-8 output tests on Windows,
* document that this is not portable.

hadley:
@gaborcsardi can you see why the test didn't previously fail? i.e. why are `brio::write_lines()` and https://github.com/r-lib/testthat/pull/1122/files#diff-89468fcd24514220d6f0b85be3e03b24L8 different?

gaborcsardi:
The difference is not the writing, it is the reading. Previously we were reading file in `"unknown"` encoding, but brio now marks the newly read character vector as UTF-8 (even though it is not UTF-8). So the raw strings are actually still the same, but the encodings are different.

So yeah, these two failing test are actually testing something else than I thought. (Their "works in non-UTF-8 locale" label mislead me, I thought that you are really producing UTF-8 output, which does not work on Windows.)

We can fix these tests, I am not immediately sure how, but if the assumption is that brio is UTF-8 only, then we'll have to make the strings as `"unkown"` manually.

As for the question of not being able to produce UTF-8 output reliably across platforms, I guess we'll just have to document that. (Although I do have a rather involved hack in mind, to force R to output UTF-8 or Windows Unicode.)

gaborcsardi:
> We can fix these tests, I am not immediately sure how, but if the assumption is that brio is UTF-8 only, then we'll have to make the strings as "unkown" manually, I guess.

Or should brio re-encode strings in UTF-8? Or should we re-encode them before writing them out?

R/reporter-junit.R
hadley: Thanks for catching this!

hadley: Oh actually no, because `self$cat_line()` sends it to the right place. Phew!

gaborcsardi: oh, ok.

tests/testthat/test-expect-known.R
```diff
@@ -33,27 +33,23 @@ test_that("igores incomplete last line", {
...
-  expect_success(expect_known_output(cat(text, sep = "\n"), file, update = FALSE))
-  withr::with_locale(
-    c(LC_CTYPE = "C"),
-    {
-      expect_false(l10n_info()$`UTF-8`)
-      expect_success(expect_known_output(cat(text, sep = "\n"), file, update = FALSE))
-    }
-  )
+
+  if (Sys.info()[["sysname"]] == "Windows") {
```
hadley, line 47: @gaborcsardi does this seem like a reasonable approach for now?

gaborcsardi: I think this is mostly fine for the test case. The `"\u00fc"` will (almost) always generate a warning on Windows, because that is transliterated into a latin1 or latin2 character that is cannot appear in UTF-8. But this is just the sign of a general problem.

In general we cannot currently assume that the reference file is in UTF-8. If want it in UTF-8, then we'll need to manually re-encode it, probably before writing it out with brio.

R/expect-known.R
```diff
@@ -88,7 +88,7 @@ compare_file <- function(path, lines, ..., update = TRUE, info = NULL) {
     warning("Reference output is not UTF-8 encoded", call. = FALSE)
   }
 
-  comp <- waldo_compare(lines, enc2native(old_lines), ..., x_arg = "new", y_arg = "old")
+  comp <- waldo_compare(enc2utf8(lines), old_lines, ..., x_arg = "new", y_arg = "old")
```
gaborcsardi, line 91: I think this misleading. I debugged this on Windows, and
* `lines` is apparently already marked as UTF-8 here, because `eval_with_output()` uses brio to read back the output, and brio marks it as UTF-8.
* So `enc2utf8()` does not actually do anything here.
* `lines` is not actually in UTF-8, but in the native encoding, because that's how Windows writes out the output.
* `old_lines` is also marked as UTF-8, because `brio::read_lines()` marks it so, but it is not UTF-8, it is in the native encoding.

The test case for this in `test-expect-known.R` succeeds because both `enc2utf8(lines)` and `old_lines` are in the native encoding and they are both marked as UTF-8.

If the contents of `lines` and/or `old_lines` is not compatible with UTF-8 (like in our own test case), then testthat will warn about the non-UTF-8 contents.

krlmlr:
I updated to latest master and resolved conflicts, but did not take the discussion into account. Tests now pass, we're still seeing warnings on Windows about "Reference output not UTF-8 encoded": https://github.com/r-lib/testthat/pull/1122/checks?check_run_id=2073694504#step:12:41.

R/expect-known.R
```diff
@@ -89,8 +89,8 @@ compare_file <- function(path, lines, ..., update = TRUE, info = NULL) {
   }
 
   comp <- waldo_compare(
-    x = enc2native(old_lines), x_arg = "old",
-    y = lines, y_arg = "new",
+    x = old_lines, x_arg = "old",
+    y = enc2utf8(lines), y_arg = "new",
```
krlmlr, line 93: I agree that we should look into this, we're still seeing warnings on Windows: https://github.com/r-lib/testthat/pull/1122/checks?check_run_id=2073694504#step:12:41.

From @gaborcsardi:

I think this misleading. I debugged this on Windows, and
* `lines` is apparently already marked as UTF-8 here, because `eval_with_output()` uses brio to read back the output, and brio marks it as UTF-8.
* So `enc2utf8()` does not actually do anything here.
* `lines` is not actually in UTF-8, but in the native encoding, because that's how Windows writes out the output.
* `old_lines` is also marked as UTF-8, because `brio::read_lines()` marks it so, but it is not UTF-8, it is in the native encoding.

The test case for this in `test-expect-known.R` succeeds because both `enc2utf8(lines)` and `old_lines` are in the native encoding and they are both marked as UTF-8.

If the contents of `lines` and/or `old_lines` is not compatible with UTF-8 (like in our own test case), then testthat will warn about the non-UTF-8 contents.

hadley:
Fixes #1516

hadley:
Still seeing the warnings:

```
-- Warning (test-expect-known.R:50:5): works with utf-8 output -----------------
New reference output is not UTF-8 encoded
Backtrace:
1. testthat::expect_known_output(cat(text, sep = "\n"), "test-expect-known-win.txt") test-expect-known.R:50:4
2. testthat:::compare_file(...) D:/a/testthat/testthat/check/testthat.Rcheck/00_pkg_src/testthat/R/expect-known.R:68:2
-- Warning (test-expect-known.R:50:5): works with utf-8 output -----------------
Reference output is not UTF-8 encoded
Backtrace:
1. testthat::expect_known_output(cat(text, sep = "\n"), "test-expect-known-win.txt") test-expect-known.R:50:4
2. testthat:::compare_file(...) D:/a/testthat/testthat/check/testthat.Rcheck/00_pkg_src/testthat/R/expect-known.R:68:2
```

hadley:
I think it's easiest to just skip this test on windows since there's nothing meaningful we can do here.

PR #1141, opened 2020-08-17
Parallel progress reporter

### LINKED ISSUE #1111: Custom parallel reporter
* Should report which files are being run in progress bar that remains at bottom of output
* Spinner runs continuously to show you that its working
* On completion of file, show summary of tests, along with any problems, and how long it took to run
* Exploring event throttling to make sure it's not adding unnecessary overhead

gaborcsardi:
Still needed: 

- [x] final status line for each file
- [x] smoother spinner by adjusting the poll interval
- [x] make it default for parallel interactive tests
- [x] mark other reporters for parallel support (e.g. check, but not debug, etc.)
- [x] refactoring
- [x] colored status bar (needs the fansi package, probably)
- [x] ~~tests~~ We can do these later... 

Closes #1111


R/parallel.R
```diff
@@ -51,6 +51,9 @@ test_files_parallel <- function(
   # TODO: detect number of CPUs
   num_workers <- min(getOption("Ncpus", 4), length(test_paths), 4)
 
+  message("Starting work pool of ", num_workers,
```
hadley, line 54: Maybe "Starting x processes for parallel testing" ?

R/reporter-progress.R
```diff
@@ -308,6 +338,113 @@ CompactProgressReporter <- R6::R6Class("CompactProgressReporter",
...
+
+    update = function(force = FALSE) {
+      if (!force && !self$should_update()) return()
+      self$spin_frame <- self$spin_frame + 1L
+      status <- spinner(self$frames, self$spin_frame)
+
+      message <- paste(
+        status,
+        summary_line(self$n_ok, self$n_fail, self$n_warn, self$n_skip),
+        if (length(self$files) > 0) "@" else " starting up...",
```
hadley, line 438:
```suggestion
        if (length(self$files) > 0) "@" else "Starting up...",
```

(no space needed since you use `paste()` on the next line)

R/parallel.R
gaborcsardi: Yeah. I think if we keep the processes, then we'll need some terminology about them, this is why I started using "work pool", but yeah, it is not great, and we can change it later.

gaborcsardi: @hadley Hmmm, the edges case of "Starting 1 process for parallel testing" is weird. How about just "Starting x test processes"?

hadley: Actually, when will that occur? Why would we do parallel testing with one process?

gaborcsardi: E.g. when you are starting to write tests for a new package, or more often when you do `filter = "watcher"`. It would be annoying to switch back and forth between parallel and non-parallel modes based on how many files are matching.

gaborcsardi:
> the last status update hasn't been fully cleared

Ah, this is an RStudio thing again, I think. Let me try to find it.

R/parallel.R
hadley: Oh yeah, makes sense.

gaborcsardi:
Yeah, the bad output is because of this RStudio bug: https://github.com/rstudio/rstudio/issues/7649

We can work around it by clearing the last line first, instead of overwriting it, but unfortunately that creates flickering.

Another issue is that RStudio does not keep the last line of the screen as the last line, even though there is no newline emitted. This creates more flickering: https://github.com/rstudio/rstudio/issues/7650

gaborcsardi:
@hadley OK, updated the startup message, and worked around the display bug.

hadley:
Looks good, let's merge it 😄

PR #1146, opened 2020-08-23
Detect number of CPUs to use

### LINKED ISSUE #1114: Better control over number of workers
* Parameter? Option?
* Better default detecton

gaborcsardi:
This is on top of #1141.

Docs will be in the vignette, I think.

Closes #1114.


R/parallel.R
```diff
@@ -65,42 +67,109 @@ test_files_parallel <- function(
...
 }
 
+default_num_cpus <- function() {
+  # Use common option, if set
+  ncpus <- getOption("Ncpus", NULL)
+  if (!is.null(ncpus) && !is_integer(ncpus)) {
+    stop("`getOption(Ncpus)` must be integer")
+  }
+
+  if (!is.null(ncpus)) return(ncpus)
```
hadley, line 92: Can you please put this in `{}` to match tidyverse style?

R/parallel.R
```diff
@@ -65,42 +67,109 @@ test_files_parallel <- function(
...
+
+  # Otherwise detect
+  ncpus <- ps::ps_cpu_count()
+
+  # But allow capping with an env var
+  ev <- "TESTTHAT_MAX_CPUS"
+  max_env <- Sys.getenv(ev, "")
+  if (max_env != "") {
+    max_env <- as.integer(max_env)
+    if (is.na(max_env)) stop(ev, " must be an integer")
```
hadley, line 102:
```suggestion
    if (is.na(max_env)) abort(paste0(ev, " must be an integer"))
```

R/parallel.R
```diff
@@ -65,42 +67,109 @@ test_files_parallel <- function(
...
+  }
+
+  if (!is.null(ncpus)) return(ncpus)
+
+  # Otherwise detect
+  ncpus <- ps::ps_cpu_count()
+
+  # But allow capping with an env var
+  ev <- "TESTTHAT_MAX_CPUS"
+  max_env <- Sys.getenv(ev, "")
```
hadley, line 99:
```suggestion
  max_env <- Sys.getenv("TESTTHAT_MAX_CPUS", "")
```
?

R/parallel.R
```diff
@@ -65,42 +67,109 @@ test_files_parallel <- function(
...
+  # Use common option, if set
+  ncpus <- getOption("Ncpus", NULL)
+  if (!is.null(ncpus) && !is_integer(ncpus)) {
+    stop("`getOption(Ncpus)` must be integer")
+  }
+
+  if (!is.null(ncpus)) return(ncpus)
+
+  # Otherwise detect
+  ncpus <- ps::ps_cpu_count()
```
hadley, line 95:
```suggestion
  ncpus <- ps::ps_cpu_count() - 1L
```
Just to be polite?

R/parallel.R
gaborcsardi: I did not want to repeat it twice, it is used below as well.

R/parallel.R
gaborcsardi: Actually, I think it is completely fine to take all CPUs. If this behavior bothers you because you regularly run some CPU-intensive programs on your dev machine, then you can set the new env var to cap the CPUs?

hadley: Ok, fine with me.

gaborcsardi:
Actually, I forgot the case when ps is not supported. Will add that, but nevertheless we need to test testthat on Solaris before release.

Issue #1154, opened 2020-08-27
allow callers to 'detect' skips without requiring reporter


kevinushey:
Right now, skips are caught and swallowed within a `tryCatch()` handler:

https://github.com/r-lib/testthat/blob/9e643d8ee9772c58d5a6bf0846e16bf78dde59f9/R/test-that.R#L193-L194

However, it'd be useful if that condition could be re-signaled for other entities (other than reporters) which might want to react to that signal. That is, I'd like for the following to work:

```
library(testthat)

skipped <- FALSE
withCallingHandlers(
  test_that("skips can be handled", skip("skip")),
  skip = function(e) skipped <<- TRUE
)

identical(skipped, TRUE)
```

kevinushey:
The simplest solution here (I think) would be to allow that `skip` handler here to re-signal the condition to any other listeners.

gaborcsardi:
I think you can already do this by creating your own reporter, that defines its own `add_results()` method, which can detect skips, and then just call a proper reporter.

kevinushey:
Is there a way to create a new reporter that inherits from the "default" reporter (whatever that might be when tests are run)? Or to get an instance of the 'default' reporter and then monkey-patch its `add_results()` method?

Being able to just "see" skips via `withCallingHandlers()` ultimately seems simpler.

gaborcsardi:
Yeah, e.g. here is a factory that takes a reporter, and then creates a class that inherits from it: https://github.com/r-lib/ps/blob/61fcede3fbbe2f1f124561ec62638550e4226f3e/R/testthat-reporter.R#L74

You can also use `get_reporter()` and then monkey patch the return value.

IMO re-throwing conditions does not seem like a good solution, AFAICT skip conditions are not officially part of the API, so one could argue that they are an implementation detail. This solution also might not work for you when testthat runs the tests in sub-processes, because these conditions will not pass form the sub-process to the main process, even if you re-throw them.

kevinushey:
> This solution also might not work for you when testthat runs the tests in sub-processes, because these conditions will not pass form the sub-process to the main process, even if you re-throw them.

That's a good point I hadn't considered. If it interferes with parallel execution of tests, then I think it's a non-starter.

gaborcsardi:
It does not _interfere_ per se, your PR won't break anything AFAICT, it just won't work, because it will re-throw the skip in the sub-process, and it will not get to the main process. (If you want to catch them in the main process.)

[...]

PR #1155, opened 2020-08-27
re-signal skip conditions when caught

kevinushey:
Closes https://github.com/r-lib/testthat/issues/1154.


hadley:
This doesn't match my mental model of conditions, which always bubble up and don't need to be manually resignalled. Maybe @lionel- understands it better than me.

kevinushey:
I think this is just because `tryCatch()` stops propagation of signals? For example:

```
withCallingHandlers(
  
  tryCatch(
    
    withCallingHandlers(
      
      # expression
      {
        cond <- structure(
          list(message = "test"),
          class = c("skip", "condition")
        )
        stop(cond)
      },
      
      # inner withCallingHandlers() handler
      skip = function(e) {
        writeLines("inner handler")
      }
      
    ),
    
    # tryCatch() handler
    skip = function(e) {
      writeLines("tryCatch handler")
    }
  ),
  
  # outer withCallingHandlers() handler
  skip = function(e) {
    writeLines("outer handler")
  }
  
)
```

So if anything "above" the `tryCatch()` block wants to see a signal, if `tryCatch()` catches it, it needs to re-signal it.

kevinushey:
@gaborcsardi pointed out in https://github.com/r-lib/testthat/issues/1154#issuecomment-682213637 that exposing this could interfere with the way parallel tests are run, so ultimately the right way to do this is via a custom reporter.

PR #1177, opened 2020-09-10
update expect_snapshot_file() docs

maelle:
the example in the vignette is for foo.R and I'm using it for a JSON file.

### DIFF
--- a/R/snapshot-file.R
+++ b/R/snapshot-file.R
@@ -3,7 +3,7 @@
 #' @description
 #' Whole file snapshot testing is designed for testing objects that don't have
 #' a convenient textual representation, with initial support for images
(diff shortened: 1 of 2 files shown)


R/snapshot-file.R
```diff
@@ -3,7 +3,7 @@
 #' @description
 #' Whole file snapshot testing is designed for testing objects that don't have
 #' a convenient textual representation, with initial support for images
-#' (`.png`, `.jpg`, `.svg`) and data frames (`.csv`).
```
hadley, line 6: Data frames are specifically supported, so can you please add an additional clause for text files?

maelle: :+1:

hadley:
Thanks!

PR #498, opened 2016-07-04
Report bare expectations

krlmlr:
This makes sure that all testing code is run in a reporter and with the usual testthat condition handling. Fixes #427.

CC @kevinushey.

PR #1183, opened 2020-09-27
Restore `test_dir(..., wrap = FALSE)`

brodieG:
In some unusual cases it is useful to be able to preserve the
original `testthat` behavior of not wrapping expressions outside
of `test_that` blocks.

In commit 37cc0d04fc497edcf62fe44729c73e3c77559b2e 
control of the  `wrap` functionality was removed from `test_dir`,
but after discussion in #498 it was restored.

More recently, b13ecb38c7ba49d48d12e3c97dcfa081a9634bba
removed the `wrap` functionality again from `test_dir`.  This patch
seeks to restore the functionality for serial tests only as restoring it
for parallel tests is more complicated and of dubious value since
a) the parameter is now deprecated, and b) parallel tests did not
exist prior to deprecation.

Bonus typo fixes.

### DIFF
--- a/R/test-files.R
+++ b/R/test-files.R
@@ -191,7 +191,7 @@ test_files_serial <- function(test_dir,
                        env = NULL,
                        stop_on_failure = FALSE,
                        stop_on_warning = FALSE,
-                       wrap = FALSE,
+                       wrap = TRUE,
                        load_package = c("none", "installed", "source")) {
 
   env <- test_files_setup_env(test_package, test_dir, load_package, env)
@@ -264,7 +264,7 @@ test_files_check <- function(results, stop_on_failure = TRUE, stop_on_warning = 
   invisible(results)
 }
 
-test_one_file <- function(path, env = test_env(), wrap = FALSE) {
+test_one_file <- function(path, env = test_env(), wrap = TRUE) {
   reporter <- get_reporter()
   on.exit(teardown_run(), add = TRUE)
 
--- a/R/test-files.R
+++ b/R/test-files.R
@@ -82,10 +82,6 @@ test_dir <- function(path,
     abort("No test files found")
   }
 
-  if (!missing(wrap)) {
-    lifecycle::deprecate_warn("3.0.0", "test_dir(wrap = )")
-  }
-
   want_parallel <- find_parallel(path, load_package, package)
 
   if (is.null(reporter)) {
@@ -107,6 +103,7 @@ test_dir <- function(path,
     env = env,
     stop_on_failure = stop_on_failure,
     stop_on_warning = stop_on_warning,
+    wrap = wrap,
     load_package = load_package,
     parallel = parallel
   )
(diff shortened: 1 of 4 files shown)


brodieG:
Hmm, I'll need to look into why the CI tests are failing despite having
worked locally for me.

In the meantime, @hadley any feedback on whether the concept of 
this patch is okay for your or not would be useful for me to know.

R/test-files.R
```diff
@@ -150,10 +163,18 @@ test_files <- function(test_dir,
                        env = NULL,
                        stop_on_failure = FALSE,
                        stop_on_warning = FALSE,
+                       wrap = lifecycle::deprecated(),
```
hadley, line 166: Just make this `TRUE` and then warn if `wrap` is `FALSE`

R/test-files.R
```diff
@@ -98,6 +94,22 @@ test_dir <- function(path,
   reporter <- find_reporter(reporter)
   parallel <- want_parallel && reporter$capabilities$parallel_support
 
+  if (parallel) {
+    if (!is_missing(wrap)) {
```
hadley, line 98: Given that you're the only person using this argument, I think you can just keep the warning the same for both serial and parallel.

hadley (commented):
Overall, it's fine, but I made a couple of suggestions for simplifying it, since you're the only person who needs it.

R/test-files.R
brodieG: Ok, I did this.  If for whatever reason someone ends up using `wrap=FALSE` in parallel mode they might be confused, but given they will get a deprecation warning hopefully they'll figure it out, and more likely such a person won't exist anyway.

R/test-files.R
brodieG: Done.

brodieG:
Note that this will conflict with #1184, so I will submit one more commit to remove my partial match fix.

In re failing tests, I looked at this for a bit, and it's going wrong with errors of this form (among others):

```
── Failure (Line 10): error in parallel setup code ─────────────────────────────
`err` inherits from `rlib_error/error/condition` not `testthat_process_error`.

── Failure (Line 11): error in parallel setup code ─────────────────────────────
err$message does not match "Error in setup".
Actual value: "R session finished"
```
Or:
```
── Failure (Line 11): error in parallel setup code ─────────────────────────────
err$message does not match "Error in setup".
Actual value: "R session not ready yet"
```
I could not reproduce locally on OS X R4.0.2, but did reproduce it on Ubuntu 18.04 with r-devel.  The error seems related to the `callr` sessions.  I'm pretty sure this is not my doing as I reproduced the error with the HEAD version of `testthat`, and the errors appear in #1184 as well.

hadley:
Thanks!

brodieG:
Update here: I no longer rely on the ability to turn `wrap` off.  I know this is something you'd been hoping to remove, so just letting you know that my packages will no longer be affected if you do deprecate it.

PR #1184, opened 2020-09-28
Fix partial match

jimhester:


jimhester:
`options(warnPartialMatchDollar = TRUE)` makes it pretty easy to catch these, though they rarely cause problems in practice of course.

gaborcsardi:
Thanks!

brodieG:
Note also fixed in #1183, so we'll have to reconcile on merge.

hadley:
Thanks!

PR #1186, opened 2020-10-01
Change spinner color to red if we know there are failures

krlmlr:
Ideally we would print errors as we see them, to give as early feedback as possible. If a test file runs 10 seconds, that's precious time wasted that could be spent reviewing the error instead.

### DIFF
--- a/R/reporter-progress.R
+++ b/R/reporter-progress.R
@@ -129,9 +129,9 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
         }
         status <- spinner(self$frames, data$n)
         if (data$n_fail > 0) {
-          status <- crayon::red(status)
+          status <- colourise(status, "failure")
         } else if (data$n_warn > 0) {
-          status <- crayon::magenta(status)
+          status <- colourise(status, "warning")
         }
       }
 
--- a/R/reporter-progress.R
+++ b/R/reporter-progress.R
@@ -130,6 +130,8 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
         status <- spinner(self$frames, data$n)
         if (data$n_fail > 0) {
           status <- crayon::red(status)
+        } else if (data$n_warn > 0) {
+          status <- crayon::magenta(status)
         }
       }
 
--- a/R/reporter-progress.R
+++ b/R/reporter-progress.R
@@ -128,6 +128,9 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
           return()
         }
         status <- spinner(self$frames, data$n)
+        if (data$n_fail > 0) {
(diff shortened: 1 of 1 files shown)


R/reporter-progress.R
```diff
@@ -191,7 +194,6 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
       } else if (expectation_skip(result)) {
         self$n_skip <- self$n_skip + 1
         self$ctxt_n_skip <- self$ctxt_n_skip + 1
-        self$ctxt_issues$push(result)
```
gaborcsardi, line 194: Why would you remove this?

gaborcsardi:
I don't mind the red, but you can actually see the number of errors on the last line already, no?

R/reporter-progress.R
krlmlr: Unrelated, but yes, I would remove this.

The skips are shown in the summary, that's more than enough. I have 100-s of skips in some packages, I don't care about them unless I do -- and then I go find them.

krlmlr:
My entire attention is focused on the spinner :upside_down_face:

krlmlr:
What do you think about printing the errors as we see them?

R/reporter-progress.R
hadley: I that seems reasonable, but lets check with @jennybc first.

hadley: How would you know what file the skips occurred in?

krlmlr: I would search for the skip message.

jennybc: I think I would be happier (or at least just as happy) seeing the skips at the end. I don't quite follow what's proposed for following up on skips, though, if they are unexpected.

hadley:
I'm worried that this is too late in the release cycle to make such a big change. I'm happy to consider for testthat 3.0.1.

krlmlr:
Would you accept this PR? I can fix the tests early next week.

hadley:
If you remove the change to skips

R/reporter-progress.R
```diff
@@ -130,6 +130,8 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
         status <- spinner(self$frames, data$n)
         if (data$n_fail > 0) {
           status <- crayon::red(status)
+        } else if (data$n_warn > 0) {
+          status <- crayon::magenta(status)
```
hadley, line 134: Don’t we have an existing function that converts status to colour?

hadley: Ah yes, `testthat_style()`

krlmlr:
Builds are green now.

PR #1187, opened 2020-10-02
Fix parallel cleanup: ignore failed workers

gaborcsardi:
If a worker has crashed or exited, then we cannot
run the teardown files in it. Now we ignore such
workers, so we at least tear down the rest.

This also fixes test failures with newer withr,
that fails if the deferred queue teardown errors.


hadley:
Thanks!

PR #1189, opened 2020-10-03
Works around test failure with warnPartialMatchArgs

### LINKED ISSUE #1188: Test case failure on older R versions because `options(warnPartialMatchArgs = NULL)` fails
Before R 3.6.0 
```
options(warnPartialMatchArgs = NULL)
```
errors, and now withr does not ignore failures in deferred code, so we get an error when we use 
```
  withr::local_options(list(warnPartialMatchArgs = FALSE))
```
in a test case. 

I'll submit a PR with a workaround in a minute.

gaborcsardi:
R 3.5.3 and older fails for
```r
options(warnPartialMatchArgs = NULL)
```
and now that withr does not ignore failures, our test
case that sets
```r
  withr::local_options(list(warnPartialMatchArgs = FALSE))
```

We work around this by setting the default value explicitly on
these R versions.

Closes #1188.


hadley:
Thanks!

PR #1192, opened 2020-10-05
Parallel tests should require the 3rd edition

hadley:


R/parallel-config.R
```diff
@@ -14,5 +14,18 @@ find_parallel <- function(path, load_package = "source", package = NULL) {
...
+  par <- identical(
+    toupper(desc$get_field("Config/testthat/parallel", default = "FALSE")),
+    "TRUE"
+  )
+
+  if (par) {
+    ed <- as.integer(desc$get_field("Config/testthat/edition", default = 2L))
+    if (ed < 3) {
+      inform("Running tests in parallel requires the 3rd edition")
+      want_parallel <- FALSE
```
hadley, line 26: Should be `return(FALSE)` ?

gaborcsardi: oh, sorry, it should be 
```
par <- FALSE
```

good catch.

gaborcsardi: should this be a warning, btw?

gaborcsardi: Anyway, should be good now.

hadley: I don't feel too strongly about it either way

gaborcsardi: Me neither, so let's leave it as is. :)

PR #1194, opened 2020-10-09
Make skips silent by default, except skip_if_offline()

krlmlr:
Skips are deliberate and usually do not warrant immediate attention. It should be sufficient to list them by reason in the summary, a code search for the reason or for `skip_` reveals their location of needed.


gaborcsardi:
I think people use skips for two different things. One is an expected skip, e.g. if a test case does not work on windows. The second one is an unexpected skip, e.g. when you are offline for a test that needs the internet.

I would think that people would not want to see the first kind of skips, but they do want to see the second kind of skips.

One solution would be to get rid of the first kinds of skips. E.g. instead of 

```r
skip_on_os("linux")
```

one would write

```r
if (is_os("linux")) return(expect_true(TRUE))
```

This does add an artificial expectation, unfortunately, because otherwise a `test_that()` block without an expectation will also report a skip.

krlmlr:
How about `skip(report = TRUE)` and e.g. `skip_on_os()` forwarding to `skip(report = FALSE)` ?

krlmlr:
Added a `silent = TRUE` argument to `skip()` . The information is propagated through the `"skip_silent"` subclass and through an attribute to the `message` component of an expectation.

hadley:
I like the approach in #1209 better; will discuss there.

PR #1209, opened 2020-10-27
Introduce testthat.progress.verbose_skips option

krlmlr:
so that skips can be hidden from progress output.

Closes #1194.


hadley:
How about we make this `FALSE` by default, but include a message that tells you how to turn it on (or otherwise access the skips) if any skips are seen?  That might require a mechanism other than an option, since it'll be hard to feed an option into the way that most people run tests in RStudio.

hadley:
Another idea would be to store the save the skips in a separate file — then we'd display a message showing how to find it. The advantage of a separate file would be that you could (if you wanted) check it into git, which would make it easier to see if the skips changed significantly in your development environment.

krlmlr:
I'd very much prefer the default to be `FALSE`. Can we get rid of this option completely?

I think in many cases the easiest way to find the source for a `skip()` is to search your own code for the skip message.

Exceptions:

- Functions like `skip_if_not_installed()` -- maybe they should be summarized as the code that called them so that they can be searched for easier?
- Custom skip functions: same. Should `vignette("skipping")` give advice on naming the skip message in such a case?
- Cases like DBItest where the `skip()` is not in code you control: same, the skip message should be adjusted.

We could do something like this:

```
Skipped tests:

── Skipped tests  ─────────────────────────────────────
● Not yet implemented (3)
● skip_if_not_installed("RPostgres"): package not installed
● skip_if_no_database(): can't connect to the database
● skip_if_offline(): no internet connection (1257)
● DBItest::skip_by_request(): data_logical, data_raw (1)
● DBItest::skip_by_request(): list_objects_features (1)

To find the source for a skip, search your code for e.g. "Not yet implemented" or 'skip_if_not_installed("RPostgres")'.
```

hadley:
Lets discuss this more more in the linked issue before committing to an implementation. I want to get a patch release out tomorrow, but we can plan for another release early Jan.

krlmlr:
There is no linked issue, #1194 is a previous pull request.

hadley:
I meant #1267

R/reporter-progress.R
```diff
@@ -286,6 +291,10 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
       }
       self$last_update <- time
       TRUE
+    },
+
+    get_verbose_skips = function() {
```
hadley, line 296: To keep this as simple as possible, lets drop the accessor function

krlmlr: @wch: Can I access public member variables from a subclass? It doesn't appear so, in this file in `ParallelProgressReporter$add_result()` I only see public functions from `super$`, and https://r6.r-lib.org/articles/Introduction.html#inheritance-1 doesn't talk about accessing fields from subclasses.

wch: @krlmlr You should be able to access a public field from a subclass.

PR #1238, opened 2020-11-18
setup*.R instead of setup-?

maelle:


hadley:
Sure, you can drop the `-`, but I don't recommend it.

PR #1239, opened 2020-11-18
fix typo

maxheld83:


hadley:
This isn't a typo; it's what the internal usethis function is called.

PR #1242, opened 2020-11-19
remove get_proj() from case study

maxheld83:
I'm not sure about the use inside usethis, but when I try this locally inside `devtools::check()` this fails, because there is no project inside the temp dir where the check is run.
So this might be a bit misleading as an example.


hadley:
I don't think the point is to have running code here, but to illustrate the idea.

PR #1243, opened 2020-11-23
Use pak in GHA installations

hadley:


gaborcsardi:
@hadley I fixed the pak/pkgdepends issue with the circular dependency, so this is good now.

hadley:
@gaborcsardi thanks!

Issue #1244, opened 2020-11-24
Installation error: invalid token in expression


sigmafelix:
Hello, I tried to install `testthat` in my machine, but it prints errors as below:

```
* installing *source* package ‘testthat’ ...
file ‘inst/include/testthat.h’ has the wrong MD5 checksum
** using staged installation
** libs
g++ -std=gnu++11 -I"/usr/local/lib/R/include" -DNDEBUG -I../inst/include -DCOMPILING_TESTTHAT  -I/usr/local/include   -fPIC  -g -O2  -c test-catch.cpp -o test-catch.o
test-catch.cpp:19:5: error: invalid token in expression
    expect_true((4 + 4) == 8);
    ^
../inst/include/./testthat/testthat.h:175:22: note: expanded from macro 'expect_true'
# define expect_true CATCH_CHECK
                     ^
../inst/include/./testthat/vendor/catch.h:11377:29: note: expanded from macro 'CATCH_CHECK'
#define CATCH_CHECK( expr ) INTERNAL_CATCH_TEST( "CATCH_CHECK", Catch::ResultDisposition::ContinueOnFailure, expr )
                            ^
../inst/include/./testthat/vendor/catch.h:2207:9: note: expanded from macro 'INTERNAL_CATCH_TEST'
        INTERNAL_CATCH_REACT( __catchResult ) \
        ^
../inst/include/./testthat/vendor/catch.h:2191:44: note: expanded from macro 'INTERNAL_CATCH_REACT'
    if( resultBuilder.shouldDebugBreak() ) CATCH_BREAK_INTO_DEBUGGER(); \
                                           ^
../inst/include/./testthat/vendor/catch.h:2133:75: note: expanded from macro 'CATCH_BREAK_INTO_DEBUGGER'
    #define CATCH_BREAK_INTO_DEBUGGER() if( Catch::isDebuggerActive() ) { CATCH_TRAP(); }
                                                                          ^
../inst/include/./testthat/vendor/catch.h:2111:38: note: expanded from macro 'CATCH_TRAP'
        #define CATCH_TRAP() __asm__("int $3\n" : : )
                                     ^
<inline asm>:1:6: note: instantiated into assembly here
        int $3
            ^
test-catch.cpp:19:5: error: invalid operand
    expect_true((4 + 4) == 8);
    ^
../inst/include/./testthat/testthat.h:175:22: note: expanded from macro 'expect_true'
# define expect_true CATCH_CHECK
                     ^
[...]

PR #1246, opened 2020-11-25
Fix ARM64 detection on Apple platforms in catch.h

tzakharko:
Old catch.h header assumed that all Mac platforms are x86-based, which would lead to compile errors on Apple Silicon Macs. This change back ports the definition of  `CATCH_TRAP` used in newer Catch2, which correctly accounts for the ARM-based Apple platforms.

All tests pass on my Intel and my M1-based Mac (except the X11 test since I don't have X11 installed on the later). 

This closes issue https://github.com/r-lib/testthat/issues/1244


gaborcsardi:
Cool, thanks!

PR #1250, opened 2020-11-30
skip_if_not(): do not append " is not TRUE" to custom message

### LINKED ISSUE #1247: double negation in skip_if_not() message
The `message` in `skip_if_not()` is always appended with " is not TRUE". This makes sense for the `deparse(substitute(condition))` default. It may not for custom messages. 

Case in point is the googledrive package, where `skip_if_not(drive_has_token(), "No Drive token")` [results in](https://github.com/tidyverse/googledrive/runs/1391398269?check_suite_focus=true#step:11:36) "No Drive token is not TRUE".

This can be easily fixed IMHO if `skip_if_not()` is modelled after `skip_if()`, i.e. if custom messages are shown as is and only the default message is "{condition} is not TRUE".

I can make a PR, if you agree.

dpprdan:
fixes #1247 

I noticed that `skip_if()`, `skip_if_translated()` and `skip_on_os()` are missing in the `@helpers` section. Should I add them there, while I am at it?

Should/could there be a `skip_on_github_actions()` (in a separate PR)?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `skip_if_not()` no longer appends "is not TRUE" to custom messages 
+  (@dpprdan, #1247).
+
 * `expect_snapshot_value()` now passes `...` on to `waldo::compare()` (#1222).
 
 * Fixed a couple of buglets in `auto_test_pacakge()` (@mbojan, #1211, #1214)
--- a/R/skip.R
+++ b/R/skip.R
@@ -78,7 +78,7 @@ skip_empty <- function() {
 #'   `FALSE`, `skip_if()` will skip if `TRUE`.
 skip_if_not <- function(condition, message = NULL) {
   if (is.null(message)) {
-    message <- paste(deparse(substitute(condition)), " is not TRUE")
+    message <- paste0(deparse(substitute(condition)), " is not TRUE")
   }
   if (!isTRUE(condition)) {
     skip(message)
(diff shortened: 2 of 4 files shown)


hadley:
A doc fix would be great. There's no need for a `skip_on_github_actions()` since you can just use `skip_on_ci()`. The docs probably need to be updated to reflect that we no longer recommend `skip_on_travis()` or `skip_on_appveyor()`.

dpprdan:
> The docs probably need to be updated to reflect that we no longer recommend skip_on_travis() or skip_on_appveyor().

I am not entirely sure what you mean and hence how to handle this. I first read it as "[no longer recommend Travis CI](https://ropensci.org/technotes/2020/11/19/moving-away-travis/)", but maybe you just literally mean the functions? But then I wouldn't know they are not recommended anymore. Or, if you mean the CI services, why Appveyor is no longer recommended. And, no matter what the reason is, why then not deprecate both functions, if `skip_on_ci()` basically does the same thing? So I guess I am skipping this for now. 😄 

Which means I am done here, unless you'd like me to add or change something else.

dpprdan:
It seems [you just meant the functions](https://github.com/r-lib/testthat/issues/970#issuecomment-610652026). My (soft) deprecate point still stands. 😃

hadley:
Deprecating implies that you need to stop using them for existing code, which I don't think is true, and would cause a lot of hassle for little gain.

hadley:
Thanks! I did a bit more polishing of the docs.

PR #1252, opened 2020-12-01
Unset RSTUDIO envvar

### LINKED ISSUE #1225: Unset `RSTUDIO` instead of setting it to `"0"`?
Is slightly better I think. Cf. https://github.com/r-lib/crayon/pull/101

hadley:
Fixes #1225


gaborcsardi (approved):
Thanks!

PR #1254, opened 2020-12-02
Docs about testing custom skippers

### LINKED ISSUE #1240: expect_error or expect_condition for testing custom skippers?
- In the vignette `expect_error()` is recommended however if one uses it it displays a message about the skipping reason (in the tests results)

- In testthat tests itself, `expect_condition()` is used. 

https://github.com/r-lib/testthat/blob/537ff4bff3db2076e93f398339db881b7ad72981/R/expect-self-test.R#L50-L52

Should the vignette be updated? Happy to make a PR if it is the case.

maelle:
Fix #1240

I am actually not sure I understand the second test.

For me at least it would be clearer for the example to define a custom skipper then test it in two different conditions? I have https://github.com/ropensci/vcr/blob/master/tests/testthat/test-skip-vcr-off.R

### DIFF
--- a/vignettes/skipping.Rmd
+++ b/vignettes/skipping.Rmd
@@ -60,12 +60,12 @@ skip_if_Tuesday <- function() {
 
 It's important to test your skip helpers because it's easy to miss if you're skipping more often than desired, and the test code is never run. This is unlikely to happen locally (since you'll see the skipped tests in the summary), but is quite possible in continuous integration.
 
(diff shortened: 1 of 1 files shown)


hadley:
Yeah, it might be better to make a custom skip function that uses some made up env var?

maelle:
I made a proposal of example. I changed the first custom skipper definition to make it look like the internal skippers, that are then easier to test?

I know it might be out-of-scope for the vignette but would it make sense to give a hint about how one would test `skip_on_tuesday()` (maybe just linking to mockery docs).

hadley:
Looks good, thanks @maelle!

PR #1258, opened 2020-12-02
Set language to English

### LINKED ISSUE #1213: Parallel tests use different language
I have `Config/testthat/parallel: true` in my DESCRIPTION

```r
data_uris <- function(..., mime = '', encoding = 'base64', files) {
	stopifnot(length(list(...)) == 0L)
	vapply(
		files,
		function(f) base64enc::dataURI(mime = mime, encoding = encoding, file = f),
		character(1L))
}

test_that('Extra arguments to data_uris error', {
	expect_error(
		data_uris('one', mime = 'text/html'),
		'length(list(...)) == 0L is not TRUE',
		fixed = TRUE
	)
})
```

When I run it (edit: with german locale set), I get

```
Error (test_utils.r:3:2): Extra arguments to data_uris error
Error: length(list(...)) == 0L ist nicht TRUE
Backtrace:
 1. testthat::expect_error(...)
 7. repr:::f()
 8. repr::data_uris("one", mime = "text/html")
 9. base::stopifnot(length(list(...)) == 0L) R/utils.r:137:8
```

hadley:
@MichaelChirico does this seem ok to you? (Since you're the only person I know who's testing translations). R CMD check already sets the language to English so we don't think this should affect much existing code.

Fixes #1213


R/local.R
```diff
@@ -101,7 +101,11 @@ local_reproducible_output <- function(width = 80,
     rlang_interactive = FALSE,
     .local_envir = .env
   )
-  withr::local_envvar(RSTUDIO = NA, .local_envir = .env)
+  withr::local_envvar(
+    RSTUDIO = NA,
+    LANGUAGE = "en",
```
gaborcsardi, line 106: Should we include a sentence about this in the list of things `local_test_context()` sets, up about line 44? 

Btw. that line also needs a fix for `RSTUDIO`, which is now unset.

MichaelChirico:
IINM it will work on Unix, but probably (?) not on Windows (though I do see a ✅ for the Windows GHA). We had some headaches getting non-English tests to work on Windows... see https://github.com/Rdatatable/data.table/pull/4524.

I'm not sure if it's any easier a problem to specifically force English than it was for us to temporarily set Chinese.

For reference I think you're referring to this from [WRE 1.3.1](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Checking-packages):

> All these tests are run with collation set to the C locale, and for the examples and tests with environment variable LANGUAGE=en: this is to minimize differences between platforms.

There's also this in [R administration 2.8](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Testing-a-Unix_002dalike-Installation) about testing R installation:

```
Sys.setenv(LC_COLLATE = "C", LC_TIME = "C", LANGUAGE = "en")
```

And I think our troubles were related to this note in [R administration 7.2](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Localization-of-messages):

> It is usually possible to change the language once R is running via (not Windows) Sys.setlocale("LC_MESSAGES", "new_locale"), or by setting an environment variable such as LANGUAGE, provided34 the language you are changing to can be output in the current character set. But this is OS-specific, and has been known to stop working on an OS upgrade.

hadley:
I also have a vague recollection that the translation is cached, so that after an error message has been processed once, changing `LANG` doesn't have any effect.

hadley:
Running the testthat revdeps now, just to double check that this won't accidentally break anything.

hadley:
I individually checked the four failures — they seem unrelated to the language setting.

PR #1259, opened 2020-12-02
Rstudio -> RStudio

kevinushey:
Fixes the capitalization of RStudio for the RStudio reporter.

(definitely understand if you'd prefer to just leave it as-is but this is just one of the small things that bugs me ...)


hadley:
I think you can just rename, if you want: https://github.com/search?q=org%3Acran+RStudioReporter&type=code.

hadley:
And not used in the IDE itself: https://github.com/rstudio/rstudio/search?q=RstudioReporter&type=code

hadley:
Thanks!

PR #1278, opened 2020-12-11
Show name of test when snapshots are reset

krlmlr:
and also warn when snapshots are discarded.

Do we need to always warn with duplicate tests (`is.null()` vs. `length() == 0`)?


NEWS.md
```diff
@@ -1,12 +1,16 @@
...
+* A warning is given when snapshots are discarded due to reusing the same
+  test name (#1278, @krlmlr).
+
 * New `testthat.progress.verbose_skips` option. Set to `FALSE` to stop
   reporting skips as they occur, they will still appear in the summary
   (#1209, @krlmlr).
 
 * `CheckReporter` now always shows the full test name (#1268).
 
 * Catch tests are no longer reported multiple times (#1237).
```
hadley, line 12: Do you mind deleting this bullet since it's repeated below and problem comes from a merge?

R/snapshot-reporter.R
```diff
@@ -87,7 +90,7 @@ SnapshotReporter <- R6::R6Class("SnapshotReporter",
         self$cur_snaps[[self$test]] <- self$old_snaps[[self$test]]
 
         if (self$i > 0) {
-          testthat_warn("Snapshots reset after error/skip")
+          testthat_warn(paste0("Snapshots reset after error/skip: ", test))
```
hadley, line 93: Won't the test name have been printed above?

krlmlr: Probably, let's try and see.

krlmlr:
Do we need to always warn with duplicate tests (`is.null()` vs. `length() == 0`)?

R/snapshot-reporter.R
```diff
@@ -35,6 +35,9 @@ SnapshotReporter <- R6::R6Class("SnapshotReporter",
       self$test <- test
       self$i <- 0L
 
+      if (length(self$cur_snaps[[test]]) > 0) {
+        testthat_warn("Duplicate test, discarding snapshot")
```
hadley, line 39:
```suggestion
        testthat_warn("Duplicate test, discarding previous snapshot")
```

I think this logic seems correct.

hadley:
Thanks!

PR #1279, opened 2020-12-13
Fix deprecation message for expect_output_file()

krlmlr:
Otherwise the deprecation message mentions `expect_known_output()` which is confusing.

Due to NSE it's difficult to forward to `expect_known_output()`.


hadley:
Thanks!

PR #1281, opened 2020-12-13
Require braces in test_that()

### LINKED ISSUE #1280: Error location for snapshot tests with error = TRUE
In current tibble I'm seeing:

```
───────────────────────────────────────────────────────
Failure (snapshot.R:67:5): output test
`{ ... }` threw an unexpected error.
Message: `x` must not have more than one dimension. `length(dim(x))` must be zero or one, not 4.
Class:   tibble_error_enframe_has_dim/tibble_error/rlang_error/error/condition
Backtrace:
 1. tibble::enframe(Titanic)
───────────────────────────────────────────────────────
```

The `snapshot.R:67:5` is not particularly helpful here.

krlmlr:
to ensure we get srcrefs.

This adds a few warnings from internal test code, happy to clean up if you think it's a good idea. Should we enable this for testthat 3e only?

Closes #1280.


R/test-that.R
```diff
@@ -39,6 +40,11 @@ test_that <- function(desc, code) {
   local_test_context()
 
   code <- substitute(code)
+
+  if (!is.call(code) || !identical(code[[1]], sym("{"))) {
```
hadley, line 44: You could simplify this to `is_call(code, "{")`

R/test-that.R
```diff
@@ -12,7 +12,8 @@
 #'
 #' @param desc test name.  Names should be kept as brief as possible, as they
 #'   are often used as line prefixes.
-#' @param code test code containing expectations
+#' @param code test code containing expectations.  Braces `{}` should be used
+#'   to surround the code even if it is only a single expression.
```
hadley, line 16:
```suggestion
#' @param code test code containing expectations.  Braces (`{}`) should always be 
#'   used in order to get accurate location data for test failures.
```

hadley (commented):
Yeah, probably best to make this part of 3e, just to avoid adding potentially large numbers of warnings to existing packages.

krlmlr:
Tests pass now.

hadley:
Thanks!

tests/testthat/test-reporter-debug.R
```diff
@@ -35,7 +35,7 @@ get_frame_from_debug_reporter <- function(choice, fun, envir = parent.frame()) {
     sink_number = function() 0L,
     with_reporter(
       "debug",
-      test_that("debug_reporter_test", fun())
+      test_that("debug_reporter_test", { fun() })
```
MichaelChirico, line 38: an issue with this warning is that `styler` explodes this:

```
styler::style_text('test_that("debug_reporter_test", { fun() })')
# test_that("debug_reporter_test", {
#   fun()
# })
```

It's even worse (IMO) for cases where the test description is long:

```
styler::style_text('
test_that(
  "A really long descriptive name for what this simple test is testing for",
  { expect_true(TRUE) }
)
')
# test_that(
#   "A really long descriptive name for what this simple test is testing for",
#   {
#     expect_true(TRUE)
#   }
# )
```

(this is because of the strict rules around braces in the style guide: https://style.tidyverse.org/syntax.html#indenting)

This is especially cumbersome in cases where we're writing a wrapper to `test_that` like

```
setup_teardown_test_that(x, desc, code) {
  setup(x)
  on.exit(teardown(x))

  test_that(desc, code)
}
```

is requiring `tryCatch()` for the `testthat_braces_warning` (or `suppressWarnings()`) really required here?

We are working on rule that blocks one-expression `{` usage because of all the extra whitespace it causes, it looks like we'll need a carve-out for `test_that()` 😐

hadley: Can you file an issue and we'll figure out how to weaken this?

PR #1283, opened 2020-12-13
Tweak CheckReporter output

hadley:
* Display test problems as they appear
* Show details of errors and failures at end
* No longer uppercase error/failure
* Pad expectation types to align nicely


gaborcsardi:
OTOH if there are only successes until a crash/freeze, then you still have no idea what happened. So in packages in which this failure mode is more common, ideally we'd use a slightly different reporter. 

So I don't think any more that we need to print the problems as they happen.

hadley:
We still need to show warnings/skips somewhere though.

gaborcsardi:
Sure, but then we might as well show them at the end? Maybe even in a different block, so that you can focus on the real problems?

hadley:
We could do skips, then warnings, then failures. That would put the most important stuff at the bottom, which is what you'll normally see first when looking at R CMD check results.  OTOH, there is still the potential for a lot of warnings.

gaborcsardi:
Maybe the three current uses (CRAN, CI, revdep checks) of the check reporter are slightly different? E.g. for revdep checks we could omit the skips and warnings completely, no?

hadley:
@gaborcsardi hmmmm, that's an interesting idea too. Does revdepcheck set an env var?

gaborcsardi:
> @gaborcsardi hmmmm, that's an interesting idea too. Does revdepcheck set an env var?

AFAICT, no. But we could easily set it for the local check, and would probably need to modify the script or `Dockerfile` of the cloud checks.

Issue #791, opened 2018-08-16
Complex skip condition may fail vapply in reporter


HughParsonage:
When `skip_if_not` is given a sufficiently complex condition (that is `FALSE`) the standard reporter gives an error at.

https://github.com/r-lib/testthat/blob/8a0371a2868812a63e6f382baa5957fb0c030848/R/reporter-progress.R#L151

**`test_reprex.R`**

```r
context("reprex")

test_that("reprex ", {
  AND <- `&&`
  OR <- `||`
  
  skip_if_not(OR(FALSE,
                 AND(identical(Sys.getenv("TRAVIS"), "dsf"),
                     Sys.getenv("TRAVIS_PULL_REQUEST") != "true")))
  expect_true(TRUE)
})
```

```
> devtools::test(filter = "reprex")
√ | OK F W S | Context
√ |  0     1 | reprex
--------------------------------------------------------------------------------
Error in vapply(issues, issue_summary, FUN.VALUE = character(1)) : 
  values must be length 1,
 but FUN(X[[1]]) result is length 2

== Results =====================================================================
OK:       0
Failed:   0
Warnings: 0
Skipped:  1

Woot!
```

PR #1291, opened 2021-01-03
Skip if deparse1

### LINKED ISSUE #1290: skip_if message should use deparse1
```
a_realllllllly_long_function_name = identity
a_realllllllly_long_variable_name = 1
a_realllllllly_long_threshold_name = 0
skip_if(a_realllllllly_long_function_name(a_realllllllly_long_variable_name) >= a_realllllllly_long_threshold_name)
# Error: Reason: a_realllllllly_long_function_name(a_realllllllly_long_variable_name) >=   is TRUE
#     a_realllllllly_long_threshold_name  is TRUE
```

MichaelChirico:
Closes #1290 

The earlier test for #791 is reused, but that test missed the repeated inputs. I'm not sure if the fix in #791 is now redundant; I assume not.


hadley:
Would you mind adding a backport of `deparse1()` to utils.R? I think that'll make the code a little clearer, making it less likely for this to crop up again.

hadley:
Just something like `deparse1 <- function(x) paste(deparse(x), collapse = " ")`, so as not to have to worry about any potential copyright of the base implementation.

MichaelChirico:
Sure; done.

BTW it would be great to know definitively the copyright rules about backports & non-GPL licenses. I keep hearing it's OK & not OK. Not sure if maybe that's something that's come up with the recent tidyverse license changes.

hadley:
I don't think it's possible to make a blanket declaration — for this specific case, my suspicion would be that `base::deparse1()` wouldn't be considerably copyrightable because insufficiently "creative"; i.e. the specification of the job of deparse1 is sufficiently straightforward that any other would come up with a very similar implementation.

Or do you mean the backports package? I think that's ok because the code isn't bundled with your package, but is added at load time, and the conditions of (e.g.) the GPL only apply to distribution of code. So backports itself must be GPL licensed, but packages that use it don't need to be.

R/utils.R
```diff
@@ -51,6 +51,9 @@ strrep <- function(x, times) {
   }, list(x = x, times = times), MoreArgs = list()), use.names = FALSE)
 }
 
+# Backport for R < 4.0
+deparse1 <- function(expr, ...) paste(deparse(expr, ...), collapse = "\n")
```
hadley, line 55: I converted this to a newline because I think it looks a little better (see snapshot test below)

MichaelChirico:
> wouldn't be considerably copyrightable because insufficiently "creative"

That makes sense. Especially for `deparse1` this keeps coming up for me -- you would have to write awkward code _not_ to come up with the same implementation...

Do you think that applies to any function which is a short wrapper of existing functions (e.g. `trimws` is also like this)? Or perhaps even any function that only uses R code (i.e. no new C code)?

I did check out `backports` but as you said it's also GPL, so don't think the compatibility question is likely to raise any eyebrows there.

R/utils.R
MichaelChirico: One reason I might prefer `" "` is IINM the `deparse` rules for wrapping are not strictly tied to the terminal `width`; using `" "` means "let the terminal decide when to wrap".

Not particularly important in any case, really this fix is about not building the string incorrectly.

hadley:
I think it applies to any short wrapper, regardless of language — it's more the degree of complexity.

R/utils.R
hadley: Interesting — I think we generally assume that the terminal always wraps badly so we take control as much as possible. But yeah, not really important here since this is just about avoiding an errorful edge case.

MichaelChirico: I think that's also not wrong, it's a matter of whether `deparse` or terminal has better wrapping rules... more control would come from `strwrap` but that's usually too much effort IME. I also don't remember having good experience trying to get `width.cutoff` to work naturally, but I could be wrong.

hadley: Plus in this case the first ("reason") and last ("is not TRUE") lines are longer than the others, which you'd also like to take into account somehow, but ideally you want to use code wrapping rather than prose wrapping, so doing it 100% correctly seems like a very hard problem.

MichaelChirico: :100:

PR #1292, opened 2021-01-06
Remove old testthat-problems.rds

krlmlr:
once tests are successful.

Are old tests results of any use?


hadley:
Sorry, missed the connection to this from #1314, but I basically did the same thing 😄

PR #1293, opened 2021-01-06
Always use version 2 when saving RDS files

jimhester:
This preserves compatibility with older R versions

Related to https://github.com/r-lib/pkgbuild/issues/114


hadley:
Ooops, thanks.

PR #1295, opened 2021-01-11
Use dev crayon and cli

jennybc:
Related to issue #1296

### DIFF
--- a/tests/testthat/_snaps/snapshot-serialize.md
+++ b/tests/testthat/_snaps/snapshot-serialize.md
@@ -1,0 +1,11 @@
+# snapshot output doesn't use ANSI escapes
+
+    Code
+      cli::cli_alert_success("Success!")
+    Message <cliMessage>
+      v Success!
+
+---
+
+    
+
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,7 @@ Depends:
 Imports:
     brio,
     callr (>= 3.5.1),
-    cli (>= 2.2.0),
+    cli (>= 2.2.0.9000),
     crayon (>= 1.3.4.9000),
     desc,
     digest,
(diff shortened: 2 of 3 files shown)


jennybc:
I'm not clear whether it was ever intentional to save snapshots with ANSI escapes (linked below and, in fact, I'm having the *opposite* problem in reprex):

* [tests/testthat/_snaps/expect-constant.md](https://github.com/r-lib/testthat/blob/55a14ab7cb580749d1bc6b537203fc7dfab91124/tests/testthat/_snaps/expect-constant.md)
* [tests/testthat/_snaps/expect-equality.md](https://github.com/r-lib/testthat/blob/55a14ab7cb580749d1bc6b537203fc7dfab91124/tests/testthat/_snaps/expect-equality.md)

gaborcsardi:
FWIW dev cli is cleaner and it does not depend on crayon any more. So if we are switching things around already, we might just start using that.

tests/testthat/_snaps/snapshot-serialize.md
```diff
@@ -0,0 +1,11 @@
+# snapshot output doesn't use ANSI escapes
+
+    Code
+      cli::cli_alert_success("Success!")
+    Message <cliMessage>
+      v Success!
+
+---
+
```
jennybc, line 9: Why does `expect_snapshot_output()` produce nothing here?

hadley: It's a message not output? Perhaps `expect_snapshot_output()` is confusingly named?

hadley:
My goal was to turn ANSI escapes should be off by default, but allow you to turn them on if you want them with `local_reproducible_output()`

tests/testthat/_snaps/snapshot-serialize.md
jennybc: I guess I had different expectations based on this in the docs:

> `expect_snapshot_output()` captures just output printed to the console

hadley:
I think this is now resolved via other PRs?

jennybc:
Yeah this was part of the process to figure out what was going on cli and snapshot tests. Fine to close and delete the branch.

PR #1297, opened 2021-01-12
Docs and tests for local_* reproducibility functions

jennybc:
Done while trying to understand some puzzles with saved snapshots


tests/testthat/test-local.R
```diff
@@ -1,10 +1,28 @@
-test_that("expected options are set by default", {
+test_that("local context is 'as promised' inside test_that()", {
+  # high-level expectations or expectations that preceded testthat 3e
   expect_true(is_testing())
   expect_equal(testing_package(), "testthat")
```
jennybc, line 4: This is *not* actually true if you run the test via `test_file()`:

```
> test_file("tests/testthat/test-local.R")

══ Testing test-local.R ═════════════════════════════════════════════════════════
[ FAIL 1 | WARN 0 | SKIP 0 | PASS 17 ]

── Failure (test-local.R:4:3): local context is 'as promised' inside test_that() ───────────────────────────
testing_package() (`actual`) not equal to "testthat" (`expected`).

`actual`:   ""        
`expected`: "testthat"

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 17 ]
```

Should this be fixed?

tests/testthat/test-local.R
```diff
@@ -1,10 +1,28 @@
...
   expect_false(is_interactive())
   expect_equal(Sys.getenv("R_TESTS"), "")
+
+  # set in local_test_context()
+  expect_equal(Sys.getenv("TESTTHAT"), "true")
+
+  # testthat 3e, set in local_reproducible_output()
```
jennybc, line 11: These are documented as being set by `local_test_context()`, but are actually set by `local_reproducible_output()`, which seems a bit confusing.

  * `cli.unicode`
  * `crayon.enabled`
  * `width`

 But `local_test_context()` calls `local_reproducible_output()` if using 3e, which points out that maybe I don't fully understand the ways in which all of these settings are actually changed inside `test_that()`.

hadley:
Thanks! Not sure why the code was so different to the docs.

PR #1316, opened 2021-02-02
Avoid partial match warning

jimhester:
This line generates a partial match warning because `n_skip` is defined as `n_skip_` at L386. I am assuming using the underscore was intentional there, if not an alternative solution would be to remove the underscores from the line I changed in this PR.


hadley:
Thanks!

PR #1323, opened 2021-02-07
Collapse outputs of same type

### LINKED ISSUE #1311: Snapshots should collapse duplicated output types
e.g. multiple messages and code lines

hadley:
Fixes #1311

The primary downside is that this will force many snapshot users to update testthat quickly because their tests will start failing on CI. I think it's still worth doing it, but it requires a little thought.

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -78,11 +78,11 @@ snapshot_replay <- function(x, state) {
 }
 #' @export
 snapshot_replay.character <- function(x, state) {
-  c("Output", indent(split_lines(x)))
+  c(snap_header(state, "Output"), indent(split_lines(x)))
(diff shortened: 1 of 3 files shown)


lionel-:
I like the idea, that will simplify many snapshots. I think it would be helpful to give a hint about the number of collapsed items in the heading:

    Code
      x <- 1
      y <- 1
      {
        message("a")
        message("b")
      }
    Message <simpleMessage> (2x)
      a
      b
    Code
      {
        warn("a")
        warn("b")
      }
    Warning <warning> (2x)
      a
      b

hadley:
That would require a lot more work, and I think it's only an incremental improvement 😄

Issue #1286, opened 2020-12-16
Helpers are scoped correctly with R CMD check but not `devtools::test()`


lionel-:
This concerns both CRAN and dev testthat. Add these files to a package:

```diff
2 files changed, 5 insertions(+)
tests/testthat/helper-test.R | 1 +
tests/testthat/test-test.R   | 4 ++++

new file   tests/testthat/helper-test.R
@@ -0,0 +1 @@
+list <- list("my list")
new file   tests/testthat/test-test.R
@@ -0,0 +1,4 @@
+
+test_that("helper is in scope", {
+  expect_equal(list, list("my list"))
+})
```

This fails with `devtools::test()`:

```
✔ |  OK F W S | Context
✖ |   0 1     | test [0.1 s]
────────────────────────────────────────────────────────────────────────────────
Failure (test-test.R:3:3): helper is in scope
`list` (`actual`) not equal to list("my list") (`expected`).

`actual` is a primitive function
`expected` is a list
```

But works fine with R CMD check.

This worked with `test()` in testthat pre-3.0.

hadley:
testthat already has test for exactly this situation, so I think I'll need more of a reprex.

hadley:
Are you using parallel tests? (In which case this would be #1298)

lionel-:
I just checked and I can still reproduce with CRAN and latest testthat. I am not using parallel tests. What can I do to make the reprex above more complete?

hadley:
Are you using the 3rd edition?

lionel-:
Yes, I believe this is a 3e-only bug.

hadley:
Ok, I see it on a new package. So why doesn't it fail with testthat?

hadley:
If I add `message("Hello")` the file is clearly run, so it must be run in the wrong environment. Maybe it's run in the testthat environment, which is why it works for testthat itself.

lionel-:
oops scratch that. I wrote this based on

> This worked with test() in testthat pre-3.0.

which does not mean 3e. I just tested with 3e disabled and can reproduce the issue. So it's a 3.0 issue that affects both editions.

hadley:
Problem probably lies somewhere between `test_files_setup_env()` and `test_env()`

hadley:
Ooops, no, helpers are run in `test_files_setup_state()`

hadley:
Hmmm it works if I run `test_local()`

hadley:
[...]

PR #1325, opened 2021-02-07
Don't run a single file in parallel

hadley:


gaborcsardi:
I am not sure about this. 

It is true that performance-wise it does not make much sense to run a single test file in a subprocess. (OTOH it is not much slower, either, usually.) 

I chose to run even a single file "in parallel" to make tests more reproducible. Especially when working on a single test file, it is easy to get into a situation where this test file works fine in the main process, but it does not work in the subprocess.

hadley:
One of the reasons I like running individual test files is that you can put a `browser()` in where needed; that obviously doesn't work any more.

gaborcsardi:
Yeah, that makes sense. When you do that, do you use `test_file()` or `test(filter = ...)`? Maybe it would make sense not to run `test_file()` in parallel, only `test(filter = ...)`, regardless of how many files the latter one selects?

hadley:
Well I call `devtools::test_file()` which calls `testthat::test()` with `filter` set 😬

gaborcsardi:
Oh, right. Well, then there are two ways to go I guess.

1. We do not merge this PR, and if you want interactive debugging, then you do `Sys.setenv(TESTTHAT_PARALLEL = "FALSE")`. (Or use the Debug reporter.)
2. We do merge this PR, and then if people want to run a single test file in a sub-process, they'll do `Sys.setenv(TESTTHAT_PARALLEL = "TRUE")`.

hadley:
Looks like I need to patch devtools for #1286, so I can just make it call `test_file()` instead.

hadley:
`test_file()` already relies on the default value of `parallel` (FALSE) so there's nothing more to do here.

PR #1329, opened 2021-02-11
fix test_files call in auto_test_package (fix #1320)

### LINKED ISSUE #1320: `auto_test_package()` package is not a string
## Overview
After the first run of `auto_test_package()` (i.e., when it had actually started to acting automatically), when a test changes, an error is returned:
```
Error: package is not a string
```

## Step to reproduce the problem

```
usethis::create_package("foo")
usethis::use_test("foo")
testthat::auto_test_package()  # everything works up to here
# make and save any change to test-foo.R
# "Error: package is not a string"
```

## TL;DR (proposed solution)

The `test_files` call inside the `watcher` function defined into `auto_test_package` ([auto-test.R#L43](https://github.com/r-lib/testthat/blob/master/R/auto-test.R#L43)) could be wrong:

it is `test_files(tests, env, reporter = reporter$clone(deep = TRUE))` while I think it should be:

```
test_files(
        test_dir = test_path,
        test_package = package,
        test_paths = tests,
        env = env,
        reporter = reporter$clone(deep = TRUE)
)
```

Let me know if you agree with the solution and like a pull-request.

## Details
Unfortunately, stopping the automation and calling `traceback()` is not that much informative:

```
3. Sys.sleep(1) at watcher.R#23
2. watch(c(code_path, test_path), watcher, hash = hash) at auto-test.R#132
1. auto_test_package() 
```

After some investigations, I (think I have) found the source of the problem:

`test_files` (in this case) redefine itself as `test_files_serial` ([test-files.R#L168](https://github.com/r-lib/testthat/blob/master/R/test-files.R#L168), so some attention could be needed for the parallel version), passing to it the option `test_package = test_package`, which is actually our "`env`" ([auto-test.R#L58](https://github.com/r-lib/testthat/blob/master/R/auto-test.R#L58)).

From that point, the following chain of functions are called (with the reported arguments):

1. `test_files_setup_state(test_dir, test_package, load_helpers, env)`, [test-files.R#L197](https://github.com/r-lib/testthat/blob/master/R/test-files.R#L197)
2. `local_test_directory(test_dir, test_package, .env = .env)`, [test-files.R#L229](https://github.com/r-lib/testthat/blob/master/R/test-files.R#L229) (note also that here the second argument's name is `package`)
3. `local_edition(find_edition(path, package), .env = .env)`, [local.R#L144](https://github.com/r-lib/testthat/blob/956fce8672c7b58b6d832e589b5af6b5862cfbf1/R/local.R#L144), and [edition.R#1](https://github.com/r-lib/testthat/blob/f1dd05cbae2432c1ad8a3db53849ce01819f4172/R/edition.R#L1)
4. `find_description(path, package)`, [edition.R#L10](https://github.com/r-lib/testthat/blob/f1dd05cbae2432c1ad8a3db53849ce01819f4172/R/edition.R#L10)
5. `desc::desc(package = package)`, [edition.R#L12](https://github.com/r-lib/testthat/blob/f1dd05cbae2432c1ad8a3db53849ce01819f4172/R/edition.R#L12)

and `desc::desc(package = package)` returned "Error: package is not a string" because `package` is actually our `env`.

CorradoLanera:


CorradoLanera:
The error for ubuntu-16.04 3.3 and older is the same (during the "Install system dependencies" stage):

```
 x System command 'sh' failed, exit status: 1, stdout + stderr:
E> gpg: keyring `/tmp/tmpc8i_e1j4/secring.gpg' created
E> gpg: keyring `/tmp/tmpc8i_e1j4/pubring.gpg' created
E> gpg: requesting key 15B55A9F from hkp server keyserver.ubuntu.com
E> Error: retrieving gpg key timed out.
E> gpg: keyserver timed out
E> gpg: keyserver receive failed: keyserver error 
```

R/auto-test.R
```diff
@@ -111,7 +111,12 @@ auto_test_package <- function(pkg = ".", reporter = default_reporter(), hash = T
       # If test changes, rerun just that test
       cat("Rerunning tests: ", paste0(basename(tests), collapse = ", "), "\n")
       env <- env_clone(asNamespace(package))
-      test_files(tests, env, reporter = reporter$clone(deep = TRUE))
+      test_files(
+        test_dir = test_path,
+        test_package = package,
+        test_paths = tests,
+        env = env, reporter = reporter$clone(deep = TRUE)
```
hadley, line 118:
```suggestion
        env = env, 
        reporter = reporter$clone(deep = TRUE)
```

hadley:
Thanks!

PR #1330, opened 2021-02-11
Capture tracebacks for wrapped expectations

### LINKED ISSUE #1307: Consider backtrace for wrapped expectations
e.g. pins includes a function that package authors can use to test that their pins class follows the specification. But because (IIRC) the expectations are not in the `tests/` directory, it's hard to figure out what test has failed. In this case it would be nice for failed expectations to include a back trace so that you can track back to exactly what went wrong.

hadley:
Fixes #1307

@lionel- could you please double check my reasoning here? The basic idea is to capture the "topenv" in `test_code()` and `source_file()` (as a fallback for expectations not nested inside `test_that()` blocks) and use that as the `top` argument to `trace_back()`. I then set `bottom` so that the `expect_*` call is at the bottom of the stack (this mostly happens in `expect()`, but I also handled all internal wrappers to avoid exposing implementation details). I then only show the back trace if the expectation fails, and there's more than one call in the backtrace (since the header also gives a location).

### DIFF
--- a/R/expect-that.R
+++ b/R/expect-that.R
@@ -50,7 +50,7 @@ expect_that <- function(object, condition, info = NULL, label = NULL) {
 #' test_that("this test fails", fail())
 #' test_that("this test succeeds", succeed())
 #' }
-fail <- function(message = "Failure has been forced", info = NULL, trace_env = caller_env(1)) {
+fail <- function(message = "Failure has been forced", info = NULL, trace_env = caller_env()) {
   expect(FALSE, message, info = info, trace_env = trace_env)
 }
 
--- a/R/expect-that.R
+++ b/R/expect-that.R
@@ -50,8 +50,8 @@ expect_that <- function(object, condition, info = NULL, label = NULL) {
 #' test_that("this test fails", fail())
 #' test_that("this test succeeds", succeed())
 #' }
-fail <- function(message = "Failure has been forced", info = NULL, trace = NULL) {
-  expect(FALSE, message, info = info, trace = trace %||% test_trace(2))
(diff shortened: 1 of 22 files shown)


R/expect-that.R
```diff
@@ -50,11 +50,10 @@ expect_that <- function(object, condition, info = NULL, label = NULL) {
 #' test_that("this test fails", fail())
 #' test_that("this test succeeds", succeed())
 #' }
-fail <- function(message = "Failure has been forced", info = NULL) {
-  expect(FALSE, message, info = info)
+fail <- function(message = "Failure has been forced", info = NULL, trace = NULL) {
```
lionel-, line 53: Are you planning to make this a common pattern? We probably want to avoid creating traces ahead of time and discarding them if the expectation succeeds because that will bring performance down.

lionel- (commented):
I like the definition of the top env but for the bottom env I think it's hard and often brittle to reason in terms of frame numbers. Can you pass a concrete environment as `bottom` argument instead of a number of frames to trim? This would be especially easier to use for implementors of expectations when they create a trace manually. Probably the `expect_` functions can simply pass down their own environment?

R/expect-that.R
hadley: I think the hint about passing around bottom envs resolves this — I can just pass around the bottom env, and then only create the traceback when needed. Not sure how I forgot about that when I was using it for the top env 😆

lionel- (approved):
Looks good!

Regarding style I'd favour `caller_env()` instead of `caller_env(1)` just to avoid encouraging patterns that require a frame number to be supplied. Not a big deal though.

hadley:
Ooops, thanks

PR #1377, opened 2021-05-06
Fix regexp pattern in grepl (#1375)

### LINKED ISSUE #1375: Wrong specification of regexp in `snap_from_md`
In this line: https://github.com/r-lib/testthat/blob/master/R/snapshot-serialize.R#L19, the regexp is specified as `grepl("^-{3, }", ...)`. Note that there is a space in front of } which is, according to the regex documentation, wrong. For an unknown reason, R accepts this specification, but throws an error in this case `grepl("^-{3, 5}, ...)`. If you try to use the pattern `^-{3, }`, it should match this string "-{3, }". R wrongly ignores the space there.

The correct pattern should be "^-{3,}".

Akirathan:
Fixes #1375 

This fixes the regexp pattern that is wrongly interpreted by GNU-R. Now the pattern is compatible with PCRE.

### DIFF
--- a/R/snapshot-serialize.R
+++ b/R/snapshot-serialize.R
@@ -16,7 +16,7 @@ snap_from_md <- function(lines) {
   names(tests) <- gsub("^# ", "", lines[h2])
 
   split_tests <- function(lines) {
-    sep <- grepl("^-{3, }", lines)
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1381, opened 2021-05-18
add expect_no_match()

### LINKED ISSUE #1380: Add expect_no_match
This FR is to have a complete function to expect_match that handles the negative case. This is a pattern that exists in other frameworks. Here's an example from python:
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertNotRegex

I have a working version of this. Would you be interested in a patch?

michaelquinn32:
Fixes #1380


michaelquinn32:
@MichaelChirico Let me know if this lines up with what you were thinking re `grep()` and `grepl()`

MichaelChirico:
Yep that's what I had in mind, thanks!

hadley:
I’d prefer to not use `grep()` and keep the change minimal so I don’t need to review the source code of grepl/grep to ensure there aren’t any unhappy surprises. .

MichaelChirico:
FWIW I looked into why grepl doesn't have an invert argument recently and know their sources are very tightly coupled...

but fair point -- I will try and put the backend to grep() in a different PR if so desired

michaelquinn32:
SGTM. Thanks for the feedback Hadley!

hadley (approved):
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

michaelquinn32:
Friendly ping!

Is there anything else that I can do here?

PR #1386, opened 2021-05-27
support setting edition with an environment variable

### LINKED ISSUE #1384: Support setting the value of testthat edition via an environment variable
Hi team!

I hope this finds you well, and thanks for all of the great work on `testthat`.

We use a build system to support R at Google, which unfortunately makes it somewhat difficult to access a package's description file when running tests. Instead, we've modified `testthat` internally to be able to set the edition using an environment variable.

Is this something worth supporting for other users?

Thanks!

michaelquinn32:
Fixes #1384


R/edition.R
```diff
@@ -58,14 +62,16 @@ edition_name <- function(x) {
...
 local_edition <- function(x, .env = parent.frame()) {
-  stopifnot(is.numeric(x) && length(x) == 1)
+  stopifnot(is_zap(x) || (is.numeric(x) && length(x) == 1))
   old <- edition_set(x)
   withr::defer(edition_set(old), envir = .env)
 }
 
 edition_set <- function(x) {
   if (is_zap(x)) {
+    old <- env_get(testthat_env, "edition", default = 2L)
```
hadley, line 72: Are sure it makes sense to set a default here?

michaelquinn32: TBH, didn't think that much about it.

In my tests, if no default is needed, we can simplify the whole thing to

```
edition_set <- function(x) {
  env_poke(testthat_env, "edition", x)
}
```

This still works with `zap()` and keeps a consistent return type. How does that sound?

hadley: Yeah, that makes sense. I probably wrote this for an earlier version of `env_poke()`

michaelquinn32: Thanks!

PR #1396, opened 2021-06-14
Reset snapshots on error

lionel-:
The snapshot reporter normally preserves snapshots when a skip or unexpected error occurs. However, `expect_snapshot()` currently signals a failure rather than an error, causing snapshots to be deleted. This PR fixes that by signalling an error instead of a failure.

To test `expect_snapshot()` itself (instead of testing a local snapshotter via its methods) I have added a new directory of test files.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `expect_snapshot()` no longer deletes snapshots when an unexpected
+  error occurs.
+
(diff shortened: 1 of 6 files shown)


hadley (approved):
LGTM. Do you want to do the release after this?

lionel-:
> Do you want to do the release after this?

Yup sounds good.

Issue #1393, opened 2021-06-09
auto-deletion of snapshots on devtools::test() has false positives on test failures


mjskay:
(not sure if this is a testthat or vdiffr issue, but I am 95% sure it is testthat, so posting here but tagging @lionel-)

I encountered a snag in my workflow while trying out the new workflow for vdiffr that is based on snapshots, and I am wondering if I've missed a more sensible workflow or if devtools::test() is being a bit too trigger-happy on deleting unused snapshots.

Basically, while fixing some bugs in {ggdist} (which has a large number of vdiffr test cases), I've discovered that if there is a bug in a test case that causes a snapshot to fail (e.g. a bug that causes `vdiffr::expect_doppelganger()` to error), if you run `devtools::test()` it will think that the corresponding snapshot is no longer used and delete it (I provide an example below). I think this is somewhat problematic because once you fix the bug causing the error, the snapshot to test against is no longer there! 

You can kind of work around this by filtering to specific files (which seems not to delete unused test cases), but this is still cumbersome because: (1) often (if you're like me) you'll discover the bug by running `devtools::test()`, so the test case is already deleted and (2) I'd much prefer to be able to have fast enough test cases that I can reflexively run the full suite while fixing stuff, especially if there are lots of test cases or problems in multiple parts of the package introduced by a change in an upstream dep.

## An example

Create a package, run `usethis::use_testthat()`, then add this test file:

```r
library(ggplot2)

test_that("simple plot", {
  vdiffr::expect_doppelganger("simple plot", ggplot(data.frame(), aes(1,1)) + geom_point())
})
```

Running `devtools::test()` the first time creates the snapshot:

```
i Loading testpkg
i Testing testpkg
√ |  OK F W S | Context
√ |   1   1   | test [0.1 s]                                                                              
----------------------------------------------------------------------------------------------------------
Warning (test-test.R:4:3): simple plot
Adding new file snapshot: '_snaps/test/simple-plot.svg'
----------------------------------------------------------------------------------------------------------
```

Second run of `devtools::test()` passes:

```
i Loading testpkg
i Testing testpkg
√ |  OK F W S | Context
√ |   1       | test                                                                                      

== Results ===============================================================================================
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1 ]
```

Now imagine the above test started erroring. Simulate that by changing the test case to use `stop()`:

```r
library(ggplot2)

test_that("simple plot", {
[...]

PR #1397, opened 2021-06-14
Add `announce_snapshot_file()` to prevent unwanted auto-deletion

lionel-:
Part of #1393.

Used in r-lib/vdiffr#109 to prevent auto-deletion of snapshot files in case of unexpected errors and skips.


R/snapshot-file.R
```diff
@@ -21,6 +21,17 @@
...
+#'
+#' @section Announcing snapshots:
+#' testthat automatically detects dangling snapshots that have been
+#' written to the `_snaps` directory but which no longer have
+#' corresponding R code to generate them. These dangling files are
+#' automatically deleted so they don't clutter the snapshot
+#' directory. However we want to preserve snapshot files when the R
+#' code wasn't executed because of an unexpected error or because of a
+#' [skip()]. Let testthat know about these files by calling
+#' `announce_snapshot_file()` before `expect_snapshot_file()`.
```
hadley, line 33: I think you should also add `expect_snapshot_png()` in the example below to show what a full wrapper will look like. Otherwise LGTM.

PR #1401, opened 2021-06-16
Restore breaking change for condition expectations

lionel-:
To merge after the 3.0.3 release.

The NEWS item now goes over more details about the breaking change, its motivation, and how to fix it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,35 @@
 # testthat (development version)
 
 * Condition expectations now consistently return the expected
-  condition instead of the return value (#1371).
+  condition instead of the return value (#1371). Previously, they
+  would only return the condition if the return value was `NULL`,
+  leading to inconsistent behaviour.
+  
+  This is a breaking change (it only affects edition 3). Where you
+  could previously do:
+  
+  ```
+  expect_equal(expect_warning(f(), "warning"), "value")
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,3 +1,36 @@
...
+  
+  ```
+  expect_warning(expect_equal(f(), "value"), "warning")
+  
+  # Equivalently, save the value before inspection
+  expect_warning(value <- f(), "warning")
+  expect_equal(value, "value")
+  
+  # If you use `=` instead of `<-` be sure to disambiguate
+  # assignement and argument passing with braces
```
DavisVaughan, line 25:
```suggestion
  # assignment and argument passing with braces
```

NEWS.md
```diff
@@ -1,3 +1,36 @@
+# testthat (development)
```
DavisVaughan, line 1:
```suggestion
# testthat (development version)
```

Since usethis is pretty strict about the format of this header when it is being auto updated from dev to release
https://github.com/r-lib/usethis/blob/486ed15e93c0d2421a623249441aa0685a124733/R/news.R#L33-L37

lionel-: This header is only indicative. We'll bump the version the usual way.

PR #1403, opened 2021-06-24
Use a constant value for the stack size

### LINKED ISSUE #1373: Build of testthat fails
When I try to build the current master it fails with 
`
../inst/include/testthat/vendor/catch.h:6495:33: error: size of array ‘altStackMem’ is not an integral constant-expression

 6495 |         static char altStackMem[SIGSTKSZ];

size of array ‘altStackMem’ is not an integral constant-expression

 6546 |     char FatalConditionHandler::altStackMem[SIGSTKSZ] = {};

`
With gcc version: 11.1.1

jimhester:
Fixes #1373

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,6 +58,9 @@
  * Support setting the testthat edition via an environment variable
   (`TESTTHAT_EDITION`) as well (@michaelquinn32, #1386).
  
+* The vendored Catch code used for `use_catch()` now uses an constant value for the stack size rather than relying on SIGSTKSZ.
+  This fixes compatibility for recent glibc versions where SIGSTKSZ is no longer a constant.
+
 # testthat 3.0.2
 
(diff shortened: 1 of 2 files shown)


hadley:
Does this merit a patch release?

jimhester:
I don't know, not many packages use catch and this only affects very recent versions of glibc.

NEWS.md
```diff
@@ -58,6 +58,9 @@
  * Support setting the testthat edition via an environment variable
   (`TESTTHAT_EDITION`) as well (@michaelquinn32, #1386).
  
+* The vendored Catch code used for `use_catch()` now uses an constant value for the stack size rather than relying on SIGSTKSZ.
```
hadley, line 61: I think this is in the wrong place?

hadley (approved):

PR #1410, opened 2021-07-07
Add `browser2()`

lionel-:
To browse through code running under `expect_snapshot()`.

I wonder if we should add the functionality to `browser()` itself though? It doesn't seem ever useful to browse through code without being able to inspect the environment. And in that case, do we still want to export `browser2()`? Perhaps temporarily?


hadley:
Yeah, I think you're right that we should just get it into `browser()` itself.

PR #1411, opened 2021-07-07
Tweak progress reporter colours/order

### LINKED ISSUE #1382: testing packages has too many colors in terminal output
My test iteration process is quite normal I think:

 * do some coding
 * run tests
 * and then depending on how long the testing takes:
   * wait to observe results or...
   * go back to code and look back at test results

My terminal looks like this when I test (Rtest is just an alias to run devtools::test() )

Screenshot from 2021-05-23 13-31-27

Glancing back to this terminal, from my IDE, there's a lot to take in, when all I'm interested in is wheter or not there were test failures. (One does get the feeling that whomever coded the terminal testoutput thought terminal colors were a terrific hammer)

* Colors are a strong visual aid. However the colors in the test output are _non-informative_ to the test result, they are decribing generic sections, not specific results.
* The test specfic bits of the output is white text (the numbers of test outcomes), which is the last thing my brain sees.
In this case:

* Orange FAIL hits my eyesight first, giving the initial interpration that there are test failures here.
* The brain then remembers not to get alarmed by alarming colors, and starts looking for more info.
* It spends some extra energy not getting caught up by lines and structure leading it to where it shouldn't be lead.
* Eyes then doesn't easily see the "0" next ot FAIL, but instead finds the magenta WARN that draws my eyes to itself.
* At this point the 0's next to both FAIL and WARN start to get noticed, and I feel reassured and but search on for the PASS number all at the end of the line, with a reassuringly high number next ot it.
* Also the double lines lead my eyes to see above it for what goes on, when the first thing I want to see is the test fail count. This can perhaps be learned. Though I have been testing packages for many years. If you fix the colors, I am inclined to accept or rewire brain for the double line.

## I propose:

* Use the colors to indidcate _specific result_, not _generic function_:
  * When there are test failures, only hightlight FAIL with red
  * When there are no test failures, only highlight PASS with green.

In a busy everyday life with rapidly approaching deadlines, this would make this one less thing to spend extra energy on.

Yes, I could perhaps tune my terminal colors, but I won't do that for thet sole purpose of accomodating R testoutput (not yet atleast)

Also if I only worked with R and only ever used TDD with R packages in my everyday life, I would get acustomed to the specific quirks of its test output. But I don't, I would guess most poeple don't. There are all kinds of systems, languages and platofrm, each of course will do things their own way,

hadley:
* In final summary, only warn/fail/skip if non-zero and only colour pass if no failures. This draws attention to the most important results.
* Colour per-file fails/warnings/skips to make it easier to see which it is (since the header is often off-screen)
* Match order of per-file results to order of test suite results (i.e. always show failures first and passes last)

Fixes #1382

Example of final summary:
Screenshot 2021-07-08 at 9 03 39 AM
Screenshot 2021-07-08 at 9 03 03 AM

Example of per-file output:
Screenshot 2021-07-08 at 9 04 51 AM

@torbjorn how does this look to you? @jennybc, @lionel-, @gaborcsardi any thoughts/comments on this change?

### DIFF
--- a/R/reporter-check.R
+++ b/R/reporter-check.R
@@ -73,13 +73,17 @@ CheckReporter <- R6::R6Class("CheckReporter",
 )
 
 summary_line <- function(n_fail, n_warn, n_skip, n_pass) {
+  colourise_if <- function(text, colour, cond) {
+    if (cond) colourise(text, colour) else text
+  }
+
   # Ordered from most important to least important
   paste0(
     "[ ",
-    colourise("FAIL", "failure"), " ", n_fail, " | ",
-    colourise("WARN", "warn"),    " ", n_warn, " | ",
-    colourise("SKIP", "skip"),    " ", n_skip, " | ",
-    colourise("PASS", "success"), " ", n_pass,
+    colourise_if("FAIL", "failure", n_fail > 0), " ", n_fail, " | ",
+    colourise_if("WARN", "warn", n_warn > 0),    " ", n_warn, " | ",
+    colourise_if("SKIP", "skip", n_skip > 0),    " ", n_skip, " | ",
+    colourise_if("PASS", "success", n_fail == 0), " ", n_pass,
     " ]"
   )
 }
--- a/R/reporter-progress.R
+++ b/R/reporter-progress.R
@@ -159,7 +159,7 @@ ProgressReporter <- R6::R6Class("ProgressReporter",
       if (complete && time > self$min_time) {
         message <- paste0(
           message,
-          cli::col_cyan(sprintf(" [%.1f s]", time))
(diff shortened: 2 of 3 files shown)


gaborcsardi:
I think it is definitely an improvement. Personally I would also make the running times grey-ish, so they do not stand out, either.

jennybc:
LGTM / I agree with 👆

Slightly off-topic: Does anyone else think the space between running time and the "s" is a bit weird?

hadley:
Yeah, that looks better I think:

Screenshot 2021-07-08 at 10 20 35 AM

lionel-:
Looks good!

PR #1412, opened 2021-07-07
Eliminate restart in favour of a condition

hadley:
@lionel- can you see any reason not to make this change? It feels better to stick to a well known tool, and I don't see any big advantage to using a restart here.


lionel- (approved):
LGTM. I agree that in simple cases it is better to use conditions.

PR #1414, opened 2021-07-08
Optionally scrub text from snapshot_output

### LINKED ISSUE #1345: Add replace argument to expect_snapshot()
For when you have sensitive or random components:

```R
  expect_snapshot(
    gargle_oauth_sitrep(tmp_cache)
    replace = c("{path}" = tmp_cache)
  )
```

hadley:
Fixes #1345

Any thoughts on what the argument name should be?


jennybc:
I'm evaluating this relative to what I'm doing in googledrive and have some questions:

What if I need multiple scrubbers (which I do, a lot, in googledrive)?
What I I need to pass another argument to the scrubber?

This snippet is a really typical one in googledrive atm:

``` r
  drive_download_message <- drive_download_message %>%
    scrub_filepath(download_filepath) %>%
    scrub_filepath(file_to_download) %>%
    scrub_file_id()
```

Would I have to create a custom function potentially inside each snapshot-using test?

hadley:
Yeah, the expectation is that you'd wrap up your own functions for this.  I think it would be relatively straightforward to rework what you currently have. Something like:

```R
gargle_scrub <- function(paths = character()) {
  function(x) {
    for (path in paths) {
      x <- sub(path, "{path}", x, fixed = TRUE)
    }
    x <- sub("id_pattern", "{id}", x, perl = TRUE)
    x
  }
}
```

To make it a little easier, I can apply `as_function()` so you could use `~` notation if you wanted.

hadley:
Call it `transform`

jennybc:
Do you want me to actually try translating my workaround into this?

hadley:
@jennybc yeah, that'd be great. I think I'll merge this, and then I'll incorporate any feedback you want to pass along.

PR #1416, opened 2021-07-11
A message or warning shouldn't count as a expectation

hadley:
So if a test contains only messages or warnings it should generate a "empty test skip" which will be reported in the skip summary. This is particularly important for snapshot tests which generate a warning the first time they are run.

@lionel- I'm pretty sure this is correct, and flagging messages/warnings as "handled" was just an oversight when we originally wrote this code, but I'd appreciate a second opinion.


lionel- (approved):
Agreed, this makes sense.

PR #1419, opened 2021-07-21
Show calls in errors

### LINKED ISSUE #1418: Show call in error
In testthat:

```R
Error: unused argument (paths[[1]])
Backtrace:
 1. testthat::expect_equal(...) test-utils.R:20:2
 2. testthat:::expect_waldo_equal("equal", act, exp, info, ..., tolerance = tolerance)
 3. testthat:::waldo_compare(...)
 4. waldo::compare(x, y, ..., x_arg = x_arg, y_arg = y_arg)
 5. waldo:::compare_structure(x, y, paths = c(x_arg, y_arg), opts = opts) /Users/hadley/Documents/r-lib/waldo/R/compare.R:149:2
 6. waldo::compare_proxy(x, paths[[1]]) /Users/hadley/Documents/r-lib/waldo/R/compare.R:155:4
```

When called directly:

```R
Error in compare_proxy.spec_tbl_df(x, paths[[1]]) : unused argument (paths[[1]])
```

hadley:
Fixes #1418

### DIFF
--- a/R/expectation.R
+++ b/R/expectation.R
@@ -158,11 +158,10 @@ format.expectation_success <- function(x, ...) {
 
 #' @export
 format.expectation <- function(x, simplify = "branch", ...) {
-  message <- exp_message(x)
   # Access error fields with `[[` rather than `$` because the
   # `$.Throwable` from the rJava package throws with unknown fields
   if (is.null(x[["trace"]]) || trace_length(x[["trace"]]) == 0L) {
-    return(message)
+    return(x$message)
   }
 
   max_frames <- if (simplify == "branch") 20 else NULL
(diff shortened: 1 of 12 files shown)


R/expectation.R
```diff
@@ -199,7 +190,13 @@ as.expectation.expectation <- function(x, srcref = NULL) {
 
 #' @export
 as.expectation.error <- function(x, srcref = NULL) {
-  expectation("error", cnd_message(x), srcref, trace = x[["trace"]])
+  if (is.null(x$call)) {
+    msg <- paste0("Error: ", cnd_message(x))
+  } else {
+    msg <- paste0("Error in ", expr_deparse(x$call[1]), ": ", cnd_message(x))
```
lionel-, line 196: `expr_deparse()` might return multiple lines in case of quoted `{` and inlined functions. In rlang complex expressions are skipped. I guess we want to strip arguments? The call should also be backticked.

Maybe it's easier to wait for rlang to export the relevant formatter?

hadley: Shouldn’t the `[1]` be sufficient?

lionel-: Not always because the CAR might be a complex expression. You could use `as_label()` instead.

hadley: Oooh yeah. Probably unlikely to affect us, but better to be safe.

lionel-: Did we want to also backtick the call for consistency with rlang errors?

hadley: Do we consistently backtick code? I thought we only did that for argument names?

lionel-: rlang is following cli formatting:

```r
cli::cli_text("{.code foo()}")
#> `foo()`
```

Personally I like the consistency with markdown syntax.

hadley: Makes sense. I'll fix momentarily.

Issue #1241, opened 2020-11-19
`expect_equal` cannot replicate `expect_equivalent`'s behavior


egonulates:
Since `expect_equivalent` function deprecated, I tried to use `expect_equal(ignore_attr = TRUE)` without any success:

```
library(testthat)

x <- 1
(y <- setNames(1, 'a'))
#> a 
#> 1

# Old function working perfectly
expect_equivalent(x, y)
# New function does not work
expect_equal(x, y, ignore_attr = TRUE)
#> Error: `x` not equal to `y`.
#> names for current but not for target
```

hadley:
`expect_equal()` no longer ignores names because they're so important for describing details. I'd recommend using an explicit `unname()` instead.

PR #1422, opened 2021-08-05
`ignore_attr = TRUE` does not ignore names

ms609:
#1241 caught me out: is it perhaps worth noting here?


hadley:
Thanks for pointing this out. I'll use a slightly different wording in an upcoming PR.

PR #1442, opened 2021-09-13
Improve labelling of infix expressions

### LINKED ISSUE #1432: quasi_label of multi-line expression with operator looks wrong
I was looking at a test failure of the following:

```
  expect_identical(X[key1 == "val1" &
    key2 == "val2"]$key3, 0.08)
```

Fails like:

```
...$NULL not identical to 0.08.
Objects equal but not identical
```

The key is (1) the expression is >1 line, so this branch is triggered:

https://github.com/r-lib/testthat/blob/18b2c6c5cc8c4254da499d0d17ba4694faac1d1d/R/quasi-label.R#L70-L77

And then (2) the outermost call is `$`, so building the call as `$(...)` gives this funny output:

```
as.call(list(quote(`$`), quote(...)))
# ...$NULL
```

The question is, what do we want to be the right behavior in this case?

It looks like maybe

```
...$key3
```

would be an improvement, but I'm not sure if this is sufficiently general or what other edge cases might look worse.

hadley:
Fixes #1432


MichaelChirico:
LGTM, great!

PR #1453, opened 2021-09-16
Tweaks to max failures

hadley:


DavisVaughan (approved):
LGTM and I tried it out locally and it worked as expected

PR #1456, opened 2021-09-22
Add a class to the testthat braces warning

### LINKED ISSUE #1455: Add a condition to the 3e warning on braced code
Hi team!

In https://github.com/google/patrick/issues/7, a user pointed out that `patrick` is throwing a lot of warnings in testthat 3e. I've tracked it down to this condition:
https://github.com/r-lib/testthat/blob/0ce7295b2954d752b445e240e5e218c987cc72b2/R/test-that.R#L46

I'm happy to catch this with a calling handler (the line numbers still appear to be correct in `patrick`), but I would appreciate a more robust way to identify this warning. Right now, I'm checking the string contents of the message. Would it be possible to add a class here, ala `rlang::warn()`?
https://rlang.r-lib.org/reference/abort.html

If you're fine with the change, I'd be happy to submit the PR.

Thanks!

michaelquinn32:
Fixes #1455

### DIFF
--- a/R/test-that.R
+++ b/R/test-that.R
@@ -43,7 +43,10 @@ test_that <- function(desc, code) {
 
   if (edition_get() >= 3) {
     if (!is_call(code, "{")) {
-      warn("The `code` argument to `test_that()` must be a braced expression to get accurate file-line information for failures.")
+      warn(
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1460, opened 2021-09-28
Set `cli.condition_width` in tests

### LINKED ISSUE #1458: Add `cli.condition_width` to `local_reproducible_output()`
From https://github.com/r-lib/cli/pull/356.

Should be set to `Inf` to avoid width-wrapping error messages. Newlines might prevent error matches and cause confusing behaviour (reported by @DavisVaughan).

lionel-:
Closes #1458


tests/testthat/test-expect-condition.R
```diff
@@ -153,6 +153,18 @@ test_that("cnd expectations consistently return condition (#1371)", {
...
 })
 
+test_that("cli width wrapping doesn't affect text matching", {
+  skip_if_not_installed("cli", "3.0.2")
+  skip_if_not_installed("rlang", "1.0.0")
+
+  local_use_cli()
+
+  expect_error(
+    abort("foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz foobarbaz"),
```
hadley, line 163: Is there a reason to not call `cli_abort()` directly?

lionel-: `cli_abort()` will need to be updated after rlang 1.0 is out so I thought I'd leave that out of the equation for now.

hadley: Ok, makes sense

PR #1465, opened 2021-09-30
Overwrite file snapshots

hadley:
So that new failures replace old ones

cc @schloerke


schloerke:
LGTM. I also do not see any other usage of `file.copy` where `overwrite = TRUE` be useful
Thank you!

PR #1469, opened 2021-10-04
Expect shape

### LINKED ISSUE #1423: Proposed expectations: expect_nrow(), expect_ncol(), expect_dim()
`testthat::expect_length()` is convenient for testing the size of vectors.

A natural extension would seem to be similar expectations for `data.frame`s/matrices/arrays, where expectations like

```
expect_nrow(filter(DF), 10L)
expect_ncol(transform(DF), 5L)
expect_dim(transformation(array), c(1L, 5L, 6L))
```

Would come in handy. For list-extensions like `data.frame`, of course `expect_length` works, but an `expect_ncol` would also be more readable for this case.

Does that make sense to include in `testthat` itself? Happy to file the PR if so.

From a quick glance at our particular subset of CRAN, I see 413 usages of `expect_equal(nrow(`, 98 for `expect_equal(ncol(`, and 130 of `expect_equal(dim(`, so it's clearly used a decent amount already by downstreams.

MichaelChirico:
Closes #1423

Implementation as of now has a few design choices, happy to adapt as seen fit:

 * Subsumes `expect_length()`, i.e., `expect_shape(x, n)` matches `expect_length(x, n)`. Makes it easier for a series of tests to _all_ use `expect_shape()`, instead of jumping between `expect_length()` and `expect_shape()` from test to test
 * Allows `expect_shape(., nrow=m, ncol=n)`, which is usually the same as `expect_shape(., c(m, n))`, expect when `length(dim(.)) > 2`, a rather boutique use case of wanting to test only the first 2 dimensions of an array.


MichaelChirico:
SGTM, I've just pulled everything up to latest `main` & moved `expect_shape()` into its own file.

R/expect-shape.R
```diff
@@ -0,0 +1,76 @@
...
+#' @seealso [expect_length()] to specifically make assertions about the
+#'   [length()] of a vector.
+#' @inheritParams expect_that
+#' @param shape Expected shape, a numeric vector.
+#' @param nrow Expected number of rows, numeric.
+#' @param ncol Expected number of columns, numeric.
+#' @family expectations
+#' @export
+#' @examples
+expect_shape = function(object, shape, nrow, ncol) {
```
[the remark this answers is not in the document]
MichaelChirico: One idea is to use `...` to require any arguments used to be named.

At first glance I would find `expect_shape(x, c(NA, 1))` a bit confusing, as a reader/reviewer you'd almost certainly need to consult the help.

Another idea would be to disallow using _both_ `nrow=` and `ncol=` so that only three things are allowed:

 1. `expect_shape(x, shape=...)`
 2. `expect_shape(x, nrow=...)`
 3. `expect_shape(x, ncol=...)`

Still another (tangential) idea is to drop the `shape=` argument in favor of things more conceptually precise in R:

```r
expect_shape = function(x, ..., dim, length, nrow, ncol)
```

Any invocation only uses one argument, but across the suite we always see the same `expect*` function used for analysis of object shape.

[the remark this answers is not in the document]
hadley: Yeah, I much prefer the mutually exclusive arguments, as well as forcing them to be named. Your final proposal looks good to me. (Although I'd maybe order the arguments `length`, `nrow`, `ncol`, `dim` but that mostly matters for docs.)

tests/testthat/test-expect-shape.R
```diff
@@ -0,0 +1,51 @@
...
+})
+
+test_that("returns input", {
+  x <- list(1:10, letters)
+  out <- expect_shape(x, 2)
+  expect_identical(out, x)
+})
+
+test_that("at least one argument is required", {
+  expect_error(expect_shape(1:10), "`shape` must be provided for one-dimensional inputs", fixed = TRUE)
```
hadley, line 49: These should now use `expect_snapshot()` so we can review the errors.

tests/testthat/test-expect-shape.R
```diff
@@ -0,0 +1,51 @@
+test_that("shape computed correctly", {
+  # equivalent to expect_length
+  expect_success(expect_shape(1, 1))
+  expect_failure(expect_shape(1, 2), "has length 1, not length 2.")
+  expect_success(expect_shape(1:10, 10))
+  expect_success(expect_shape(letters[1:5], 5))
+
+  # testing dim()
+  expect_success(expect_shape(matrix(nrow = 5, ncol = 4), c(5L, 4L)))
+  expect_failure(expect_shape(matrix(nrow = 6, ncol = 3), c(6L, 2L)))
```
hadley, line 10: A few of these should use `expect_snapshot_failure()` so we can see the failure messages.

R/expect-shape.R
```diff
@@ -0,0 +1,76 @@
...
+    missing(nrow) || is.numeric(nrow),
+    missing(ncol) || is.numeric(ncol)
+  )
+
+  dim_object <- dim(object)
+  if (is.null(dim_object)) {
+    if (missing(shape)) {
+      stop("`shape` must be provided for one-dimensional inputs")
+    }
+    return(expect_length(object, shape))
```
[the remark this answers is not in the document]
MichaelChirico: I'm retaining this for now, re (a) yes, but it can be desirable to have a consistent entry point in a given suite rather than have to remember which to use (b) now `length=` is required so I think that's mitigated.

As commented, this helps ensure the implementations are in sync; possibly, that could mean we move the "meat" of `expect_length()` here instead and just make `expect_length()` into a stub calling `expect_shape()`.

tests/testthat/test-expect-shape.R
MichaelChirico: I don't have a good sense of when to use one vs. the other; at least one `snapshot_failure()` for each of the arguments makes sense to me superficially. I can add more plain `expect_failure()` if you see fit.

R/expect-length.R
```diff
@@ -1,6 +1,7 @@
 #' Does code return a vector with the specified length?
 #'
-#' @seealso [expect_vector()] to make assertions about the "size" of a vector
+#' @seealso [expect_vector()] to make assertions about the "size" of a vector,
```
MichaelChirico, line 3: My docker image doesn't have documentation files so I haven't managed to `roxygenize()` yet...

R/expect-shape.R
```diff
@@ -0,0 +1,57 @@
...
+  # Re-use expect_length() to ensure they stay in sync.
+  if (!missing(length)) {
+    return(expect_length(object, length))
+  }
+
+  act <- quasi_label(enquo(object), arg = "object")
+  # need base:: qualification or we might trigger an error for missing(dim)
+  dim_object <- base::dim(object)
+
+  if (!missing(nrow)) {
```
hadley, line 35: We might want to first check `length(dim_object)` since I think `expect_shape(1, nrow = 1)` will return a confusing error. Or `expect_shape(array(1), ncol = 1)`?

MichaelChirico: Great call, added some examples & fixed up the code a fair amount

R/expect-shape.R
```diff
@@ -15,42 +15,53 @@
...
-      act$ncol == ncol,
+      identical(as.integer(act$ncol), as.integer(ncol)),
       sprintf("%s has %i columns, not %i.", act$lab, act$ncol, ncol)
     )
   } else { # !missing(dim)
     act$dim <- dim_object
 
     expect(
-      isTRUE(all.equal(act$dim, dim)),
+      length(act$dim) == length(dim),
```
hadley, line 59: There is some principle that I only vaguely remember and should have written down, but I think you want to ensure that there's only one path that will somewhere yield a `success()`. (Otherwise if this `expect_shape()` succeeds, two successes will be recorded by the reporter).

So I think you should write this as `if (length(act$dim) != length(dim)) fail(sprintf(...))`

hadley: I almost wonder if using `expect()` is actually an anti-pattern because it makes it so hard to generate the correct number of successes.

MichaelChirico: Glancing, `fail()` just wraps `expect()`, though, so won't it have the same problem? Or should it just be an early return `if (!...) return(fail(...))`?

MichaelChirico: Relatedly, is there a clean way to check that there's only one success? I'm thinking with whatever the fix is, it would be good to add a few extra tests that say "succeeds once, and only once" (which would fail on the current commit)

MichaelChirico: Further aside about this hidden rake:

4bc32fca30403d406b46aa2bba313b6cf6d3daa0 found that, AIUI, because `expect()` _does_ pass for one of the two invocations, `expect_success()` was reporting "good to go" even though the latter `expect()` failed.

hadley: I think you can assume that `fail()` will always throw, so you can treat it like an early return (but you don't need the `return()`).

I'm not sure it's worth testing that an expectation returns only 1 success, because you'd need to test it for every possible code path. (And I don't believe ensuring a single success is sufficiently valuable to add that much clutter to all of the other tests)

hadley: But maybe `expect_success()` should be "expect exactly one success and no failures". I think that would help eliminate the rake you discovered.

MichaelChirico: Punting to #2110

hadley (approved):
Looks good! I'll make a few small tweaks when I'm not on a 🛩️,

hadley:
I have merged this, resisting the temptation to wait until October 4 🤣

PR #1476, opened 2021-10-14
Check and normalize snapshot file path

### LINKED ISSUE #1359: Snapshot file root directory issues
Perhaps it's my mental model of `expect_snapshot_file()` that is broken, but I'm having numerous issues related to directory paths. I created a reprex package here: https://github.com/malcolmbarrett/file.snapshot with the relevant test file here: https://github.com/malcolmbarrett/file.snapshot/blob/main/tests/testthat/test-file-snapshot.R

Here's a summary of the issues I'm having:
* testthat tells me that a snapshot is getting created even if a file doesn't exist. It seems like testthat just doesn't write an empty file even though it says it does, but it also seems to apply this logic to non-existant files.
* testthat is working from the testthat directory rather than the package root. I believe that's by testthat's design, and normally this probably makes sense, but I think this either should be addressed or better explained in the docs for this type of test. 
* If I change the working directory, the snapshots are not created even if I provide absolute paths

I suspect this is a mixture of intentional and unintentional behavior (and possibly my own misunderstanding!) but unfortunately, I'm not sure which is which.

malcolmbarrett:
This PR improves `expect_snapshot_file()` in the following ways:
* Checks that the file being snapshotted exists and errors if not
* Normalizes the snapshot directory, better allowing files from other directories
* Adds tests for both situations

Closes #1359


tests/testthat/test-snapshot-file.R
```diff
@@ -31,6 +31,29 @@ test_that("expect_snapshot_file works", {
...
+  current_wd <- getwd()
+  path <- tempfile()
+  dir.create(path)
+
+  withr::defer({
+    setwd(current_wd)
+    unlink(path)
+  })
+
+  setwd(path)
```
hadley, line 46: I think you can replace all this with `withr::local_tempdir()` 😄

malcolmbarrett: Good point!

hadley:
Thanks!

PR #1489, opened 2021-11-12
dont indent() an empty message

### LINKED ISSUE #1487: snapshot transform "leaves a trace" when returning character(0)?
I am trying to deal with some Windows-specific output issues while converting to snapshot tests:

https://github.com/MichaelChirico/potools/pull/257, as of 8558cf2479195e41a783677c4141eebf933a7a05

To deal with some output that is not produced on Windows, I wrote this helper:

```
normalize_output <- function(x) {
  # different platforms/installations of gettext apparently
  #   produce a different number of "." in "progress" output; normalize
  x <- gsub("\\.{2,}", ".", x)

  # en@quot translations are not produced on Windows (as of now)
  x <- grep("Generating en@quot translations", x, fixed = TRUE, invert = TRUE, value = TRUE)

  # this is produced alongside the previous message, but in a different iteration of transform()
  idx <- grep("running msgfmt on (?:R-)?en@quot\\.po", x)
  if (length(idx)) {
    browser()
    x <- x[-(idx + 0:1)]
  }
  x
}
```

Whenever either of the `grep()` match the snapshot input `x`, the resulting output `x` is `character()` (confirmed by `browser()`-ing `normalize_output()` while running tests interactively). That is as intended -- I want the snapshot to act as if that output was never produced, in order to imitate the corresponding output on Windows.

However, on Windows, my CI build still fails:

https://github.com/MichaelChirico/potools/runs/4174597881?check_suite_focus=true

```
-- Failure (test-translate-package.R:101:1): (code run outside of `test_that()`) --
Snapshot of code has changed:
old[2:8] vs new[2:7]
  "  translate_package(languages = \"cy\")"
  "Message <simpleMessage>"
  "  Writing R-rMsg.pot"
- "  "
  "  'cy' is not a known language. "
  "  Please help supply some metadata about it. You can check https://l10n.gnome.org/teams/<language>"
  "Output"
```

It seems `- "  "` is a residual line left after running `normalize_output()`. Either that, or the `"  "` line is not supplied to `transform()` and thus not available to remove. Either way, it seems I have no way of matching my local snapshot to that produced on Windows CI.

Laterally, is it feasible to support Windows/platform-specific snapshots? Windows in particular seems potentially fruitful given how much the Windows/non-Windows builds of R already differ (terminal access, file paths, etc). I'd rather not use `skip_on_os('windows')` (e.g., the same failing tests seem to be showing me some encoding issues that I need to fix, so keeping the snapshots running on windows would be ideal).

MichaelChirico:
Fixes #1487 -- `character()` is not passed to `indent()`, so `"  "` won't show up in the output.

I'm not sure this is a complete solution, or if it might have unintended consequences, so I am just filing this simple PR demonstrating a fix for now before attempting to add tests/NEWS.

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -134,7 +134,8 @@ snapshot_lines <- function(x, transform = NULL) {
   if (!is.null(transform)) {
     x <- transform(x)
   }
-  x <- indent(x)
+  # if transform() wiped out the full message, don't indent, #1487
+  if (length(x)) x <- indent(x)
   x
 }
 
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -134,8 +134,7 @@ snapshot_lines <- function(x, transform = NULL) {
   if (!is.null(transform)) {
     x <- transform(x)
   }
-  # if transform() wiped out the full message, don't indent, #1487
-  if (length(x)) x <- indent(x)
+  x <- indent(x)
   x
 }
 
@@ -313,7 +312,8 @@ local_snapshot_dir <- function(snap_names, .env = parent.frame()) {
   path
 }
 
-indent <- function(x) paste0("  ", x)
+# if transform() wiped out the full message, don't indent, #1487
+indent <- function(x) if (length(x)) paste0("  ", x) else x
 
 check_variant <- function(x) {
   if (is.null(x)) {


R/snapshot.R
```diff
@@ -134,7 +134,8 @@ snapshot_lines <- function(x, transform = NULL) {
   if (!is.null(transform)) {
     x <- transform(x)
   }
-  x <- indent(x)
+  # if transform() wiped out the full message, don't indent, #1487
+  if (length(x)) x <- indent(x)
```
hadley, line 138: I think it would be better to fix this in indent, so that `indent(character())` returns `character()`. That preserves the invariant that the `indent()` returns a character vector the same length as the input.

(The root cause is that `paste()`'s recycling rules aren't quite what we want here)

hadley:
Thanks!

PR #1491, opened 2021-11-12
Add support for rlang 1.0.0

lionel-:
Depends on r-lib/rlang#1313.

* Update snapshots
* Don't show error class by default. New argument `cnd_class` allows opting into the old behaviour.
* Use rlang to print full prefixed error and warnings messages. Conditions of class `message` are still printed without prefix.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,17 @@
   like `!!`.
 
 * Fixed bug in expectations with long inputs that use `::` (#1472).
+* Snapshots now use rlang to print error and warning messages,
+  including the `Error:` and `Warning:` prefixes. This means the
+  `call` field of conditions is now displayed in snapshots if present.
+
+* By default, snapshots no longer show the class of errors, warnings,
+  and messages. Set the new argument `cnd_class = TRUE` to restore the
+  old behaviour.
+
+  We decided to change the default to prevent distracting snapshot
+  diffing as upstream packages add error classes. For instance, the
+  development version of R is adding classes to basic errors.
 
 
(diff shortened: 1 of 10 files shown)


NEWS.md
```diff
@@ -1,5 +1,18 @@
...
+  including the `Error:` and `Warning:` prefixes. This means the
+  `call` field of conditions is now displayed in snapshots if present.
+
+* By default, snapshots no longer show the class of errors, warnings,
+  and messages. Set the new argument `cnd_class = TRUE` to restore the
+  old behaviour.
+
+  We decided to change the default to prevent distracting snapshot
+  diffing as upstream packages add error classes. For instance, the
+  development version of R is adding classes to basic errors.
```
DavisVaughan, line 13: I like mentioning this as a reason for this change. I just had a case where I was using `warn_deprecated()` from `compat-lifecycle.R`, which calls `.Deprecated()`. My GHA 3.5.3 build failed a snapshot test because the error class used to be `"simpleWarning"` but they have recently added the class `"deprecatedWarning"`

lionel-:
@hadley This should now be ready.

- rlang 1.0 display is only used if the tested package explicitly depends on rlang 1.0 or the current dev version.

- `expect_snapshot()` gains `cnd_class` to control whether to include condition classes in snapshots but the default hasn't changed for now, unless rlang 1.0 is explicitly depended upon. This way we minimise the number of snapshot revalidations needed from users.

- Snapshot variants test the output pre- and post-rlang 1.0

hadley:
Thanks!

PR #1500, opened 2021-11-27
Add snapshot_review() in the hint of snapshots

DanChaltiel:
Hi, 

Copypasting the `Run snapshot_accept("my_file") if this is a deliberate change` hint to the console is a game-changer when using snapshots! 

However, one does not often blindly accept changes, so `snapshot_review()` seems more adapted, at least to my use.

Not sure about the grammar though (I never know when "change" should be plural in English).

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -280,7 +280,7 @@ expect_snapshot_helper <- function(lab, val,
     tolerance = tolerance,
     variant = variant
   )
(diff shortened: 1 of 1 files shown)


hadley:
But don't you already see the differences interactively? (and in git, should the interactive display not be sufficient?)

hadley:
Needs to wait on #1531

PR #1503, opened 2021-12-01
namespace hint?

maelle:
Like https://github.com/r-lib/testthat/blob/f8ccf033a74d89e2d1d3cb2ec3c6c4f63ca69055/R/snapshot-file.R#L144, for easier use by copy-pasting :-)

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -327,7 +327,7 @@ expect_snapshot_helper <- function(lab, val,
     tolerance = tolerance,
     variant = variant
   )
-  hint <- paste0("Run `snapshot_accept('", snapshotter$file, "')` if this is a deliberate change")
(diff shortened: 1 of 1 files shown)


hadley:
Why isn't testthat attached for you? `load_all()` will automatically attach it.

maelle:
Right, maybe it happens in cases where I don't load_all() before testing? Or maybe I should close this PR. :see_no_evil:

maelle:
I can confirm it happens because I don't necessarily run `load_all()` before running the tests (Ctrl + Alt + T).

maelle:
To add a bit more context, it is a package for which I actually nearly never run `load_all()`, all tests happen in temporary projects. So I'd appreciate the namespace, but understand it might be too niche.

hadley:
Lets leave this as is; at least for me it's a useful hint to run `load_all()`.

PR #1505, opened 2021-12-02
Add `in_snapshotter()` and return conditions visibly when snapshotting

DavisVaughan:
This is currently broken, because `in_snapshotter()` seems to _always_ return `TRUE` when testing a file or directory, and _always_ return `FALSE` when running a test interactively. Both of these are incorrect.

We need something that is more fine grained. I think `expect_snapshot()` itself might need to set something that announces that you are "in a snapshot expectation". That would mean that both interactive and non-interactive testing could return visible conditions where appropriate.

### DIFF
--- a/R/expect-condition.R
+++ b/R/expect-condition.R
@@ -264,8 +264,17 @@ expect_condition_matching <- function(base_class,
   expect(is.null(msg), msg, info = info, trace = act$cap[["trace"]])
 
   # If a condition was expected, return it. Otherwise return the value
-  # of the expression.
-  invisible(if (expected) act$cap else act$val)
+  # of the expression. If we are returning the condition and we are in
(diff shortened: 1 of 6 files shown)


hadley:
I'm now having second thoughts about conditionally changing the visibility depending on whether or not you're in a snapshot or not. It just feels a bit too magical. What do you think?

hadley:
Part of #1471

DavisVaughan:
If we need to continue wrapping `expect_error()` in another set of parenthesis to avoid this feeling too magical, that is okay with me.

I would like to avoid seeing the `err()` wrapper in too many places, I'd rather be explicit about the `(expect_error())` pattern
(mentioned in https://github.com/r-lib/testthat/issues/1471#issuecomment-962979862)

PR #1511, opened 2021-12-07
Fix snapshots of conditions with empty lines

### LINKED ISSUE #1509: testthat 3.1.1 and empty lines in snapshots
With testthat 3.1.1 empty lines are removed from snapshots.

E.g. if a function is

```r
bla <- function() {
  cli::cli_inform("thing")
  cli::cli_h2("bla")
  cli::cli_inform("another thing")
}
```

the snapshot will show no empty lines before and after the H2 whereas it used to.

gaborcsardi:
It seems to me that the simplest is to treat snapshots of conditions differently. So for these we dot remove the empty lines now. 

To have a workaround that works with both testthat 3.1.1 (removing the empty lines), record your test with testthat 3.1.0 or with the fix included in this PR (will be version 3.1.1.9001) and skip them on the problematic testthat versions. E.g. something like this in a `helper.R` file should work:

```r
expect_snapshot <- function(...) {
  if (packageVersion("testthat") >= "3.1.1" &&
      packageVersion("testthat") <= "3.1.1.9000") {
    skip("testthat bug with snapshots")
  }
  testthat::expect_snapshot(...)
}
```

Closes #1509.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: testthat
 Title: Unit Testing for R
-Version: 3.1.1.9000
+Version: 3.1.1.9001
 Authors@R: 
     c(person(given = "Hadley",
              family = "Wickham",
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -154,7 +154,7 @@ snapshot_replay.condition <- function(x,
     type <- paste0(type, " <", class(x)[[1]], ">")
   }
 
-  c(snap_header(state, type), snapshot_lines(msg, transform))
+  c(snap_header(state, type), snapshot_lines(msg, transform, keep_empty = TRUE))
 }
 
 snapshot_replay_condition_legacy <- function(x, state, transform = NULL) {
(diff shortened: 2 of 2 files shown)


hadley:
I think my mental model of this code must be missing something because I don't understand why an empty line corresponds to `character()`?

gaborcsardi:
I think it is because the message is a single `\n` with `appendLF = FALSE` (or an empty string with `appendLF = TRUE`) and this will end up being a `character()` when testthat breaks it apart into lines.

hadley:
I think the root cause is this weirdness in `strsplit()`:

``` r
strsplit("a\nb", "\n")[[1]]
#> [1] "a" "b"
strsplit("a\nb\n", "\n")[[1]]
#> [1] "a" "b"

strsplit("a\n", "\n")[[1]]
#> [1] "a"
strsplit("a", "\n")[[1]]
#> [1] "a"

strsplit("", "\n")[[1]]
#> character(0)
strsplit("\n", "\n")[[1]]
#> [1] ""
```

<sup>Created on 2022-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

So I think a slightly cleaner fix will be in `snapshot_lines()`. I'll do it since I'll be working on testhat for a bit.

hadley:
Closing in favour of #1524

PR #1512, opened 2021-12-08
Set `cli.num_colors` in addition to `crayon.enabled`

lionel-:
I've set `cli.num_colors` to a custom value in my RProfile today, to override the default that is assigned to ESS. This caused tests to fail because `cli.num_colors` has precedence over `crayon.enabled`. To fix this, the local reproducibility options now set both values.


lionel-:
Now also sets `cli_theme_dark`.

@gaborcsardi Does cli need a `cli::local_reproducible()` to take care of using a default reproducible theme, disabling colours, etc?

lionel-:
@gaborcsardi Setting `cli_theme_dark` does not seem to work, I guess for the same reason that you need to reload RStudio after changing the theme. Is there any way to clear options cached by cli?

gaborcsardi:
IDK, there are a lot of options that modify the cli output, I am not sure if the right way to handle them is to add everything to `local_reproducible_context()`.

lionel-:
Agreed, this is why I suggested that cli exports a `local_reproducible()` function to handle all of this.

gaborcsardi:
Or we can just have a single env var / option that forces everything to the default.

lionel-:
Good point, that would have the big advantage of working across packages.

And for the rare cases where we need to reset cached options, we can still provide `local_` functions if needed.

gaborcsardi:
I am not so sure any more that a single turn-off option would work, because you'd still want to customize some of the options. In any case, I don't think we need to add the option for the theme here. You can run the tests in a clean process until this is fixed in cli.

lionel-:
Now reverted.

R/local.R
```diff
@@ -101,6 +102,7 @@ local_reproducible_output <- function(width = 80,
     cli.dynamic = FALSE,
     cli.unicode = unicode,
     cli.condition_width = Inf,
+    cli.num_colors = if (crayon) 8L else 0L,
```
gaborcsardi, line 105: I would set it to 1 instead of 0. 0 might work, but nevertheless we usually use 1:
```r
❯ withr::with_envvar(c(NO_COLOR = "true"), cli::num_ansi_colors())
[1] 1
```

R/local.R
```diff
@@ -31,6 +31,7 @@
 #' * `crayon.enabled` (default: `FALSE`) suppresses ANSI colours generated by
 #'   the crayon package (normally colours are used if crayon detects that you're
 #'   in a terminal that supports colour).
+#' * `cli.num_colors` (default: `0L`) Same as the crayon option.
```
gaborcsardi, line 34: Default should be 1 here as well.

R/local.R
lionel-: Done, thanks!

gaborcsardi:
Thanks! Look good to me now.

PR #1524, opened 2022-01-03
Preserve empty lines in snapshot output

### LINKED ISSUE #1509: testthat 3.1.1 and empty lines in snapshots

hadley:
Fixes #1509


hadley:
The root cause is that some conditions implicitly adding a trailing new line, and some implicitly swallow a trailing new line. This now matches the behaviour I see on the console.

tests/testthat/_snaps/snapshot.md
```diff
@@ -22,11 +22,44 @@
...
+      f()
+    Output
+      1
+      
+    Message
+      2
+      
+    Condition
+      Warning in `f()`:
+      3
```
hadley, line 59: @lionel- should this get a prefix? And I think something in rlang needs to change in order to preserve the trailing nl? (Alternatively, maybe we only care about this for output and messages?)

hadley: I think something about how rlang/cli is reformatting this makes it impossible for testthat to recover the trailing nl here.

gaborcsardi:
FWIW this PR does not fix the test cases in cli. After removing the temporary `cli::expect_snapshot()` helper function now I get:

```
testthat::test_local(filter = "contain")
```

```
✔ | F W S  OK | Context
✖ | 1       6 | containers [0.3s]
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Failure (test-containers.R:78:3): margin is squashed
Snapshot of code has changed:
     old                 | new
[23] "  "                | "  "                [23]
[24] "  "                | "  "                [24]
[25] "  "                | "  "                [25]
[26] "  until here"      - "  "                [26]
[27] "  "                - "  until here"      [27]
[28] "  "                | "  "                [28]
[29] "  "                | "  "                [29]
[30] "  no space, still" - "  "                [30]
[31] "  "                | "  "                [31]
[32] "  "                - "  no space, still" [32]
 ... ...                   ...                 and 5 more ...

Run `snapshot_accept('containers')` if this is a deliberate change
```

Is this expected?

hadley:
@gaborcsardi do you mind boiling that down to a more minimal reprex? I find it hard to reason about exactly what should occur given all the `\r`.

gaborcsardi:
This test was actually not broken in version 3.1.1, but this PR breaks it. It is not about `\r` but messages that consist of only `\n` characters.

## Example

```r
library(testthat)
testthat::test_that("snapshots and whitespace", {
  local_edition(3)
  msgs <- list("start\n", "\n\n", "until here\n", "\n\n",
    "still\n", "\n\n", "again\n")
  expect_snapshot(local({
    for (m in msgs) message(m, domain = NA, appendLF = FALSE)
  }))
})
```

## testthat 3.1.1 output

```
i Current value:
Code
  local({
    for (m in msgs) message(m, domain = NA, appendLF = FALSE)
  })
Message <simpleMessage>
  start

  until here

  still

  again
```

## testthat #1524 output

```
i Current value:
Code
  local({
    for (m in msgs) message(m, domain = NA, appendLF = FALSE)
  })
Message <simpleMessage>
  start


  until here


  still


  again
```

hadley:
I think this PR is correct, because when I run this at the console I see:

```
start


until here


still


again
```

But I think evaluate or knitr gets it wrong because this is the reprex:

``` r
msgs <- list("start\n", "\n\n", "until here\n", "\n\n",
             "still\n", "\n\n", "again\n")
local({
  for (m in msgs) message(m, domain = NA, appendLF = FALSE)
})
#> start
#> 
#> until here
#> 
#> still
#> 
#> again
```

<sup>Created on 2022-01-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

But this is what I see if I run at the console:

hadley:
i.e. this is wrong:

``` r
msgs <- c("line 1", "\n", "line 4")
local({
  for (m in msgs) message(m)
})
#> line 1
#> 
#> line 4
```

<sup>Created on 2022-01-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

gaborcsardi (approved):

PR #1525, opened 2022-01-04
Only use part of local_reproducible_output() in Reporter$local_user_output()

### LINKED ISSUE #1285: Can we make it easier to test UTF-8 handling?
Porting the UTF-8 tests of rlang to testthat3 was a bit tricky. The `waldo_compare()` function called by all testthat expectations calls `reporter$local_user_output()` which automatically skips the test when the locale is set to non-utf-8.

I found this workaround:

```r
local_utf8_test <- function(frame = caller_env()) {
  reporter <- get_reporter()

  old <- reporter$unicode
  defer(reporter$unicode <- old, envir = frame)

  reporter$unicode <- FALSE
}
```

This doesn't seem like a proper long-term solution though.

hadley:
Fixes #1285


gaborcsardi (approved):
This looks good to me, but just to make sure, @lionel- does it fix your problem?

lionel- (approved):
It works, thanks!

PR #1531, opened 2022-01-05
Make it easier to accept snapshot variants

### LINKED ISSUE #1508: Snapshot management in the presence of a variant
First I am aware that variants are experimental. :sweat_smile: 

When using variants it seems I can't use `snapshot_accept()`. Example https://github.com/maelle/clii where I changed the code after recording a snapshot. When I run the tests again I get

```r
Snapshot of code has changed:
    old                    | new                       
[4] "Message <message>"    | "Message <message>"    [4]
[5] "  thing"              | "  thing"              [5]
[6] "Message <cliMessage>" | "Message <cliMessage>" [6]
[7] "  -- bla --"          - "  -- blaaa --"        [7]
[8] "Message <message>"    | "Message <message>"    [8]
[9] "  another thing"      | "  another thing"      [9]

Run `snapshot_accept('bla')` if this is a deliberate change
```

But if I run `snapshot_accept('bla')` I get the message "No snapshots to update".

hadley:
Fixes #1508. 

@maelle would you mind kicking the tires on this and making sure it works for your use case?


hadley:
Just fixed another annoying buglet — if a snapshot changed in one variant, `.new` files were written for all variants, even if they didn't have any changes.

hadley:
I've merged this because I actually need to use it for testthat itself, but if you see any other problems let me know!

PR #1533, opened 2022-01-05
Check with dev rlang

hadley:


hadley:
The remaining failures are mostly because `"Message <message>"` is now `"Message <rlang_message>"` (when using `inform()`).  Since that seems likely to cause widespread snapshot changes, I wonder if we should work around it somehow. What do you think @lionel-?

hadley:
What if we just ignored the class and did `Message <message>` and `Warning <warning>` (or whatever the most common values are)? I'm also starting to see snapshot failures in R-devel because of the new condition classes.

lionel-:
> What if we just ignored the class and did Message <message> and Warning <warning> (or whatever the most common values are)?

Do you mean that we'd only ignore the rlang classes but not others? For instance `c("rlang_warning", "warning", "condition")` would show `Warning <warning>` but `c("foo", "rlang_warning", "warning", "condition")` would show `Warning <foo>`? Then do we treat `simpleWarning` and `simpleMessage` the same way?

I'm a bit worried about hiding only some classes.

On the other hand I'm not really concerned about creating snapshot noise for developers, it's only a mild annoyance as long as the package that causes noise is on CRAN. It's annoying only when dealing with dev versions because there are multiple versions of the truth to juggle with, but that only concerns us.

That's why the snapshot diffs caused by multiple versions of R seem much more problematic because there are multiple versions to deal with on CI. I don't see any other solution than using snapshot variants or not showing the class at all in snapshots.

So I think we have to make a trade off between:

1. Letting developers track condition classes.
2. Dealing with multiple versions of truth on CI.

It's worth noting that (2) is not only a problem with classes but also with messages.

Perhaps a general way of dealing with this is to silently skip snapshot failures on CI for versions other than r-release? Unless a specific variant is created for a version?

hadley:
I meant ignore _all_ classes (so that the class temporarily becomes redundant, in preparation to the switch to the new format). This means effectively we're not showing the class in the snapshot, but doing it in a way that doesn't change most existing snapshots.

I think you're underestimating how much pain this is/will cause; I'm having to switch between dev and CRAN rlang many times per day.

lionel-:
> I think you're underestimating how much pain this is/will cause; I'm having to switch between dev and CRAN rlang many times per day.

I understand but isn't the rlang 1.0 release a very specific (and lingering) case? Also I would expect other diffing noise caused CRAN/dev rlang differences than just error classes.

That said I'm fine with removing error classes from snapshots if you think they don't add sufficient value. Base R is more likely to change error classes than error messages, so that could be a pragmatic solution to the management of error snapshots across R versions on CI.

lionel-:
> I meant ignore all classes (so that the class temporarily becomes redundant, in preparation to the switch to the new format). This means effectively we're not showing the class in the snapshot, but doing it in a way that doesn't change most existing snapshots.

Sorry I'm not sure I fully understand your proposal. Do you mean that we'd temporarily hardcode the display of the class until we switch to the new format?

hadley:
> Sorry I'm not sure I fully understand your proposal. Do you mean that we'd temporarily hardcode the display of the class until we switch to the new format?

Yes. I might be wrong, but I don't think this will affect much existing code outside of the tidyverse, since it's relatively rare for folks to use custom condition classes.

hadley:
Lets leave this as is for now, given that the rlang release is imminent.

Issue #1461, opened 2021-09-28
Should initial snapshot error on CI?


hadley:
Since this usually indicates that you've forgotten to check the snapshot in.

hadley:
Activating now just requires turning on in one place: `local_snapshotter("_snaps", fail_on_new = on_ci(), .env = .env)`

PR #1539, opened 2022-01-06
Fail when a new variant is found on CI (not just warn)

MichaelChirico:
As mentioned in #1538, a failure appears to be necessary to generate artifacts.

I am not sure if this is a GHA-only thing, in which case we could try fine-tuning this more. Also happy to change the approach -- e.g. if we want to just `stop()` in lieu of `testthat_warn()`, rather than in addition to it as is done here.

Filing this approach now for feedback / to start discussion.

### DIFF
--- a/R/expect-condition.R
+++ b/R/expect-condition.R
@@ -153,7 +153,8 @@ expect_warning <- function(object,
       ...,
       inherit = inherit,
       info = info,
-      label = label
+      label = label,
+      trace_env = caller_env()
     )
   } else {
     act <- quasi_capture(enquo(object), label, capture_warnings)
@@ -185,7 +186,8 @@ expect_message <- function(object,
       ...,
       inherit = inherit,
       info = info,
-      label = label
+      label = label,
+      trace_env = caller_env()
     )
   } else {
     act <- quasi_capture(enquo(object), label, capture_messages)
@@ -213,7 +215,8 @@ expect_condition <- function(object,
       ...,
       inherit = inherit,
       info = info,
-      label = label
+      label = label,
+      trace_env = caller_env()
     )
   } else {
     act <- quasi_capture(enquo(object), label, capture_condition, entrace = TRUE)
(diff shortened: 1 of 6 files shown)


hadley:
The number of build failures suggest that this is (a) a good idea (since these tests weren't previously doing anything) and (b) is likely to cause existing packages to start failing on CI. Overall, I'm almost certain we should do this, my main question is whether we should do this in 3.1.2 or it's going to be a big enough of a change that it needs to be in 3.2.0.

MichaelChirico:
3.2.0 sounds good. I would venture most of the CI failures it'll cause were unintentional / users didn't realize there was an issue to begin with... but still better to leave disruptive changes for "bigger" version bumps

hadley:
Fixes #1461

hadley:
Also needs a similar fix in `snapshot_file_equal()`

hadley:
Logic isn't quite right yet because the snapshotting tests shouldn't fail in CI, just warn.

hadley:
@MichaelChirico can you please take a look at the tests output for the failing results and let me know what you think? If it all looks good to you, I'll set `fail_on_new` to `FALSE` (instead `on_ci()`) and then merge this PR. Then in the next minor release I can turn the new behaviour back on by changing a single line of code.

MichaelChirico:
I had a look and I'm not fully following. Is the `trace_env` part separate to this PR?

Maybe it would help to see some example input/output for before/after the PR (I don't see any new/changed tests)

hadley:
Sorry, the output now looks like:

```
== Failed tests ================================================================
-- Failure (test-snapshot-file.R:50:3): expect_snapshot_file works with variant --
Adding new file snapshot: 'tests/testhat/_snaps/nickname.txt'
-- Failure (test-snapshot.R:2:3): can access nickname --------------------------
Adding new snapshot for variant 'R3.6' in CI:
Code
  version$nickname
Output
  [1] "Holding the Windsock"
```

Does that look reasonable to you?

The `trace_env` stuff is about ensuring that you don't get a traceback here, because the error comes from a failing expectation immediately within `test_that()` (i.e. not an error inside your code, or an expectation called within a wrapper function).

hadley:
In the interests of getting this patch release out, I've turned the switch off. If you spot any improvements to the output, we can incorporate in the next minor release.

MichaelChirico:
LGTM, thanks for taking over the patch!

PR #1543, opened 2022-01-07
In CheckReporter, also display summary at top

hadley:
This is useful when look at the full results in CI

### DIFF
--- a/R/reporter-check.R
+++ b/R/reporter-check.R
@@ -35,6 +35,14 @@ CheckReporter <- R6::R6Class("CheckReporter",
     },
 
     end_reporter = function() {
+      self$cat_line(summary_line(
+        n_fail = self$problems$size(),
+        n_warn = self$warnings$size(),
+        n_skip = self$skips$size(),
+        n_pass = self$n_ok
+      ))
+      self$cat_line()
+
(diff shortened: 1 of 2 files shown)


gaborcsardi (commented):
Should we add some optimization to avoid showing the summary twice if there is no other output? I.e. if there are no skips, warnings, errors, etc. at all?

hadley:
I thought about that but couldn't be bothered since it seems unlikely to come up that much.

gaborcsardi:
It comes up for every clean test suite, no?

Not a big deal, just a weird to have this duplication.

```
> library(testthat)
> library(pingr)
>
>  test_check("pingr")
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 21 ]

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 21 ]
>
> proc.time()
   user  system elapsed
  0.297   0.085   3.724
```

hadley:
Oh I see what you mean. I guess I was just thinking about the actions, where you only ever look at this display if there are test failures.

hadley:
I'll fix momentarily (`if (self$skips$size() || self$warnings$size() || self$problems$size()))`

PR #1549, opened 2022-01-10
Introduce tolerance to inequality comparisons (non-strict)

MKyhos:
Currently, the functions `expect_gte` and `expect_lte` have not implemented means to introduce a tolerance for the "equality" expectation part (as for example is done in `expect_equal(numeric(1), numeric(1))`.  However, in some cases it might be desirable to allow for a certain amount of tolerance. (Please excuse me and ignore this PR in case this possibility was already discussed and discarded in the past :see_no_evil: :v:)

In this PR, a possible solution is implemented that, for example in the `lte` case, substitutes (roughly speaking, see the changed code for actual implementation) `a <= b` with `a < b || all.equal(a, b, tol)`.

Issues/tbd:
- currently, it is assumed that `expect_lte(NA_real_, NA_real_)` fails. However, `all.equal(NA, NA)` usually evaluates to `TRUE`. This behavior is circumvented by the fact that once the first value is NULL or NA, the comparison shall never evaluate to `TRUE`.
- imo waldo comparison function for e3 are not required here, as the functions are for comparisons of length-1 vectors anyways

### Benchmarks

The changes would come at cost of a small decrease in speed, see the benchmark below (up to the maintaining developers to debate the significance :grimacing: :grinning: ) .

```r
library(devtools)
tmp_dir <- tempdir()
file.copy(".", tmp_dir, recursive = TRUE)
system(sprintf("cd %s && git checkout main", tmp_dir))

load_all(tmp_dir)
expect_compare_classic <- testthat:::expect_compare

load_all(".")
expect_compare_tol <- testthat:::expect_compare

ps_act <- list(val = 0)
ps_exp <- list(val = 1)

bench::mark(
  expect_compare_classic("<=", ps_act, ps_exp),
  expect_compare_tol("<=", ps_act, ps_exp)
)
```

**Results:**
```r
# A tibble: 2 x 13
  expression                                        min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    memory                 time             gc                  
  <bch:expr>                                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    <list>                 <list>           <list>              
1 expect_compare_classic("<=", ps_act, ps_exp)   50.3µs   57.7µs    16882.        0B     34.0  7939    16      470ms <dbl [1]> <Rprofmem[,3] [0 × 3]> <bch:tm [7,955]> <tibble [7,955 × 3]>
2 expect_compare_tol("<=", ps_act, ps_exp)       78.7µs   89.3µs    10861.        0B     32.0  5089    15      469ms <dbl [1]> <Rprofmem[,3] [0 × 3]> <bch:tm [5,104]> <tibble [5,104 × 3]>
```

### Checklist

- [x] Tests have been adapted / added
- [x] PR description written
- [x] Documentation updated

### DIFF
--- a/R/expect-comparison.R
+++ b/R/expect-comparison.R
@@ -19,9 +19,23 @@
 #' @name comparison-expectations
 NULL
 
-expect_compare <- function(operator = c("<", "<=", ">", ">="), act, exp) {
+expect_compare <- function(operator = c("<", "<=", ">", ">="),
+                           act, exp,
+                           tolerance = 0) {
   operator <- match.arg(operator)
-  op <- match.fun(operator)
+
+  if (operator %in% c("<=", ">=")) {
+    eq <- isTRUE(all.equal(act$val, exp$val, tolerance = tolerance, check.attributes = TRUE))
+    # Fence against all.equal(NA, NA) and all.equal(NULL, NULL) being TRUE
+    eq <- eq && !any(is.na(act$val) | is.null(act$val))
+  } else {
+    eq <- FALSE
+    if (tolerance != 0) {
+      message(sprintf("Tolerance (%g) will be ignored for operator '%s'.", tolerance, operator))
+    }
+  }
+
+  op <- match.fun(substr(operator, 1, 1))
 
   msg <- c(
     "<" =  "not strictly less than",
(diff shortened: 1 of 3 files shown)


R/expect-comparison.R
```diff
@@ -30,7 +46,7 @@ expect_compare <- function(operator = c("<", "<=", ">", ">="), act, exp) {
     ">=" = "not more than"
   )[[operator]]
 
-  cmp <- op(act$val, exp$val)
+  cmp <- op(act$val, exp$val) | eq
```
MKyhos, line 49: It was necessary to use `|` instead of `||` as the length of `op()` is checked just below, so at this point we cannot assume it is unity.

MKyhos:
Closing the PR, as on a second view it does not seem to be very beneficial.

PR #1565, opened 2022-02-08
add workflow for upgrading to the 3rd edition

hfrick:
It's somewhat of a repetition of the changes but formulated as more hands-on guidance on what to do with that information. Alternatively we could also just mention `usethis::use_testthat(3)` in the `Activating` section.


hadley:
Thanks!

PR #1568, opened 2022-02-09
Update to rlang 1.0

lionel-:
Update snapshots and use `local_bindings()` instead of deprecated `scoped_bindings()` function.

I've added a remote to dev rlang because the callr test in https://github.com/r-lib/testthat/blob/main/tests/testthat/test-parallel-crash.R fails with CRAN rlang. It is fixed by this commit: https://github.com/r-lib/rlang/commit/c84d52b6f01aaad98adfc57c29bca64a655f636e. This won't get in the way of a testthat patch release because this test is skipped on CRAN.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,6 +56,8 @@ Suggests:
     xml2
 VignetteBuilder: 
     knitr
+Remotes:
+    r-lib/rlang
 Config/testthat/edition: 3
 Config/testthat/start-first: watcher, parallel*
 Config/Needs/website: tidyverse/tidytemplate
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,7 +40,7 @@ Imports:
     R6 (>= 2.2.0),
     rlang (>= 0.4.9),
     utils,
-    waldo (>= 0.2.4),
+    waldo (>= 0.3.1),
     withr (>= 2.4.3)
 Suggests: 
     covr,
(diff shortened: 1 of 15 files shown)


lionel-:
@hadley The codecov failure is because it has different srcrefs in the snapshots. We could fix this by removing backtrace srcrefs from these snapshots with `options(rlang_trace_format_srcrefs = FALSE)`. Does that sound good?

I'm not sure what's causing the window 3.6 snapshot failures, e.g.

```
 - "`actual` is \033[32man integer vector\033[39m (1)"
+ "`actual` is \033[32minteger\033[39m (1)"
```

Do you know where these strings are generated?

hadley:
I think it would be better to skip those snapshot tests during coverage than generally lose the ability to check that we're printing srcrefs in the right place. 

I think those messages are coming from waldo:

``` r
waldo::compare(1, "a")
#> `old` is a double vector (1)
#> `new` is a character vector ('a')
```

<sup>Created on 2022-02-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I can't see why it would change on windows, but it's probably related to inlining `rlang::friendly_type` in waldo (https://github.com/r-lib/waldo/commit/2d0ef95c3d7737c56caabc800cfdec3a7d82b6a1) so maybe the it's getting an old version of waldo?

...

Yeah, that run is installing waldo 0.2.5. I don't know why though 😞

hadley:
Ah that makes sense. So we should explicitly require latest waldo in this RP.

PR #1581, opened 2022-03-02
Preserve `lifecycle_verbosity` value in `test_that()` blocks

lionel-:


hadley:
I can't figure out if this is a good idea or not. I can see the benefit, but is it inconsistent to only do this for one option?

hadley:
Closing since it's been 3 years

PR #1596, opened 2022-03-29
Move callr and friends to Suggests

lionel-:
To make webR support easier until we figure out callr.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,6 @@ Depends:
     R (>= 3.1)
 Imports:
     brio,
-    callr (>= 3.5.1),
     cli (>= 3.1.1),
     crayon (>= 1.3.4),
     desc,
@@ -35,19 +34,20 @@ Imports:
     methods,
     pkgload,
     praise,
-    processx,
-    ps (>= 1.3.4),
     R6 (>= 2.2.0),
     rlang (>= 1.0.2),
     utils,
     waldo (>= 0.3.1),
     withr (>= 2.4.3)
 Suggests: 
+    callr (>= 3.5.1),
     covr,
     curl (>= 0.9.5),
     diffviewer (>= 0.1.0),
     knitr,
     mockery,
+    processx,
+    ps (>= 1.3.4),
     rmarkdown,
     rstudioapi,
     shiny,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -38,7 +38,7 @@ Imports:
     processx,
     ps (>= 1.3.4),
     R6 (>= 2.2.0),
-    rlang (>= 1.0.1),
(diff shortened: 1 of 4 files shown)


gaborcsardi:
This will probably break packages that have 
```
Config/testthat/parallel: true
```
on CRAN.

gaborcsardi:
An alternative would be to qualify all calls with `::` and then they are not loaded with testthat. If it is possible to install ps, processx and callr, then this can work, and it does not break anything.

gaborcsardi:
>  An alternative would be to qualify all calls with :: and then they are not loaded with testthat. If it is possible to install ps, processx and callr, then this can work, and it does not break anything.

Which is already the case AFAICT:
```
❯ R -q --vanilla -e 'library(testthat); loadedNamespaces()'
> library(testthat); loadedNamespaces()
 [1] "compiler"  "magrittr"  "R6"        "cli"       "graphics"  "utils"
 [7] "grDevices" "stats"     "datasets"  "methods"   "brio"      "testthat"
[13] "rlang"     "base"
```

lionel-:
It is currently not possible to install callr, ps, and processx. It might be possible without too many changes but I haven't investigated it yet. At the moment I am trying to make testthat work (it causes webR to freeze).

> This will probably break packages that have `Config/testthat/parallel: true`

Could we fall back to serial testing if callr is not installed?

gaborcsardi:
> Could we fall back to serial testing if callr is not installed?

Maybe, but they are not interchangeable.

In general it would be better to have a solution that does not break people's current workflows. Which is possible, I think. E.g. we can change processx etc. to install cleanly in webR, and fail for the function calls.

If you don't want to deal with this now, I guess you can use your testthat branch that does not have callr?

lionel-:
yes there is no hurry, webr-repo currently uses this branch.

gaborcsardi:
Btw. there is this thing: https://browsix.org/ that handles `fork()`, etc. on top of web workers, so I guess at some point callr etc. will work in webassembly.

lionel-:
yup but browsix is basically unmaintained and unused. The WASI folks seem interested in process launching and IPC the long term but there is currently not much interest in the wasm community, so we're probably years away of a working solution. It seems the best approach would be to specially handle the case of calling back to R by instantiating a new emscripten module in a web worker, but it's not clear how we can make them share the same virtual filesystem.

gaborcsardi:
Btw. if you set the `TESTTHAT_PARALLEL` env var to `FALSE` then you probably don't need a custom testthat branch. You could set it at startup I guess?

lionel-:
This branch fixes issues at install time rather than runtime: https://github.com/lionel-/webr-repo/blob/7f0c7bb7337b9071bf3952cb8e779b52b91f201a/install.R#L6

PR #1608, opened 2022-04-21
Snapshot hint override

### LINKED ISSUE #1606: local_reproducible_context() is active when the snapshot advice is printed
Which means that the advice will have no hyperlinks.

hadley:
Fixes #1606


gaborcsardi:
I think the test needs to depend on the new cli version, to fix the CI failures, which run with old cli.

hadley:
Oops, I forgot to check that it had actually been accepted by CRAN.

gaborcsardi:
Unrelated to this PR, but seems like the variants are not working properly here: https://github.com/r-lib/testthat/runs/6147007072?check_suite_focus=true#step:7:180

Maybe when our CI `release` switches to R-release really that will fix this.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # testthat (development version)
 
+* Fix bug revealed by fix to another bug in cli (#1606).
+
```
gaborcsardi, line 4: I am not sure if this is very useful to the end user, maybe we don't need to mention it at all, as nobody has really seen this bug?

tests/testthat/_snaps/snapshot.md
```diff
@@ -243,11 +243,11 @@
     Code
       cat(snapshot_accept_hint("_default", "bar.R"))
     Output
-      * Run `snapshot_accept('bar.R')` to accept the change
-      * Run `snapshot_review('bar.R')` to interactively review the change
+      * Run `]8;;rstudio:run:testthat::snapshot_accept('bar.R')snapshot_accept('bar.R')]8;;` to accept the change
+      * Run `]8;;rstudio:run:testthat::snapshot_review('bar.R')snapshot_review('bar.R')]8;;` to interactively review the change
```
gaborcsardi, line 247: cli is not using the ideal ansi sequence for links currently, because rstudio does not support that yet. But once we fix this in rstudio, cli will switch to that as well, so these snapshots will change. fyi.

gaborcsardi (approved):

NEWS.md
hadley: Yeah, not sure why I wrote that. I'll just add something generic about cli.

tests/testthat/_snaps/snapshot.md
hadley: Ok, sounds good.

Issue #1548, opened 2022-01-10
Reconsider expect_snapshot_error()


hadley:
Based on discussion in #1471.  In particular do we want to show the code, perhaps as a part of 3.1 edition?

hadley:
I think we just want to steer people away from it generally.

PR #1616, opened 2022-05-02
Explain the deal w/ expect_snapshot_error() vs. expect_snapshot(error…

jennybc:
… = TRUE)

Related to getting stuff out of my head and/or out of the (revised) testing chapter in R Packages.

I've needed to be reminded of this multiple times now.


jennybc:
This appears to be related to

https://github.com/r-lib/testthat/issues/1471#issuecomment-1005372257  and #1548

Maybe this PR is ?wrong? depending on how that goes.

jennybc:
Reading the linked issues reminds me:

the other difference and reason one might prefer `expect_snapshot_error()` is if you really want that explicit `class` argument.

Issue #1621, opened 2022-05-10
Add markdown reporter


Bisaloo:
It would be very nice to have a markdown reporter, so that test results can be displayed directly using the new GitHub Actions [job summaries feature](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) without having to dig into the logs.

If this seems like a desirable feature for testthat, I could try to have a look at implementation but can't guarantee a PR as it might be beyond my abilities.

hadley:
We also need to consider who will be responsible for turning this reporter on — should testthat automatically use this reporter if `GITHUB_STEP_SUMMARY` is set, or should `r-lib/actions` be responsible for explicitly enabling it.

hadley:
See discussion in #1635: due to GitHub's implementation of this feature, it's not that useful, and it doesn't seem like we can make it more useful on our side.

PR #1635, opened 2022-06-01
Markdown summary on GHA

gaborcsardi:
Cf:https://github.com/r-lib/testthat/pull/1635#issuecomment-1143625018

- [x] maybe remove of the emoji from the table head?
- [x] remove (some?) of the emoji from the table body?
- [x] better printing of timings
- [ ] check if we can link tests to the source code
- [ ] distinguish between headings of multiple archs
- [x] "context" should be "file" 
- [ ] check if we can include the skip reason and the warning text in the table.
- [x] make "Test Details" the text of the dropdown.
- [ ] consider removing the timings from the details.
- [ ] consider also showing the slowest tests


gaborcsardi:
@hadley Thanks for the feedback, I should have said that this is a first draft. :) Yeah, most of these are possible, many of them I was planning on doing. Others are more challenging, but definitely worth taking a better look. Details soon.

hadley:
Fixes #1621

hadley:
I'm not sure how much further to push this — summaries are considerable less useful that I had expected since (as far as I can tell) you need two clicks to get to them from this page and there's no easy way to know that they exist. I think if we were to do more, it would make sense to make this part of the `CheckReporter` and write hand-in with the console outputs in #2183.

gaborcsardi:
Yeah, I forgot if it is was possible to have a single summary for all runs, probably not? If not, then I don't really want this, it is too much to have a separate summary for each job, even if we improve the output.

Maybe we could have a job that summarizes the tests of all jobs? But then this does not really belong in testthat.

Issue #1617, opened 2022-05-02
Favor `.md` over `.json` in `?expect_snapshot()`


jennybc:
In the [workflow section](https://testthat.r-lib.org/reference/expect_snapshot.html#workflow) of the help for `expect_snapshot()`, it strikes me as odd that `.json` is emphasized, whereas my practical experience suggests `.md` is much more common.

It feels like `.json` will only come up for certain uses of `expect_snapshot_value()`.

PR #1640, opened 2022-06-03
update snapshot docs

dpprdan:
Some minor fixes to the snapshot.R docs:

- snapshots are saved as Markdown files, not JSON (right?)
- fix paths to snapshot folder
- typo


R/snapshot.R
```diff
@@ -22,19 +22,19 @@
 #'
 #' @section Workflow:
 #' The first time that you run a snapshot expectation it will run `x`,
-#' capture the results, and record in `tests/testthat/snap/{test}.json`.
+#' capture the results, and record them in `tests/testthat/_snap/{test}.md`.
```
IndrajeetPatil, line 25: Shouldn't this actually be `_snaps`? 
E.g., [here](https://github.com/r-lib/testthat/tree/main/tests/testthat/_snaps).

dpprdan: 😆 good catch! Thank you @IndrajeetPatil

hadley:
Fixes #1617

hadley:
Thank you!

PR #1647, opened 2022-06-22
set local state for detecting default edition

MichaelChirico:
If the test suite is run in an environment where `TESTTHAT_EDITION=3` is set, this test will fail; reset that locally to ensure success


hadley:
Thanks!

PR #1655, opened 2022-07-08
Broken link to testing chapter in R packages book

ChrisBeeley:
There is a broken link in the README, it looks like this: https://r-pkgs.org/tests.html

There doesn't seem to be a link to the chapter, but this goes to test basics https://r-pkgs.org/testing-basics.html


hadley:
Thank you!

PR #1659, opened 2022-07-20
Use file.path() for robustness

MichaelChirico:
Obviously it hasn't caused any issues so far, but still probably best to use the tailor-made function


hadley:
Thanks

PR #1662, opened 2022-07-30
Minimize output noise when R has no NLS support

### LINKED ISSUE #1566: Warning: 'Changing language has no effect when R installed without NLS'
I'm now seeing warnings in tests where I wasn't before:
```
Warning (test-types.R:12:3): TensorShapes are not converted to lists
Changing language has no effect when R installed without NLS
Backtrace:
 1. testthat::expect_true(y[2] == shape(10))
      at test-types.R:12:2
 2. testthat:::expect_waldo_constant(act, TRUE, info = info)
 3. testthat:::waldo_compare(...)
 4. reporter$local_user_output()
 5. testthat::local_reproducible_output(...)
 6. withr::local_language(lang, .local_envir = .env)
```
The test is calling `expect_true()`, which should behave the same regardless of language, I think.

dgkf:
Closes #1566

A pretty minor change to implement the suggested changes in #1566.

I made one additional design choice to emit a once-per-session warning when `local_reproducible_output` is run without `NLS`, but I'm not sure exactly what the best behavior would be. I could see the argument either way - either as a good reminder that the tests may still fail on another system, or as an unfortunate state-dependent test result. I would have no problem removing it if reproducibility of test result warnings is the more important behavior. 

When run without `NLS` capability, the first test to check against `local_reproducible_output` will emit a warning:

```
───────────────────────────────────────────────────────────
Warning (test-base.R:1:1): (code run outside of `test_that()`)
Local reproducible output can't be checked fully because R is installed without NLS
This warning is displayed once per session.
Backtrace:
 1. testthat::test_that(...)
      at test-base.R:1:0
 2. testthat::local_test_context()
      at testthat/R/test-that.R:40:2
 3. testthat::local_reproducible_output(.env = .env)
      at testthat/R/local.R:70:4
───────────────────────────────────────────────────────────
```


dgkf:
This issue has been annoying me for a while, so when I saw that the suggested changes were pretty minimal I just went ahead and fixed it locally. It was just a little extra work to prep it to share. Happy to make whatever changes are needed to get this issue resolved.

R/local.R
```diff
@@ -117,7 +117,18 @@ local_reproducible_output <- function(width = 80,
     .local_envir = .env,
   )
   withr::local_envvar(RSTUDIO = NA, .local_envir = .env)
-  withr::local_language(lang, .local_envir = .env)
+
+  if (isTRUE(capabilities("NLS"))) {
+    withr::local_language(lang, .local_envir = .env)
+  } else {
+    warning_msg <- paste0(
```
hadley, line 124: I think you can just omit this — it's not going to affect much code, so in most cases it will be a false positive.

hadley (commented):
Thanks for working on this! As well as the suggestion below, can you please also add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

R/local.R
dgkf: Warning and related tests removed and an entry was added to `NEWS.md`.

hadley:
Thanks! I ended up removing the test because I don't think it actually tests the case we really care about, and hence just gives a false sense of security.

PR #1664, opened 2022-08-01
Improve check for $srcref member in expectation

gaborcsardi:
Otherwise if the original error object has a $srcref member that is
not of the srcref class then testthat fails when reporting the error.


hadley:
Thanks!

PR #1682, opened 2022-09-20
Snapshot documentation improvements

### LINKED ISSUE #1548: Reconsider expect_snapshot_error()

### LINKED ISSUE #1629: Link snapshot tests instead of `verify_output()` from `expect_output()` man page
https://github.com/r-lib/testthat/blob/2f48ca6c86fe205cd9283cd6cc8eb0e2a07031b5/R/expect-output.R#L4

### LINKED ISSUE #1649: Clarify expext_snapshot_value() vs expect_snapshot_output()?
- On https://testthat.r-lib.org/reference/expect_snapshot.html they could be explained together (rather than having two items between them).
- I need to look up docs because "output" is what one can call the "return value" so maybe there could be a warning for users (in case my thinking isn't too weird).

hadley:
Fixes #1548. Fixes #1629. Fixes #1649.


R/snapshot.R
```diff
@@ -8,17 +8,9 @@
...
-#' * `expect_snapshot_output()` captures just output printed to the console.
-#' * `expect_snapshot_error()` captures an error message and
-#'   optionally checks its class.
-#' * `expect_snapshot_warning()` captures a warning message and
-#'   optionally checks its class.
-#' * `expect_snapshot_value()` captures the return value.
-#'
-#' (These functions supersede [verify_output()], [expect_known_output()],
-#' [expect_known_value()], and [expect_known_hash()].)
+#' `expect_snapshot()` executes code like you ran it at the console, and
```
jennybc, line 11:
```suggestion
#' `expect_snapshot()` runs code as if you had executed it at the console, and
```

R/snapshot.R
```diff
@@ -273,6 +279,12 @@ expect_snapshot_condition <- function(base_class, x, class, cran = FALSE, varian
   )
 }
 
+#' Snapshot testing for values
+#'
+#' Captures the result of function, flexibly serializing into into a text
```
jennybc, line 284:
```suggestion
#' Captures the result of function, flexibly serializing it into a text
```

R/snapshot.R
```diff
@@ -284,11 +296,11 @@ expect_snapshot_condition <- function(base_class, x, class, cran = FALSE, varian
 #'   * `serialize()` produces a binary serialization of the object using
 #'     [serialize()]. This is all but guaranteed to work for any R object,
 #'     but produces a completely opaque serialization.
-#' @param ... For `expect_snapshot_value()` only, passed on to
-#'   [waldo::compare()] so you can control the details of the comparison.
-#' @export
+#' @param ... Passed on to [waldo::compare()] to you can control the details of
```
jennybc, line 299:
```suggestion
#' @param ... Passed on to [waldo::compare()] so you can control the details of
```

jennybc (approved):
It probably makes sense to make a few changes to the snapshotting vignette at the same time. Specifically the sections below here https://github.com/r-lib/testthat/blob/8b4c265aa5eb7d7eb03d708553924477bc27ff01/vignettes/snapshotting.Rmd#L213

hadley:
Oh yes, the vignette!

PR #1683, opened 2022-09-20
Clickable advice for setting max number of failures

### LINKED ISSUE #1628: Advice to use `Sys.setenv('TESTTHAT_MAX_FAILS' = Inf)` should be clickable
i.e. here https://github.com/r-lib/testthat/blob/0f24eae7bd2cd7d0fbe5a5492636731d66d0dd26/R/reporter-progress.R#L211

hadley:
Fixes #1628


NAMESPACE
```diff
@@ -155,6 +155,7 @@ export(not)
 export(prints_text)
 export(quasi_label)
 export(run_cpp_tests)
+export(set_max_fails)
```
hadley, line 158: Not sure about this name

R/reporter-progress.R
```diff
@@ -540,3 +540,16 @@ skip_bullets <- function(skips) {
   tbl <- table(skips)
   paste0(cli::symbol$bullet, " ", names(tbl), " (", tbl, ")")
 }
+
+
+#' Set maximum number of test failures before aborting the run
```
DavisVaughan, line 545:
```suggestion
#' Set maximum number of test failures allowed before aborting the run
```

DavisVaughan (approved):
Name sounds fine to me, it is in the same vein as `testthat::set_reporter()`.

I don't _think_ we need `get_max_fails()`, but wanted to mention it in case you think we do

R/reporter-zzz.R
```diff
@@ -59,7 +59,7 @@ with_reporter <- function(reporter, code, start_end_reporter = TRUE) {
 
 stop_reporter <- function(message) {
   signal(message, "testthat_abort_reporter")
```
DavisVaughan, line 61: Is signaling a `message` with cli inline markup problematic at all?

hadley: I don't think anything actually uses the message, so might be safer to set to `""`.

DavisVaughan: @lionel- could `signal()` allow `message` to be missing, i.e. `signal(class = "testthat_abort_reporter")`? I also just had a case in dplyr where I just wanted to do the signaling but didn't care about the message

lionel-: I think it'd be fine to take `""` by default.

hadley: Actually, I can just combine the `cli_abort()` and `signal()` — I think this must be a remnant from when the code used `stop()` and it was harder to class the error.

PR #1690, opened 2022-09-22
Implement expect_without_warnings() and friends

### LINKED ISSUE #1679: `expect_without_warning()`
Or similar

hadley:
Fixes #1679

I've kept the interface deliberately very minimum; let me know if you think there's anything I'm missing.


NAMESPACE
```diff
@@ -124,6 +124,10 @@ export(expect_type)
 export(expect_vector)
 export(expect_visible)
 export(expect_warning)
+export(expect_without_condition)
```
lionel-, line 127: I think these names are a bit too long. They will dominate the line width, making it harder to read the actual logic.

R/expect-no-condition.R
```diff
@@ -0,0 +1,113 @@
+#' Does code running without error, warning, message, or other condition?
```
lionel-, line 1:
```suggestion
#' Does code run without error, warning, message, or other condition?
```

R/expect-no-condition.R
```diff
@@ -0,0 +1,113 @@
...
+#'
+#' These expectations assert the opposite of [expect_error()],
+#' `expect_warning()`, `expect_message()`, and `expect_condition()`, namely
+#' that the code runs without error, warning, or message respectively.
+#'
+#' @inheritParams expect_error
+#' @param message,class The default usage with `message = NULL, clas = NULL`
```
lionel-, line 8: This is a hard one. On the one hand, we want users to test for specific warnings. On the other hand, we don't want them to test for messages that are most probably generated in other packages.

This makes me think that `expect_warning(x, NA)` should be replaced by snapshots?

R/expect-no-condition.R
```diff
@@ -0,0 +1,113 @@
...
+#' expect_without_warning(foo(), message = "bananas")
+#'
+#' # warning does match so causes a failure:
+#' try(expect_without_warning(foo(), message = "problem"))
+expect_without_error <- function(object,
+
+                                 ...,
+                                 message = NULL,
+                                 class = NULL
+                                 ) {
```
lionel-, line 32: Here and below.

```suggestion
                                 class = NULL) {
```

lionel- (commented):
I'm having doubts about this. It looks like we're encouraging testing strings generated in other packages. Should we encourage snapshots instead?

R/expect-no-condition.R
```diff
@@ -0,0 +1,113 @@
...
+#' These expectations assert the opposite of [expect_error()],
+#' `expect_warning()`, `expect_message()`, and `expect_condition()`, namely
+#' that the code runs without error, warning, or message respectively.
+#'
+#' @inheritParams expect_error
+#' @param message,class The default usage with `message = NULL, clas = NULL`
+#'   will match any error/warning/message/condition. This is a broad assertion
+#'   so you'll typically want to make it narrower by either supplying `message`
+#'   (a regular expression that must match the message of the condition) or
+#'   `class` (which the condition must inherit from).
```
jennybc, line 12:
```suggestion
#'   will match any error/warning/message/condition. This is a broad assertion,
#'   so sometimes you may want to be more specific about the condition you are
#'   hoping to **not** see, i.e. the condition that motivates the expectation.
#'   Similar to `expect_error()` and friends, you can specify the `message` (a
#'   regular expression that the message of the condition must match) and/or the
#'   `class` (a class the condition must inherit from).
```

My main comment in this area is that the significance of `message` and `class` are a bit mind-bend-y here and that requires careful wording.

In the positive case (so: `expect_error()`), you're saying what you *expect* to see.

In this negative case, you're describing the condition you're worried about, i.e. the one that motivated you to write this expectation.

I don't think the current wording is wrong *per se*, but I've proposed an alternative

I also think one is much less likely to specify `message` or `class` for the `expect_without_*()` expectation and that strikes me as totally OK.

jennybc (approved):
I'm happy with this. I made a few inline proposals and comments. I also test drove it in a couple of packages where I had the sort of DIY test helper that this is meant to replace and all seems well (not a big surprise).

R/expect-no-condition.R
hadley: Moved to slack for quicker iteration.

R/expect-no-condition.R
hadley: I've tweaked the text to reinforce that you should be using `message` for your own conditions.

hadley:
I marked these as experimental, which makes me feel better about including them in a patch release.

NAMESPACE
lionel-: Might be too literary but:

`expect_sans_warnings(foo(), class = "foo")`?

hadley: Or too French 😛

lionel-: > `expect_no_warnings(foo(), class = "abc")` doesn't make as much sense to me.

I don't love the "no" particle but I like the concision.
Does it make sense to read it as "expect no warnings of class 'abc'"?

hadley: Ok, you've convinced me

Issue #1694, opened 2022-09-23
Compilation error


jennybc:
I can't compile or check testthat at the moment. I see:

```
  clang++ -I"/Library/Frameworks/R.framework/Versions/4.2/Resources/include" -DNDEBUG -I../inst/include -DCOMPILING_TESTTHAT  -I/usr/local/include   -fPIC  -g -O0 -Wall -pedantic -mtune=native -c test-runner.cpp -o test-runner.o
   In file included from test-runner.cpp:7:
   In file included from ../inst/include/testthat.h:1:
   In file included from ../inst/include/testthat/testthat.h:72:
   ../inst/include/testthat/vendor/catch.h:7187:59: error: use of undeclared identifier 'rng'
               random_shuffle( vector.begin(), vector.end(), rng );
                                                             ^
   1 error generated.
```

@kevinushey I think you were in this neighborhood recently (#1688). Is there more information about my system that would be helpful? macOS Big Sur 11.7 with up-to-date Xcode command line tools

PR #1695, opened 2022-09-23
don't pass 'rng' to random_shuffle

kevinushey:
Closes https://github.com/r-lib/testthat/issues/1694.


jennybc (approved):
LGTM and fixes my compilation problem

PR #1704, opened 2022-10-07
Chain source errors

hadley:
To make it easier to understand where problems are coming from


R/source.R
```diff
@@ -36,7 +36,15 @@ source_file <- function(path, env = test_env(), chdir = TRUE,
   if (wrap) {
     invisible(test_code(NULL, exprs, env))
   } else {
-    invisible(eval(exprs, env))
+    tryCatch(
```
lionel-, line 39: To allow recover and deep backtraces:

```suggestion
    withCallingHandlers(
```

R/source.R
```diff
@@ -36,7 +36,15 @@ source_file <- function(path, env = test_env(), chdir = TRUE,
   if (wrap) {
     invisible(test_code(NULL, exprs, env))
   } else {
-    invisible(eval(exprs, env))
+    withCallingHandlers(
+      invisible(eval(exprs, env)),
+      error = function(err) {
+        abort(
+          paste0("Failed to source ", encodeString(path, quote = '"')),
```
lionel-, line 43:
```suggestion
          paste0("Can't source ", encodeString(path, quote = '"')),
```

hadley: I wonder if I like "failed" when the problem is not directly related to the argument supplied.

But maybe this should actually just be `In: `

lionel-: oh yeah looks like a good candidate for an info "In" bullet

PR #1717, opened 2022-12-06
Remap `sprintf` to `Rprintf`

### LINKED ISSUE #1715: Compiled code warning on CRAN R-devel from `test-runner`
Hi testthat developers,

we just got an email from the CRAN team that warns as follows about https://github.com/openpharma/mmrm/blob/main/src/test-runner.cpp which is merely the standard file from `testthat` for C++ unit tests:
image
(see https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/mmrm-00check.html for the full log).
This is on R-devel (starting from 2022-12-03 r83400), but it seems only on Linux OS (Debian, Fedora) and not Windows.

I would guess this is observed in other R-packages using the Catch framework via `testthat` as well. Is there something we can help to fix this in `testthat`? (and as you would guess we only have 13 days now to fix this...)

Thanks,
best regards
Daniel

lionel-:
Fix for https://github.com/wch/r-source/commit/a0dfa3df365e3ee2b4496fc2374963ff20da7043 which causes this warning on our CRAN checks as well as in packages that include `inst/testthat.h`:

```
Version: 3.1.5
Check: compiled code
Result: WARN
    File 'testthat/libs/testthat.so':
     Found 'sprintf', possibly from 'sprintf' (C)
     Object: 'test-runner.o'
    
    Compiled code should not call entry points which might terminate R nor
    write to stdout/stderr instead of to the console, nor use Fortran I/O
    nor system RNGs nor [v]sprintf.
    
    See 'Writing portable packages' in the 'Writing R Extensions' manual.
Flavors: [r-devel-linux-x86_64-debian-clang](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-clang/testthat-00check.html), [r-devel-linux-x86_64-fedora-clang](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/testthat-00check.html)
```

Bumps version to 3.1.6 for a patch release.
Closes #1715.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: testthat
 Title: Unit Testing for R
-Version: 3.1.5.9000
+Version: 3.1.6
 Authors@R: c(
     person("Hadley", "Wickham", , "hadley@rstudio.com", role = c("aut", "cre")),
     person("RStudio", role = c("cph", "fnd")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,7 @@
-# testthat (development version)
+# testthat 3.1.6
+
+* Fixed CRAN warning about usage of `sprintf()` in `Catch.h` (#1715).
+
 
 # testthat 3.1.5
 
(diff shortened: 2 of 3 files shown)


hadley:
Duplicate of #1716 😄

lionel-:
This patch wasn't good anyway, I got printf and sprintf confused!

PR #1735, opened 2023-02-06
Include the "Manage test reporting" article in the Reporters block.

jonthegeek:
Evidently "Reporter" doesn't qualify as `ends_with("Reporter")` when pkgdown builds the site. This leads to confusing documentation that doesn't describe what a Reporter is in general.

This made the [Reporters](https://testthat.r-lib.org/reference/index.html#reporters) section of the reference difficult to follow at first.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -62,6 +62,7 @@ reference:
 
 - title: Reporters
   contents:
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1739, opened 2023-02-12
Re-enter the mocking game

hadley:


R/mock2.R
```diff
@@ -0,0 +1,96 @@
+#' Mocking tools
+#'
+#' @description
+#' `r lifecycle::badge("experimental")`
+#'
+#' These functions represent a second attempt at bringing mocking to testthat,
+#' incorporating what we've learned from mockr and mockery There are
+#' two styles of mocking:
+#'
+#' * `local_mocked_bindings()` and `local_mocked_bindings()` work by temporarily
```
gaborcsardi, line 10:
```suggestion
#' * `with_mocked_bindings()` and `local_mocked_bindings()` work by temporarily
```

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+#'   changing variable bindings in the package namespace. It affects the
+#'   operation of every function in the current package, but only the current
+#'   package.
+#'
+#' * `with_mocked_env()` modifies a function, inserting a new environment in the
```
gaborcsardi, line 15: Is there a reason why we don't have `local_mocked_env()`?

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+    cli::cli_abort(
+      "All elements of {.arg ...} must be named.",
+      call = error_call
+    )
+  }
+
+  is_fun <- map_lgl(x, is.function)
+  if (!any(is_fun)) {
+    cli::cli_abort(
+      "All emeents of {.arg ...} must be functions.",
```
gaborcsardi, line 92:
```suggestion
      "All elements of {.arg ...} must be functions.",
```

gaborcsardi (commented):
Cool!

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+#' @param f A function.
+#' @param ... Name-value pairs providing functions to mock.
+with_mocked_env <- function(f, ...) {
+  if (!is.function(f)) {
+    cli::cli_abort("{.arg f} must be a function, not {.obj_type_friendly {f}}.")
+  }
+  bindings <- list2(...)
+  check_bindings(bindings)
+
+  new_env <- child_env(get_env(f), !!!bindings)
```
lionel-, line 30:
```suggestion
  new_env <- env(get_env(f), !!!bindings)
```

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+#'
+#' * `with_mocked_env()` modifies a function, inserting a new environment in the
+#'   chain of parents with the specified bindings. It can mock any function,
+#'   but because it creates a new function, it doesn't affect any functions
+#'   that use that function.
+#'
+#' @export
+#' @param f A function.
+#' @param ... Name-value pairs providing functions to mock.
+with_mocked_env <- function(f, ...) {
```
lionel-, line 23: Is this variant going to be useful? It seems it might be confusing / error prone for users. Even with the note in the documentation, I can see users being confused and still expect that if the mocked function calls another function from the same namespace, then the bindings will be transitively locked, but that won't the case.

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+with_mocked_bindings <- function(code, ...) {
+  quo <- enquo(code)
+  local_mocked_bindings(...)
+  code
+}
+
+#' @export
+#' @param .env Environment that defines effect scope. For expert use only.
+#' @rdname with_mocked_env
+local_mocked_bindings <- function(..., .env = caller_env()) {
```
lionel-, line 46: Take `ns = ns_env(dev_package())` argument?

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+  bindings <- list2(...)
+  check_bindings(bindings)
+
+  ns_env <- ns_env(dev_package())
+
+  # Unlock bindings, if needed
+  nms <- names(bindings)
+  locked <- env_binding_are_locked(ns_env, nms)
+  env_binding_unlock(ns_env, nms[locked])
```
lionel-, line 55: `env_binding_unlock()` returns the old state, so the `env_binding_are_locked()` call above can be simplified.

R/mock2.R
```diff
@@ -0,0 +1,96 @@
...
+  withr::defer(env_binding_lock(ns_env, nms[locked]), envir = .env)
+
+  # Set values
+  old_ns <- env_bind(ns_env, !!!bindings)
+  withr::defer(env_bind(ns_env, !!!old_ns), envir = .env)
```
lionel-, line 60: This can be simplified with `local_bindings()`.

R/mock2.R
lionel-: Probably this variant should be called `set_mocked_env()` since it's not really a `with_` function (hence why it doesn't have a `local_` variant).

lionel- (commented):
> incorporating what we've learned from mockr and mockery

So what did we learn? I'm curious as I have never used these packages.

R/mock2.R
hadley: I've made it `.package = NULL` since I think that's a bit easier to use.

R/mock2.R
hadley: I originally thought this would cover some cases that `local_mocked_bindings()` would not, but now I don't see where it's useful, unless you want to mock a closure generated by the package, which feels pretty esoteric. Maybe it's fine to drop it?

hadley:
What we learned:

* We now have the knowledge and tools to radically simplify the implementation. Compare `with_mocked_bindings()` to `with_mock()`.
* Now that I've more carefully inspected the implementation of mockery and mockr, I think it actually uses a new technique, and this now seems like "obvious" technique for it to use.
* It's really important for the mocking tool to work in both batch test mode and interactively. mockery

hadley:
Note that this doesn't appear to work to mock S3 methods.

```R
# In package
mockable_generic <- function(x) {
  UseMethod("mockable_generic")
}
#' @export
mockable_generic.integer <- function(x) {
  x + 1
}

# In test

test_that("local_mocked_bindings affects S3 methods", {
  local_mocked_bindings(mockable_generic.integer = function(x) 100)
  expect_equal(mockable_generic(1L), 100)
})
```

I tried also binding in the S3 methods table:

```R
  method_env <- ns_env[[".__S3MethodsTable__."]]
  methods <- names(bindings) %in% names(method_env)
  local_bindings(!!!bindings[methods], .env = method_env, .frame = .env)
  bindings <- bindings[!methods]
```

But while that works when called interactively, it doesn't work in tests.

lionel-:
For S3 methods it is usually sufficient to bind them in the global environment.

hadley:
@lionel- yeah, that only works for new methods, I think.

But maybe my attempt didn't work because of something to do with the way that testthat sets up environments. It certainly looks like pkgload modifies the same place when unloading packages: https://github.com/r-lib/pkgload/blob/main/R/unload.R#L157-L164. (However, that does suggest that, in general, I need to look in the namespace of the generic, not just the namespace of the current package.)

hadley:
Oooooh, I wonder if it's because `test_env()` does `env_clone(asNamespace(package))`.

PR #1745, opened 2023-03-01
Improve expect_setequal() feedback

### LINKED ISSUE #1657: `expect_setequal()` should make more of an effort to display actual values

hadley:
Fixes #1657

### DIFF
--- a/R/expect-setequal.R
+++ b/R/expect-setequal.R
@@ -36,11 +36,11 @@ expect_setequal <- function(object, expected) {
 
   if (any(exp_miss) || any(act_miss)) {
     fail(paste0(
-      act$lab, " and ", exp$lab, " don't have the same values\n",
+      act$lab, " (`actual`) and ", exp$lab, " (`expected`) don't have the same values.\n",
       if (any(exp_miss))
-        paste0("* Missing from expected: ",  strings(act$val[act_miss]), "\n"),
+        paste0("* Only in `actual`:   ",  values(act$val[act_miss]), "\n"),
       if (any(act_miss))
-        paste0("* Missing from actual: ",  strings(exp$val[exp_miss]), "\n")
+        paste0("* Only in `expected`: ",  values(exp$val[exp_miss]), "\n")
     ))
   } else {
     succeed()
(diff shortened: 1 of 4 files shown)


R/expect-setequal.R
```diff
@@ -32,41 +32,37 @@ expect_setequal <- function(object, expected) {
...
-  }
 
-  if (!any(exp_miss) && !any(act_miss)) {
+  if (any(exp_miss) || any(act_miss)) {
+    fail(paste0(
+      act$lab, " and ", exp$lab, " don't have the same values\n",
+      if (any(exp_miss))
+        paste0("* Missing from expected: ",  strings(act$val[act_miss]), "\n"),
+      if (any(act_miss))
+        paste0("* Missing from actual: ",  strings(exp$val[exp_miss]), "\n")
```
DavisVaughan, line 43: How would you feel about framing this as `Only in expected:` instead of `Missing from expected:`?

I have trouble reading this example:

```r
x <- c("a", "b")
y <- c("b", "c")

expect_snapshot_failure(expect_setequal(x, y))
#> `x` and `y` don't have the same values
#> * Missing from expected: "a"
#> * Missing from actual: "c"
```

and I think it is because my brain is juggling too many things. I have to realize that `y` is "expected" but then look over at `x` to see that it did indeed have `"a"` which isn't in `y`.

This makes more sense to me

```r
x <- c("a", "b")
y <- c("b", "c")

expect_snapshot_failure(expect_setequal(x, y))
#> `x` (`actual`) and `y` (`expected`) don't have the same values
#> * Only in `actual`: "a"
#> * Only in `expected`: "c"
```

I also like seeing ``` `x` (`actual`) ``` because I can't ever remember which is which. It is also what `expect_equal()` and `expect_identical()` seem to do

```r
> testthat::expect_identical(1, 1.5)
Error: 1 (`actual`) not identical to 1.5 (`expected`).

  `actual`: 1.0
`expected`: 1.5
> testthat::expect_equal(1, 1.5)
Error: 1 (`actual`) not equal to 1.5 (`expected`).

  `actual`: 1.0
`expected`: 1.5
```

R/expect-setequal.R
```diff
@@ -32,41 +32,37 @@ expect_setequal <- function(object, expected) {
...
   invisible(act$val)
 }
 
-is_vector <- function(x) is.list(x) || (is.atomic(x) && !is.null(x))
-
-locations <- function(i) {
-  loc <- which(i)
-  if (length(loc) == 1) {
-    return(loc)
+strings <- function(x) {
```
DavisVaughan, line 52: Can't you use `expect_setequal()` on more than strings though? How does this look on numeric values?

R/expect-setequal.R
hadley: I like it.

R/expect-setequal.R
hadley: Oops, yes.

PR #1748, opened 2023-03-04
Bind in imports

hadley:


lionel- (approved):
Looks great!

This could be extracted in a function, and then we could loop over the imports of all dependencies of the current package. Then we have full mocking across the whole session. This will be useful for packages that depend on each other in an integrated stack.

hadley:
What are you thinking would be extracted into a function?

hadley:
If I understand what you're saying, I think you can already do that with (e.g.)

```R
local_mocked_bindings(sym = function(...) "x", .package = "rlang")
```

lionel-:
oh right. Well that does look good enough and also maybe safer to do it explicitly.

hadley:
Yeah, for now, I wanted to make this explicit, because it somehow feels wrong to make it easy to change code owned by others.

PR #1755, opened 2023-03-21
Clarify argument names in `test_files_setup_state()`

lionel-:
I got confused while reading `test_files_setup_state()` because it takes the test environment as `env` and the exit frame as `.env`. This PR clarifies the naming.

### DIFF
--- a/R/parallel.R
+++ b/R/parallel.R
@@ -227,7 +227,7 @@ queue_process_setup <- function(test_package, test_dir, load_helpers, load_packa
     test_package = test_package,
     load_helpers = load_helpers,
     env = env,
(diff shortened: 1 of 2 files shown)


R/parallel.R
```diff
@@ -227,7 +227,7 @@ queue_process_setup <- function(test_package, test_dir, load_helpers, load_packa
     test_package = test_package,
     load_helpers = load_helpers,
     env = env,
-    .env = .GlobalEnv
+    exit_frame = .GlobalEnv
```
hadley, line 230: Why `exit_frame` and not just `frame`?

Issue #1562, opened 2022-02-01
Is calling `test_path()` at a global level within a `helper-*.R` file not supported?


DavisVaughan:
In the censored package we had a `helper-` file that loaded some helper data objects with `test_path()` and `readRDS()`. Like this:
https://github.com/tidymodels/censored/pull/145/files#diff-86973057dbc44244492fcd424b56710d5bccdab7c751ca7d469b4c744f85606aL1

It turns out that this doesn't work. You can easily reproduce with this repo:
https://github.com/DavisVaughan/testthathelperpath

Just do `load_all()`, which errors with:

```r
> devtools::load_all(".")
ℹ Loading testthathelperpath
Error in `test_path()`:
! Can't find `tests/testthat/` in current directory.
Run `rlang::last_error()` to see where the error occurred.

> rlang::last_trace()
<error/rlang_error>
Error in `test_path()`:
! Can't find `tests/testthat/` in current directory.
Backtrace:
     ▆
  1. └─devtools::load_all(".")
  2.   └─pkgload::load_all(...)
  3.     ├─withr withr_with_envvar(...)
  4.     │ └─base::force(code)
  5.     └─testthat testthat_source_test_helpers(find_test_dir(path), env = pkg_env(package))
  6.       └─testthat::source_dir(path, "^helper.*\\.[rR]$", env = env, wrap = FALSE)
  7.         └─base::lapply(files, source_file, env = env, chdir = chdir, wrap = wrap)
  8.           └─testthat FUN(X[[i]], ...)
  9.             └─base::eval(exprs, env)
 10.               └─base::eval(exprs, env)
 11.                 ├─base::readRDS(test_path("data", "data.rds")) at tests/testthat/helper-load.R:2:0
 12.                 └─testthat::test_path("data", "data.rds")
 13.                   └─rlang::abort("Can't find `tests/testthat/` in current directory.")
```

There are lots of ways to work around this, but I was a bit surprised that this didn't work so we figured it was best to open an issue

gaborcsardi:
That might not work, depending on code paths, because e.g. `devtools::test()` sources the helper files before calling testthat, so testthat couldn't have had time to set up the package, paths, etc. yet.

DavisVaughan:
Do you think it is a bug? Maybe a devtools bug? If not, I am happy to close this since we have workarounds

gaborcsardi:
[...]

PR #1756, opened 2023-03-21
improve `test_path()` manual

DanChaltiel:
Clarifies that `test_path()` will not work in helper files.

Fixes https://github.com/r-lib/testthat/issues/1562


hadley:
This isn't quite right, because it only doesn't work at the top-level; it'll work if you wrap it in a function that's called from a test file. And the rest of the existing docs aren't great either. So I think it's just better to rework this function. PR incoming shortly.

PR #1764, opened 2023-03-29
Include discussion of embedded skips

### LINKED ISSUE #1639: Special skipping technique
I'm pulling this out of the material drafted for the testing chapter of R Packages 2e.

It might fit into <https://testthat.r-lib.org/articles/skipping.html>?

---

At the complete other extreme, another practice that can make sense in very specific situations is to build a `skip()` directly into a package function.
Here's an example from pkgdown, in `pkgdown::convert_markdown_to_html()`, which absolutely, positively cannot work if the Pandoc tool is unavailable:

<https://github.com/r-lib/pkgdown/blob/98d5a5c735eb244cb98b2e6bab1d54bb27c0af95/R/markdown.R#L95-L106>

```{r eval = FALSE}
# in r-lib/pkgdown/R/markdown.R
convert_markdown_to_html <- function(in_path, out_path, ...) {
  if (rmarkdown::pandoc_available("2.0")) {
    from <- "markdown+gfm_auto_identifiers-citations+emoji+autolink_bare_uris"
  } else if (rmarkdown::pandoc_available("1.12.3")) {
    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash+header_attributes"
  } else {
    if (is_testing()) {
      testthat::skip("Pandoc not available")
    } else {
      abort("Pandoc not available")
    }
  }
  
  rmarkdown::pandoc_convert(...)
  invisible()
}
```

If Pandoc is not available when `convert_markdown_to_html()` executes, it throws an error *unless* it appears to be part of a test run, in which case the test is skipped.
This is an alternative to implementing a custom skipper, e.g. `skip_if_no_pandoc()`, and inserting it into many of pkgdown's tests.

`testthat::is_testing()` is a small convenience function that checks the value of the `TESTTHAT` environment variable, which will be true when testthat is running tests:

```{r eval = FALSE}
is_testing <- function() {
  identical(Sys.getenv("TESTTHAT"), "true")
}
```

You usually want to avoid a runtime dependency on testthat, i.e. you won't want to use testthat in the code below `R/`.
In that case, you should inline the source of `is_testing()` into your package, in whatever location you are using for any other test helper functions.

*Note: The example above actually appears to still have a runtime dependency on testthat, because of the call to `testthat::skip()`.
But `testthat::skip` is only ever run when you're inside a test, implying that branch operates under the "testing rules", where we generally assume that all Suggested packages are installed.*

Advantages of the embedded `skip()` approach: Feels elegant and concise.

Disadvantages of this approach: Seems to violate "Good production code is well-factored; good test code is obvious", i.e. when you're focused on the tests, the existence of this skip is easy to miss.

hadley:
Fixes #1639

### DIFF
--- a/vignettes/skipping.Rmd
+++ b/vignettes/skipping.Rmd
@@ -62,7 +62,7 @@ You should keep an on eye this when developing interactively to make sure that y
 ## Helpers
 
 If you find yourself using the same `skip_if()`/`skip_if_not()` expression across multiple tests, it's a good idea to create a helper function.
-This function should start with `skip_` and live in a `test/helper-{something}.R`:
+This function should start with `skip_` and live in a `test/helper-{something}.R` file:
 
 ```{r}
 skip_if_dangerous <- function() {
--- a/vignettes/skipping.Rmd
+++ b/vignettes/skipping.Rmd
@@ -51,57 +51,66 @@ All reporters show which tests as skipped.
 As of testthat 3.0.0, ProgressReporter (used interactively) and CheckReporter (used inside of `R CMD check`) also display a summary of skips across all tests.
 It looks something like this:
 
-    ── Skipped tests  ───────────────────────────────────────────────────────
-    ● No token (3)
-    ● On CRAN (1)
+```         
+── Skipped tests  ───────────────────────────────────────────────────────
+● No token (3)
+● On CRAN (1)
+```
 
 You should keep an on eye this when developing interactively to make sure that you're not accidentally skipping the wrong things.
 
 ## Helpers
 
 If you find yourself using the same `skip_if()`/`skip_if_not()` expression across multiple tests, it's a good idea to create a helper function.
-This function should start with `skip_` and live somewhere in your `R/` directory.
-
-```{r}
-skip_if_Tuesday <- function() {
-  if (as.POSIXlt(Sys.Date())$wday != 2) {
-    return(invisible(TRUE))
-  }
-  
-  skip("Not run on Tuesday")
-}
-```
-
-It's important to test your skip helpers because it's easy to miss if you're skipping more often than desired, and the test code is never run.
-This is unlikely to happen locally (since you'll see the skipped tests in the summary), but is quite possible in continuous integration.
-
-For that reason, it's a good idea to add a test that you skip is activated when you expect.
-skips are a special type of condition, so you can test for their presence/absence with `expect_condition()`.
-For example, imagine that you've defined a custom skipper that skips tests whenever an environment variable `DANGER` is set:
+This function should start with `skip_` and live in a `test/helper-{something}.R`:
 
 ```{r}
 skip_if_dangerous <- function() {
-  if (identical(Sys.getenv("DANGER"), "")) {
-    return(invisible(TRUE))
+  if (!identical(Sys.getenv("DANGER"), "")) {
+    skip("Not run in dangerous environments.")
+  } else {
+    invisible()
   }
-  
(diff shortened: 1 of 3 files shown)


vignettes/skipping.Rmd
```diff
@@ -51,57 +51,66 @@ All reporters show which tests as skipped.
...
+● On CRAN (1)
+```
 
 You should keep an on eye this when developing interactively to make sure that you're not accidentally skipping the wrong things.
 
 ## Helpers
 
 If you find yourself using the same `skip_if()`/`skip_if_not()` expression across multiple tests, it's a good idea to create a helper function.
-This function should start with `skip_` and live somewhere in your `R/` directory.
+This function should start with `skip_` and live in a `test/helper-{something}.R`:
```
jennybc, line 65:
```suggestion
This function should start with `skip_` and live in a `test/helper-{something}.R` file:
```

vignettes/skipping.Rmd
```diff
@@ -51,57 +51,66 @@ All reporters show which tests as skipped.
...
 
 ```{r}
-skip_if_Tuesday <- function() {
-  if (as.POSIXlt(Sys.Date())$wday != 2) {
-    return(invisible(TRUE))
+skip_if_dangerous <- function() {
+  if (!identical(Sys.getenv("DANGER"), "")) {
+    skip("Not run in dangerous environments.")
+  } else {
+    invisible()
```
jennybc, line 72: Connecting back to my nitpick comment on the other PR, I find negated conditions sort of mind-bendy. Can we do it the other (more natural) way around?

```suggestion
skip_if_dangerous <- function() {
  if (identical(Sys.getenv("DANGER"), "")) {
    invisible()
  } else {
    skip("Not run in dangerous environments.")
```

Another nice example:

```
skip_if_not_utf8 <- function() {
  skip_if_not(l10n_info()$`UTF-8`)
}
```

vignettes/skipping.Rmd
```diff
@@ -51,57 +51,66 @@ All reporters show which tests as skipped.
...
 ```
 
-Testing `skip_if_Tuesday()` is harder because there's no way to control the skipping from the outside.
-That means you'd need to "mock" its behaviour in a test, using the [mockery](https://github.com/r-lib/mockery) or [mockr](https://krlmlr.github.io/mockr/) packages.
+It might look like the code appears to still have a runtime dependency on testthat, because of the call to `testthat::skip()`.
+But `testthat::skip` is only ever run when you're inside a test, where you can assume that testthat is installed.
+
+We have mixed feelings about this approach.
+On the one hand, it feels elegant and concise, and it absolutely guarantees that you'll never miss a needed skip in one of your tests.
+On the other hand, it mixed code and tests in an unusual way, and it when you're focused on the tests, the existence of a skip in code is easy to miss.
```
jennybc, line 116:
```suggestion
On the other hand, it mixes code and tests in an unusual way, and when you're focused on the tests, it's easy miss the fact that a package function contains a `skip()`.
```

jennybc (approved):
I feel like there's some lesson to be gleaned from this as well:

https://github.com/rstudio/pins-r/issues/640

Now, that is not actually about skipping. But it is about using `is_testing()` in the code below `R/` and maybe this is the best place to caution folks about getting carried away with this idea. It feels related to the downside of embedded skips, i.e. it's easy to puzzle yourself or others in the future.

vignettes/skipping.Rmd
hadley: Let's discuss that more there, because I find the current structure more clear.

PR #1768, opened 2023-04-05
Mention new mocking functions where appropriate

### LINKED ISSUE #1766: Update mocking advice in 3e vignette
This needs an update, in light of `with_mocked_bindings()` and `local_mocked_bindings()`:

https://github.com/r-lib/testthat/blob/HEAD/vignettes/third-edition.Rmd#L92

hadley:
Fixes #1766

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,10 @@
   can more easily run BDD tests interactively (#1587)
 
 * `with_mocked_bindings()` and `local_mocked_bindings()` can now bind in the
-  imports namespace too.
+  imports namespace too. This changes makes them very close in capability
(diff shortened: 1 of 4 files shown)


jennybc (approved):
I looked at `?with_mock` locally and it feels too easy for someone to think they have to use mockr or mockery. I guess I technically understand why that's the 2e advice and why the 3e section just says they are deprecated.

But pragmatically I think I would just open with advice to use `with_mocked_bindings()` and `local_mocked_bindings()`.

If a 2e user is contemplating getting rid of `with_mock()` or `local_mock()` that means the code is no longer at rest and maybe they are a candidate for upgrading to 3e and its features.

PR #1785, opened 2023-05-11
Set minimum pkgload version for is_loading()

MichaelChirico:
Required here:

https://github.com/r-lib/testthat/blob/86b85130beec27d8a73b75049fea585f9ff34dc0/R/test-path.R#L21

{pkgload} ref:

https://github.com/r-lib/pkgload/blob/60a0eae8d464acedc3f6bb42c8cd8cdda26ef057/NEWS.md?plain=1#L74-L75

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Imports:
     lifecycle,
     magrittr,
     methods,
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1789, opened 2023-05-22
CRAN never runs code interactively

hadley:
This makes it easier to run through code line-by-line, particularly when `skip_on_cran()` is nested inside another skip helper.

@DavisVaughan I think we've talked about this before, but I've decided to bite the bullet because it makes line-by-line testing in renv easier.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # testthat (development version)
 
+* `skip_on_cran()` no longer skips (errors) when run interactively.
+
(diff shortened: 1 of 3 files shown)


DavisVaughan (approved):
I think this makes sense

To bring your original thoughts back up:

> the main downside of the current behaviour is that you sometimes want to nest skip_on_cran() inside a helper, and then you can’t run that interactively

PR #1790, opened 2023-05-22
Improve teardown_env() implementation

hadley:
Now captures a frame, rather than relying on withr special behaviour.


lionel- (approved):
Nice!

PR #1792, opened 2023-05-23
Switch to new condition display for all packages

hadley:
@DavisVaughan @lionel- I think it's time. Any objections?


DavisVaughan (commented):
Do you need to remove https://github.com/r-lib/testthat/blob/ae014e22d9b8f7c95296d488db0910674fecf070/R/local.R#L177-L180?

and then also the `has_dep()` helper

DavisVaughan (approved):

lionel- (approved):

PR #1793, opened 2023-05-23
Don't use snapshot reporter in parent thread

hadley:


gaborcsardi (approved):
Doesn't this cause issues with removing unused snapshots? (Not the PR itself, but how things are currently.)

hadley:
There are two cases where this might happen:

1. You remove all snapshots from a test file
2. You remove an entire test file

1 is handled by the `$end_file()` in snapshot reporter in the subprocess, so I think it's already ok.

2 is handled by `$end_reporter()`, but only when `all(tests %in% self$test_file_seen)`, which isn't true in the subprocess and will no long be handled by the parent process. Nevertheless, when I delete a test file, it's corresponding snapshot still seems to be removed. 🤔

hadley:
If I trace both `unlink()` and `file.remove()`, I don't see the snapshot being deleted. I'll take another tomorrow with a fresh mind.

R/parallel.R
```diff
@@ -70,13 +72,15 @@ test_files_parallel <- function(
...
   )
 
-  with_reporter(reporters$multi, {
-    parallel_updates <- reporter$capabilities$parallel_updates
-    if (parallel_updates) {
-      parallel_event_loop_smooth(queue, reporters, test_dir)
-    } else {
-      parallel_event_loop_chunky(queue, reporters, test_dir)
-    }
+  withr::with_dir(test_dir, {
```
hadley, line 75: I had to do this so that `SnapshotReporter$end_reporter()` was called in the correct directory. Does that seem ok @gaborcsardi? If so, I can remove the `test_dir` args from the event loops.

R/snapshot-reporter.R
```diff
@@ -127,6 +129,11 @@ SnapshotReporter <- R6::R6Class("SnapshotReporter",
     },
 
     end_file = function() {
+      if (self$parallel) {
```
hadley, line 132: I think this narrow scope is a bit safer/cleaner.

hadley (commented):
There's still a problem because the snapshot reporter in the main thread doesn't know about any full file snapshots.  @gaborcsardi can you think of some way to forward `SnapshotReport$announce_file_snapshot()` calls in the worker threads back to the main thread?

PR #1798, opened 2023-05-26
Set option when snapshotting

### LINKED ISSUE #1796: Set some option in snapshot tests
So you can use that to automatically turn snapshotting on.

hadley:
Fixes #1796


R/snapshot.R
```diff
@@ -458,3 +465,20 @@ check_variant <- function(x) {
     abort("If supplied, `variant` must be a string")
   }
 }
+
+with_is_snapshotting <- function(code) {
+  withr::local_options(testthat_is_snapshotting = TRUE)
+  code
```
jennybc, line 471: I don't necessarily think this matters but looking at `with_` functions I have implemented, I seem to do it like this:

``` r
with_is_snapshotting <- function(code) {
  withr::with_options(list(testthat_is_snapshotting = TRUE), code = code)
}
```

I guess maybe the one difference is whether you `force(code)` or not?

jennybc (approved):

kevinushey (approved):
LGTM, although I'm curious what the convention is between `testthat.option` versus `testthat_option` naming is.

hadley:
@kevinushey we're trying to move towards `_` everywhere, but there's a lot of places that `.` remain.

R/snapshot.R
hadley: I prefer the `local_` syntax pretty much everywhere.

PR #1804, opened 2023-06-04
Initial experiment with trace()

hadley:


hadley:
Not successful because you can’t use `return()` with `trace()` because it wraps the tracing statement with `.doTrace` which does `eval.parent(substitute(expr))` instead of just `expr`.  But maybe we could mock `.doTrace`?

PR #1808, opened 2023-06-04
Clarify test_code() defaults

### LINKED ISSUE #1782: `describe_it()` should set the `test_code()` reporter like `test_that()`
https://github.com/r-lib/testthat/blob/4076b66b5ee7ea9d079e5ca4580cf2f69d131333/R/describe.R#L76

In bench there is this test, which I can't run interactively:

```r
describe("bench_time", {
  skip_on_cran()
  skip_on_ci()

  res <- bench_time(1 + 1:1e7)
  it("returns process and real time", {
    expect_equal(names(res), c("process", "real"))
  })
  # snip...
})
```

The problem is that `local_interactive_reporter()` isn't set up, which is what sets `NOT_CRAN=true`, so the test fails in `skip_on_cran()` because it "looks like we are on cran"

`test_that()` does this nice thing where if it detects that there is no reporter active, it sets up an interactive reporter for you, and I think `describe_it()` (used by both `describe()` and `it()`) should do the same thing
https://github.com/r-lib/testthat/blob/4076b66b5ee7ea9d079e5ca4580cf2f69d131333/R/test-that.R#L41-L44

That should mean you can run both the full `describe()` block and just an `it()` block interactively and both should get `NOT_CRAN=true` set up correctly.

---

Hadley also mentioned during discussion about this:

> also I was wondering if skip_on_cran() should check interactive() first

> the main downside of the current behaviour is that you sometimes want to nest skip_on_cran() inside a helper, and then you can’t run that interactively

hadley:
Fixes #1782

### DIFF
--- a/R/describe.R
+++ b/R/describe.R
@@ -80,6 +80,7 @@ describe_it <- function(description, code, env = parent.frame()) {
     description,
     code,
     env = env,
+    default_reporter = local_interactive_reporter(),
     skip_on_empty = FALSE
   )
 }
--- a/R/source.R
+++ b/R/source.R
@@ -34,7 +34,12 @@ source_file <- function(path, env = test_env(), chdir = TRUE,
 
   withr::local_options(testthat_topenv = env)
   if (wrap) {
-    invisible(test_code(NULL, exprs, env))
+    invisible(test_code(
+      test = NULL,
+      code = exprs,
+      env = env,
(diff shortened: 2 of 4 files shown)


DavisVaughan (approved):
Moving the fallback reporter selection down into `test_code()` definitely makes it clearer!

PR #1816, opened 2023-06-05
New tool to compare global state before and after each test

### LINKED ISSUE #1674: Easy way to check test invariants
If you accidentally affect global state in one test (e.g. working directory, options) it can be very hard to later figure out where the problem is. It'd be cool if you could somehow register a function that testtat would call before and after each test, and then report if the objects were different.

hadley:
Fixes #1674

@lionel- could you please take a look at the implementation? 
@jennybc could you please look at the docs?


R/test-that.R
```diff
@@ -13,6 +13,46 @@
...
+#' One of the most pernicious challenges to debug is when a test runs fine
+#' in your test suite, but fails when you run it interactively (or similarly,
+#' it fails randomly when running your tests in parallel). One of the most
+#' common causes of this problem is accidentally changing global state in a
+#' previous test (e.g. changing an option, an environment variable, or the
+#' working directory). This is hard to debug, because it's very hard to figure
+#' out which test made the change.
+#'
+#' Luckily testthat provides a tool to figure out if tests are changing global
+#' state. You can write a function called `testthat_state` and
```
jennybc, line 27:
```suggestion
#' state. You can write a function called `testthat_state()` and
```

The phrasing ("called `testthat_state`") makes it hard to know whether the parens are appropriate, but I still think so. Or re-phrase.

R/test-that.R
```diff
@@ -13,6 +13,46 @@
...
+#'     envvars = Sys.getenv()
+#'   )
+#' }
+#' ```
+#'
+#' (You might discover other packages outside your control are changing
+#' the global state, in which case you might want to modify this function
+#' to ignore those values.)
+#'
+#' Other problem that can be troublesome to resolve are CRAN check notes that
```
jennybc, line 46:
```suggestion
#' Other problems that can be troublesome to resolve are CRAN check notes that
```

jennybc (approved):
I'm very excited about this!

My main feedback documentation-wise is that it's not clear where `testthat_state()` should be defined. Below `R/` or in a test helper file?

kevinushey:
Does this implementation leave the door open in case someone wants to check state only in certain tests, or configure how state is checked on a test-by-test basis?

hadley:
No, because I'm mostly imagining that people will only want to turn this on occasionally during debugging, just because it's a lot of work to run down all the false positives (as we've learned in renv).

R/test-that.R
```diff
@@ -13,6 +13,46 @@
...
+#'
+#' One of the most pernicious challenges to debug is when a test runs fine
+#' in your test suite, but fails when you run it interactively (or similarly,
+#' it fails randomly when running your tests in parallel). One of the most
+#' common causes of this problem is accidentally changing global state in a
+#' previous test (e.g. changing an option, an environment variable, or the
+#' working directory). This is hard to debug, because it's very hard to figure
+#' out which test made the change.
+#'
+#' Luckily testthat provides a tool to figure out if tests are changing global
```
lionel-, line 26: It's not clear from this documentation where you should define this function. Within a test, outside a test, within a helper file, in the `testthat.R` file?

I see in the implementation that you're looking in `testthat_env` so I guess the best place would be a helper file for this to take global effect?

Is this a common pattern that we use to implement hooks? I think I would have expected the UI to be something like:

```r
# Enable
testthat::register_state_watch(function() list())

# Disable? Then need to check for `NULL` in `after`
testthat::register_state_watch(NULL)
```

Then as a user I'd feel more comfortable registering a watcher because I know it will take effect after the call. This makes it easier to wrap in another function as well, perhaps exported by a different package.

lionel- (approved):
Looks great!

R/test-that.R
hadley: Oh I like that API much better.

hadley: Maybe `set_state_watcher()` since that at least somewhat parallels `set_max_fails()`.

PR #1817, opened 2023-06-05
Implement `expect_contains()`

### LINKED ISSUE #1346: expect_contains()
`expect_contains(haystack, needle)` shortcut for `all(needle %in% haystack)` but shows all of haystack if needle not found.

hadley:
Fixes #1346


R/expect-setequal.R
```diff
@@ -117,3 +118,28 @@ check_names_ok <- function(x, label) {
     stop("All elements in `", label, "` must be named")
   }
 }
+
+#' @export
+#' @rdname expect_setequal
+expect_contains <- function(object, expected) {
```
DavisVaughan, line 124: I am wondering if this should just be `expect_in(needles, haystack)`?

And the error could be like 

```r
fail(paste0(
      act$lab, " isn't fully contained within ", exp$lab, ".\n",
      paste0("* Unaccounted for in `actual`: ",  values(act$val[act_miss]), "\n"),
      paste0("* Present in `expected`:   ",  values(exp$val), "\n")
    ))
```

DavisVaughan: I will say it took me longer to understand `expect_contains()` because it doesn't map directly to `%in%` since the args are backwards

tests/testthat/test-expect-setequal.R
```diff
@@ -77,3 +77,21 @@ test_that("succeeds if comparing empty named and unnamed vectors", {
...
+
+# contains ----------------------------------------------------------------
+
+test_that("expect_contains() succeeds when appropriate", {
+  expect_success(expect_contains(letters, "a"))
+  expect_success(expect_contains(letters, letters))
+  expect_success(expect_contains(letters, character()))
+})
+
+test_that("expect_contains() gives useful message on faillure", {
```
DavisVaughan, line 89:
```suggestion
test_that("expect_contains() gives useful message on failure", {
```

DavisVaughan (approved):
Approving because my comments are blocking, but maybe worth considering if it should be `expect_in()` or not

R/expect-setequal.R
hadley: Hmmmm, good point. But it's also going to be confusing if its the only expectation where the expected values come first. Maybe this is an argument for keeping it as `expect_contains()`?

DavisVaughan: > the only expectation where the expected values come first.

Maybe I'd have a different use case than you? I see it as:

```r
values <- generate_output()
expect_in(values, some_set_of_possible_results)
```

hadley: I see it used in places more like `expect_contains(dir(tempdir(), "foo.csv")`.

hadley: Might as well include it now?

R/expect-setequal.R
```diff
@@ -117,3 +121,53 @@ check_names_ok <- function(x, label) {
...
+
+  if (!is_vector(act$val) || !is_vector(exp$val)) {
+    abort("`object` and `expected` must both be vectors")
+  }
+
+  act_miss <- !act$val %in% exp$val
+
+  if (any(act_miss)) {
+    fail(paste0(
+      "All values of ", act$lab, " (`actual`) aren't contained in ", exp$lab, " (`expected`).\n",
```
DavisVaughan, line 164: Alternative suggestion? I like that `act$lab` comes first, at least. 

```suggestion
      act$lab, " (`actual`) isn't fully contained within ", exp$lab, " (`expected`).\n",
```

`fully` is optional but you could also do this in `expect_contains()` to make fairly symmetric messages:

```
act$lab, " (`actual`) doesn't fully contain ", exp$lab, " (`expected)`.\n",
```

I'm not super picky about this though so if you like what you have that is fine too

tests/testthat/_snaps/expect-setequal.md
```diff
@@ -12,3 +12,31 @@
...
+    `x1` (`actual`) doesn't contain all the values in `x3` (`expected)`.
+    * Missing from `actual`: "d", "e"
+    * Present in `actual`:   "a", "b", "c"
+    
+
+# expect_in() gives useful message on failure
+
+    All values of `x1` (`actual`) aren't contained in `x2` (`expected`).
+    * Missing from `expected`: "a"
+    * Present in `expected`:   "b", "c"
```
DavisVaughan, line 33: I think this is particularly confusing because you could interpret `All values of` as implying that literally every single element of `x1` isn't contained, but really it is trying to say "at least 1 isn't contained"

DavisVaughan (approved):

R/expect-setequal.R
```diff
@@ -117,3 +121,53 @@ check_names_ok <- function(x, label) {
...
+
+  if (!is_vector(act$val) || !is_vector(exp$val)) {
+    abort("`object` and `expected` must both be vectors")
+  }
+
+  exp_miss <- !exp$val %in% act$val
+
+  if (any(exp_miss)) {
+    fail(paste0(
+      act$lab, " (`actual`) doesn't contain all the values in ", exp$lab, " (`expected)`.\n",
```
DavisVaughan, line 139:
```suggestion
      act$lab, " (`actual`) doesn't contain all the values in ", exp$lab, " (`expected`).\n",
```

DavisVaughan: If you take my suggestions make sure to catch this typo!

PR #1824, opened 2023-06-16
fix message typos

### LINKED ISSUE #1823: expect_no_message
The function is not working as expected because of 3 "s" in the word "messsage" in the code

function (object, ..., message = NULL, class = NULL) 
{
    check_dots_empty()
    expect_no_("messsage", {
        {
            object
        }
    }, regexp = message, class = class)
}

kevinushey:
Closes #1823.

### DIFF
--- a/tests/testthat/test-expect-no-condition.R
+++ b/tests/testthat/test-expect-no-condition.R
@@ -2,14 +2,14 @@
 test_that("expect_no_* conditions behave as expected", {
 
   # base R
-  expect_error(expect_no_error(stop("error")))
-  expect_error(expect_no_warning(warning("warning")))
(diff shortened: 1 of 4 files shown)


tests/testthat/test-expect-no-condition.R
```diff
@@ -1,3 +1,18 @@
+
+test_that("expect_no_* conditions behave as expected", {
+
+  # base R
+  expect_snapshot_failure(expect_no_error(stop("error")))
```
hadley, line 5: I switched these to use snapshot tests which made me re-read and re-consider the failure text. What do you think of the new output?

kevinushey: I like it!

Issue #1825, opened 2023-06-17
leftover browser() calls causes some tests to be silently ignored


kevinushey:
Will try to provide a reprex soon.

PR #1826, opened 2023-06-18
treat browser() as error

kevinushey:
Closes https://github.com/r-lib/testthat/issues/1825.

### DIFF
--- a/R/local.R
+++ b/R/local.R
@@ -65,7 +65,7 @@
 #'   cat("\n")
 #' })
 local_test_context <- function(.env = parent.frame()) {
-  withr::local_envvar(TESTTHAT = "true", .local_envir = .env)
+  withr::local_envvar("_R_CHECK_BROWSER_NONINTERACTIVE_" = "true", TESTTHAT = "true", .local_envir = .env)
   if (edition_get() >= 3) {
     local_reproducible_output(.env = .env)
   }
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # testthat (development version)
 
+* `testthat` now sets the `_R_CHECK_BROWSER_NONINTERACTIVE_` environment variable
+  when running tests. This should ensure that left-over `browser()` statements
+  will trigger an error if encountered while running tests. (#1825)
+
 # testthat 3.1.9
 
 * New `expect_contains()` and `expect_in()` that works similarly to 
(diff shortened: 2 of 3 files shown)


R/utils.R
```diff
@@ -133,3 +133,7 @@ local_tempfile1 <- function(lines, env = parent.frame()) {
   writeLines(lines, path)
   path
 }
+
+use_browser <- function() {
```
kevinushey, line 137: This function needs to be in the package namespace (rather than a helper) just to be certain that the mocked base binding really is visible to functions in the package namespace.

R/test-that.R
```diff
@@ -190,6 +190,12 @@ test_code <- function(test, code, env, default_reporter, skip_on_empty = TRUE) {
   old <- options(rlang_trace_top_env = test_env)[[1]]
   on.exit(options(rlang_trace_top_env = old), add = TRUE)
 
+  if (!interactive()) {
```
hadley, line 193: Rebinding something in the base namespace makes me a bit nervous. Maybe we should skip this on CRAN?

Ooooh, or should we set `_R_CHECK_BROWSER_NONINTERACTIVE_ `? https://github.com/wch/r-source/blob/22dc90fa26e0509cf7ec39202d6cc3416841467f/src/main/main.c#L1363

kevinushey: That is a much better solution!

R/utils.R
kevinushey: What's the right place to set that? `local_test_context`? https://github.com/r-lib/testthat/blob/f31b02446cdc6254566e7602ab5dbfded68cc538/R/local.R#L67-L72

R/test-that.R
kevinushey: What's the right place to set that? https://github.com/r-lib/testthat/blob/f31b02446cdc6254566e7602ab5dbfded68cc538/R/local.R#L67-L72?

hadley: I _think_ `local_test_context()`?

R/local.R
```diff
@@ -65,7 +65,7 @@
 #'   cat("\n")
 #' })
 local_test_context <- function(.env = parent.frame()) {
-  withr::local_envvar(TESTTHAT = "true", .local_envir = .env)
+  withr::local_envvar("_R_CHECK_BROWSER_NONINTERACTIVE_" = "true", TESTTHAT = "true", .local_envir = .env)
```
hadley, line 68: I think we still want to gate this behind `!interactive()` because it useful to use `browser()` when interactively running tests.

kevinushey: The environment variable only has an effect in non-interactive sessions so "technically" it's still okay to set here. (It also makes sure it's set if you run R as a sub-process from an interactive parent process, which could potentially pop up in some tests)

hadley: Oh duh.

Issue #1827, opened 2023-06-18
`test_local()` no longer reports on failures with parallel testing


krlmlr:
Introduced in #1800.

Expecting `out$failed` to be 1 in the output.

## 496f827aafe23da9500c48db3794325e4e0221b2 (#1800)

``` r
usethis::create_package("demo")
#> ✔ Creating 'demo/'
#> ✔ Setting active project to '/private/var/folders/dj/yhk9rkx97wn_ykqtnmk18xvc0000gn/T/RtmpoAQCgM/reprex-5c90538fc2eb-full-husky/demo'
#> ✔ Creating 'R/'
#> ✔ Writing 'DESCRIPTION'
#> Package: demo
#> Title: What the Package Does (One Line, Title Case)
#> Version: 0.0.0.9000
#> Authors@R (parsed):
#>     * First Last <first.last@example.com> [aut, cre] (YOUR-ORCID-ID)
#> Description: What the package does (one paragraph).
#> License: `use_mit_license()`, `use_gpl3_license()` or friends to pick a
#>     license
#> Encoding: UTF-8
#> Roxygen: list(markdown = TRUE)
#> RoxygenNote: 7.2.3
#> ✔ Writing 'NAMESPACE'
#> ✔ Setting active project to '<no active project>'
setwd("demo")
usethis::use_testthat(parallel = TRUE)
#> ✔ Setting active project to '/private/var/folders/dj/yhk9rkx97wn_ykqtnmk18xvc0000gn/T/RtmpoAQCgM/reprex-5c90538fc2eb-full-husky/demo'
#> ✔ Adding 'testthat' to Suggests field in DESCRIPTION
#> ✔ Adding '3' to Config/testthat/edition
#> ✔ Adding 'true' to Config/testthat/parallel
#> ✔ Creating 'tests/testthat/'
#> ✔ Writing 'tests/testthat.R'
#> • Call `use_test()` to initialize a basic test file and open it for editing.
usethis::use_test("bad")
#> ✔ Writing 'tests/testthat/test-bad.R'
#> • Edit 'tests/testthat/test-bad.R'

writeLines(
  c(
    'test_that("bad test", { expect_true(FALSE) })'
  ),
  "tests/testthat/test-bad.R"
)

out <- testthat::test_local(stop_on_failure = FALSE)
#> Starting 1 test process
#> ✔ | F W S  OK | Context
#> 
[...]

PR #1828, opened 2023-06-18
Fix reporting for parallel tests

gaborcsardi:
Closes https://github.com/r-lib/testthat/issues/1827.


krlmlr:
Thanks for the super-quick fix, works for my ultimate use case!

Are you planning a CRAN release soon?

hadley:
We can do a patch release next week

PR #1829, opened 2023-06-19
Add missing parenthesis for a sintatically valid example

llrs:


llrs (commented):
was reading the documentation about set_state_inspector (which will be very helpful) and noticed a minor error.

Thanks

PR #1840, opened 2023-07-31
Run single test

### LINKED ISSUE #1776: Feature request: Run a single named test within a file
Currently, the smallest unit of tests that can be run with this package is a file, with `testthat::test_file()`. However, for our use case each testing file (we've already split up into about 10 of them) takes 5-10 minutes to run, so the dev loop for tests at the end of a file is painfully slow. Therefore I propose something like `testthat::test_single(file, test_name)` to run a single named test within a file, skipping over all tests whose names do not match test_name.

hadley:
Fixes #1776

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -5,7 +5,7 @@
 #' @param path Path to files.
 #' @param pattern Regular expression used to filter files.
 #' @param env Environment in which to evaluate code.
-#' @param label If not-`NULL`, will run only test with this label.
+#' @param desc If not-`NULL`, will run only test with this `desc`ription.
 #' @param chdir Change working directory to `dirname(path)`?
 #' @param wrap Automatically wrap all code within [test_that()]? This ensures
 #'   that all expectations are reported, even if outside a test block.
@@ -14,7 +14,7 @@
 source_file <- function(path,
                         env = test_env(),
                         chdir = TRUE,
-                        label = NULL,
+                        desc = NULL,
                         wrap = TRUE,
                         error_call = caller_env()) {
   stopifnot(file.exists(path))
@@ -28,7 +28,7 @@ source_file <- function(path,
   con <- textConnection(lines, encoding = "UTF-8")
   on.exit(try(close(con), silent = TRUE), add = TRUE)
   exprs <- parse(con, n = -1, srcfile = srcfile, encoding = "UTF-8")
-  exprs <- filter_label(exprs, label, error_call = error_call)
+  exprs <- filter_desc(exprs, desc, error_call = error_call)
 
   n <- length(exprs)
   if (n == 0L) return(invisible())
@@ -60,8 +60,8 @@ source_file <- function(path,
   }
 }
 
-filter_label <- function(exprs, label = NULL, error_call = caller_env()) {
-  if (is.null(label)) {
+filter_desc <- function(exprs, desc = NULL, error_call = caller_env()) {
+  if (is.null(desc)) {
     return(exprs)
   }
 
@@ -72,17 +72,19 @@ filter_label <- function(exprs, label = NULL, error_call = caller_env()) {
     expr <- exprs[[i]]
 
     if (!is_call(expr, "test_that", n = 2)) {
-      include[[i]] <- TRUE
+      if (!found) {
+        include[[i]] <- TRUE
+      }
     } else {
       if (!is_string(expr[[2]]))
         next
 
-      test_label <- as.character(expr[[2]])
-      if (test_label != label)
+      test_desc <- as.character(expr[[2]])
+      if (test_desc != desc)
         next
 
       if (found) {
-        abort("Found multiple tests with specified label", call = error_call)
+        abort("Found multiple tests with specified description", call = error_call)
       }
       include[[i]] <- TRUE
       found <- TRUE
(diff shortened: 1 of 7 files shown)


R/source.R
```diff
@@ -5,13 +5,18 @@
...
 #' @param wrap Automatically wrap all code within [test_that()]? This ensures
 #'   that all expectations are reported, even if outside a test block.
 #' @export
 #' @keywords internal
-source_file <- function(path, env = test_env(), chdir = TRUE,
-                        wrap = TRUE) {
+source_file <- function(path,
+                        env = test_env(),
+                        chdir = TRUE,
+                        label = NULL,
```
kevinushey, line 17: Do you think it's worth leaving the door open to multiple labels here?

R/source.R
```diff
@@ -46,13 +52,50 @@ source_file <- function(path, env = test_env(), chdir = TRUE,
...
+  found <- FALSE
+  include <- rep(FALSE, length(exprs))
+
+  for (i in seq_along(exprs)) {
+    expr <- exprs[[i]]
+
+    if (!is_call(expr, "test_that", n = 2)) {
+      include[[i]] <- TRUE
+    } else {
+      if (!is_string(expr[[2]]))
```
kevinushey, line 77: Do you need to worry about people running code like `test_that({}, desc = "weird")`?

R/source.R
```diff
@@ -23,6 +28,7 @@ source_file <- function(path, env = test_env(), chdir = TRUE,
   con <- textConnection(lines, encoding = "UTF-8")
   on.exit(try(close(con), silent = TRUE), add = TRUE)
   exprs <- parse(con, n = -1, srcfile = srcfile, encoding = "UTF-8")
+  exprs <- filter_label(exprs, label, error_call = error_call)
```
kevinushey, line 31: Does this preserve source references?

kevinushey (approved):
LGTM!

R/source.R
hadley: I'd rather start with a single label, and then later expand, if needed.

R/source.R
hadley: I believe so because they're attached to the individual expressions. Worth double-checking though.

R/source.R
hadley: If you do that you're on your own 😄 

People do weird things here (i.e. I bet someone does something like `test_that(my_fun("my label"), ...)`. If you want to handle those cases, IMO the IDE should figure out the line numbers from its parse tree and pass those through.

R/source.R
```diff
@@ -5,13 +5,18 @@
 #' @param path Path to files.
 #' @param pattern Regular expression used to filter files.
 #' @param env Environment in which to evaluate code.
+#' @param label If not-`NULL`, will run only test with this label.
```
jennybc, line 8: We're always going to match `label` against `desc` in a `test_that()` call.

Should we just call this argument `desc`?

In any case, I think `desc` has to get mentioned here, because "label" isn't an established testthat term.

R/source.R
```diff
@@ -46,13 +52,50 @@ source_file <- function(path, env = test_env(), chdir = TRUE,
...
+    return(exprs)
+  }
+
+  found <- FALSE
+  include <- rep(FALSE, length(exprs))
+
+  for (i in seq_along(exprs)) {
+    expr <- exprs[[i]]
+
+    if (!is_call(expr, "test_that", n = 2)) {
```
jennybc, line 74:
```suggestion
    if (!is_call(expr, "test_that", n = 2) && !found) {
```

What if we only execute top-level code _before_ the matching `test_that()` call?

R/test-files.R
```diff
@@ -109,12 +109,19 @@ test_dir <- function(path,
...
 #' @export
 #' @examples
 #' path <- testthat_example("success")
 #' test_file(path)
+#' test_file(path, label = "some tests have warnings")
 #' test_file(path, reporter = "minimal")
-test_file <- function(path, reporter = default_compact_reporter(), package = NULL, ...) {
+test_file <- function(path,
+                      reporter = default_compact_reporter(),
+                      label = NULL,
```
jennybc, line 122: I think I'm sort of surprised that `label` isn't being added after `...`.

jennybc (approved):
Looks good!

One more clump of comments that are not easy to hang off the PR, about `test_file()`:

* The title of the help topic is currently "Run all tests in a single file", but the addition of `label` (or `desc`?) changes things. Should we at least remove the "all"?
* The docs for `label` here suggest that we might run multiple tests, but I think we want to stay rigidly oriented towards running a single test, yeah?

R/source.R
hadley: Ooops, I thought that argument was called the label 😭

R/source.R
hadley: Oooh good thought.

R/test-files.R
hadley: I'm not sure how we should remediate functions where `...` is in the wrong place so I'm sticking with the existing convention.

R/test-files.R
```diff
@@ -109,17 +109,17 @@ test_dir <- function(path,
 #' @inheritSection test_dir Environments
 #' @param path Path to file.
 #' @param ... Additional parameters passed on to `test_dir()`
-#' @param label Optionally, supply a string here to run only tests that
-#'   with this exact label.
+#' @param desc Optionally, supply a string here to run only a single
+#'   tests that with this `desc`ription.
```
jennybc: There's a lingering wording problem here.

PR #1841, opened 2023-08-01
Default to not running helpers when loading

hadley:


jennybc (approved):
Yes, this fixes the apparent double test helper loading surfaced in recent work in gmailr.

PR #1845, opened 2023-08-08
Make skip column 2 chars wide

### LINKED ISSUE #1844: Make skip column 2 chars wide
There are situations where you might skip more than 9 tests in a file, e.g. this happens in dbplyr a lot (because a certain database isn't available). This makes the product report look a bit off:

image

It's a minor thing but I find this a bit annoying when running the dbplyr tests 😄

mgirlich:
Closes #1844.
Not sure if it is worth a NEWS bullet which is why I left it out.


hadley:
I think it's still worth a news bullet.

PR #1850, opened 2023-08-29
Add support for test files in nested directories

### LINKED ISSUE #1605: Provide optional `recursive` argument to `test_dir()`
Would it be possible to add a `recursive` argument to `test_dir()` so that it (optionally) also looks for test files in sub-directories?

Motivation:
I have a workflow where I put long running tests - or tests that needs stuff like database access - in subfolders of the `/test` directory. This works really well as they are not automatically executed by `devtools::test()`, but I can still manually trigger them. I even made a package to support this worklfow ([testthis](https://github.com/s-fleck/testthis)). 

Now if I want to run all tests in a directory and it's subdirectories at once, I have to resort to some really ugly code where i just use `with_mock` to override the test file discovery of testthat:

```
  testthat::with_mock(
    devtools::test(...),
    `testthat::find_test_scripts` = find_test_scripts_mock
  )
```

I'd _really_ love to be able to remove this abomination from my package, especially since `with_mock()` is deprecated.

radbasa:
Closes #1605 

Allows the use of directories to organize test files.

The `recursive` argument is optional and defaults to `FALSE`.

1. `testthat::test_dir("tests/testthat/")` will work as before. Just one single directory, and with the default reporter.
2. `testthat::test_dir("tests/testthat/", recursive = TRUE)` to run all of the test scripts in all directories in `tests/testthat/`. `testthat::test_dir()` will still only see test scripts with `^test.*\\.[rR]$` filenames.


king-of-poppk (approved):

king-of-poppk (approved):

radbasa:
Hi @hadley, would you have time to review this PR for the 3.2.2 release?

king-of-poppk (approved):

hadley:
@radbasa this feels like a big change, so it'd need to wait for the next minor release, sorry.

R/test-files.R
```diff
@@ -379,8 +382,8 @@ local_teardown_env <- function(frame = parent.frame()) {
 #' @return A character vector of paths
 #' @keywords internal
 #' @export
-find_test_scripts <- function(path, filter = NULL, invert = FALSE, ..., full.names = TRUE, start_first = NULL) {
-  files <- dir(path, "^test.*\\.[rR]$", full.names = full.names)
+find_test_scripts <- function(path, filter = NULL, invert = FALSE, ..., full.names = TRUE, start_first = NULL, recursive = FALSE) {
+  files <- dir(path, "^test.*\\.[rR]$", full.names = full.names, recursive = recursive)
```
radbasa, line 386: @hadley , understood. We'll wait for the next release.

If you don't mind, I'm leaving this here as a note for when you have the time to review this pull request.

This is the only significant change to the `{testthat}` code. It's a simple use of the `recursive` flag parameter of the `dir()` function. The rest of the logical/functional changes are to expose the `recursive` flag to the relevant external facing functions.

The rest of the PR are unit tests.

hadley: Right, the code change is small, but the implications are large. For example, off the top of my head, how does this affect snapshot tests? Do they end up in the right place?

radbasa: Would it help if I add unit tests for snapshot tests, both regression for non-recursive and for the new recursive mode?

hadley: That would definitely help. It'll still require that we wait for a minor release (since I'll need to spend some time thinking about other potential areas where this change might cause problems), but it's definitely something that will need to be explored before this PR can be merged.

PR #1861, opened 2023-09-18
Tweak parallel advice

### LINKED ISSUE #1855: Remove outdated advice
> Because files are run in alphabetical order, you may want to rename your slowest test files so that they start first, e.g. test-1-slowest.R, test-2-next-slowest.R, etc.

hadley:
Fixes #1855

### DIFF
--- a/vignettes/parallel.Rmd
+++ b/vignettes/parallel.Rmd
@@ -17,11 +17,23 @@ knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
 
 To take advantage of parallel tests, add the following line to the `DESCRIPTION`:
 
-    Config/testthat/parallel: true
+```         
+Config/testthat/parallel: true
+```
 
 You'll also need to be using the 3rd edition:
 
-    Config/testthat/edition: 3
+```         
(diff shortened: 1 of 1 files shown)


vignettes/parallel.Rmd
```diff
@@ -17,11 +17,23 @@ knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
...
+Config/testthat/edition: 3
+```
+
+By default, parallel testthat will use 2 cores.
+To customise that value for your local computer we recommend adding an entry for `TESTTHAT_CPUS` to your `.Rprofile`.
+The easiest way to do that is use `usethis::edit_r_profile()` and then add something like the following:
+
+```         
+TESTTHAT_CPUS=4
+```
```
gaborcsardi, line 36: 'Basic operation' right below already mentions some of this, so maybe the two paragraphs could be merged?

gaborcsardi (commented):
I think the new text could be merged with the first paragraph of 'Basic operation'?

hadley:
Good idea; I made this all a new setup section and also brought up the info about `start-first`.

PR #1865, opened 2023-09-22
Discourage use of .package

### LINKED ISSUE #1862: Discourage `.package` a bit more

hadley:
Fixes #1862


R/mock2.R
```diff
@@ -75,21 +75,12 @@
...
-#'   .package = "anotherpackage"
-#' )
-#' ```
-#'
-#' But it's not a great idea to mock a namespace that you don't own because
-#' it affects all code in that package, not just code in your package. Instead,
-#' it's safer to either import the function into your package, or make a wrapper
-#' that you can mock:
+#' argument to `local_mocked_bindings()` but we don't recommend it because
+#' it will affect all code in that package, not just code in your package.
```
jennybc, line 81:
```suggestion
#' it will affect all calls to `anotherpackage::another_function()`, not just the calls originating in your package.
```

Consider this a weakly held opinion, but somehow it feels more precise to me.

R/mock2.R
```diff
@@ -108,10 +99,11 @@
...
 #' @param .env Environment that defines effect scope. For expert use only.
 #' @param .package The name of the package where mocked functions should be
-#'   inserted. Generally, you should not need to supply this as it will be
-#'   automatically detected when whole package tests are run or when there's
-#'   one package under active development (i.e. loaded with
-#'   [pkgload::load_all()]).
+#'   inserted. Generally, you should not supply this as it will be automatically
+#'   detected when whole package tests are run or when there's one package
+#'   under active development (i.e. loaded with [pkgload::load_all()]).
+#'   We don't recommend using this to mock functions in other packages
```
jennybc, line 105:
```suggestion
#'   We don't recommend using this to mock functions in other packages,
```

jennybc (approved):

PR #1867, opened 2023-09-22
Improve round trip error message

### LINKED ISSUE #1771: `expect_snapshot_value()` fails when one attribute contains a function
Hi,

Here is a reprex:

```r
test_that("matrix", {
  set.seed(42)
  x = 5
  attr(x, "f") = function(x) x
  expect_snapshot_value(x, style="json2")
})
```

Here, the only style that works is `serialize`, but in my real-life case, it generated a 9.7MB `.md` file that is totally unusable.

The error message does not inform that the cause of the failure is a function:
>Error in `check_roundtrip(value, load(value_enc), ..., tolerance = tolerance)`: Serialization round-trip is not symmetric.
>`environment(attr(value, 'f'))` is <env:0x000000003084be50>
>`environment(attr(roundtrip, 'f'))` is <env:global>
>i You may need to consider serialization `style`

I think in this case the function should be ignored and a warning would be enough.

hadley:
Fixes #1771

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `expect_snapshot_value()` has an improved error if the object can't be
+  safely seriazlied using the specified `style` (#1771).
+
 * `skip_if_offline()` now errors if you don't have curl installed (#1854).
 
 * All packages, regardless of whether or not they use rlang 1.0.0, now
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # testthat (development version)
 
 * `expect_snapshot_value()` has an improved error if the object can't be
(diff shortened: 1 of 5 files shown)


hadley:
After review and prior to merge, should move `check_roundtrip()` to better location. Might be better to create new `snapshot-value.R` file?

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `expect_snapshot_value()` has an improved error if the object can't be
+  safely seriazlied using the specified `style` (#1771).
```
DavisVaughan, line 4:
```suggestion
  safely serialized using the specified `style` (#1771).
```

DavisVaughan (approved):

PR #1869, opened 2023-09-25
Fix accidental override in skip_on_cran()

### LINKED ISSUE #1868: `rlang::local_interactive()` causes snapshot tests to be run on CRAN
I discovered this while doing revdep checks for the glue branch that proposes a change to the print method. I saw one failure that used `capture.output()` on a printed glue object, which is not a surprise.

But then I saw another failure in pillar for a snapshot test and it is **not** a case of `expect_snapshot(cran = TRUE)`. It's a normal snapshot test, so it should never error when `NOT_CRAN` is unset.

But this particular test needs a `local_interactive()` to trigger the user-facing output that's being snapshotted:

https://github.com/r-lib/pillar/blob/92095fcc0fffa6d2b2d42c3ec6017e7501c0c99b/tests/testthat/test-tbl-format-footer.R#L94-L100

From testthat's tests, it seems like this behaviour might be known / intentional:

https://github.com/r-lib/testthat/blob/fe01a6308153d30309ef9bf28c385848818bb96c/tests/testthat/test-skip.R#L68-L69

But is that really how it should be? I think it means you can't use snapshot tests in combination with `local_interactive()`. Or I guess you have to place an explicit `skip_on_cran()` in such a test, before the `local_interactive()` call.

hadley:
Fixes #1868

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `options(rlang_interactive = TRUE)` no longer causes `skip_on_cran()` to
+  not run on CRAN (#1868).
+
 * `skip_if_offline()` now errors if you don't have curl installed (#1854).
 
 * All packages, regardless of whether or not they use rlang 1.0.0, now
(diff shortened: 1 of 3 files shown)


jennybc (approved):
LGTM

PR #1881, opened 2023-10-09
Drop obsolete references to R3.5

MichaelChirico:
New release depends on R>=3.6.0, so these skips & conditional eval are redundant

This change covers all the relevant references to 3.5 in the code base:

https://github.com/search?q=repo%3Ar-lib%2Ftestthat+%2F3%5B.%5D5%2F&type=code


hadley:
Thanks!!

PR #1882, opened 2023-10-09
Remove obsolete backports

MichaelChirico:
Don't rely on custom logic now that R>=3.6.0 is required

If these were included via {backports}' definitions, we could rely on {lintr} to flag when such definitions become obsolete (eventually):

https://github.com/r-lib/lintr/issues/2182


hadley:
Thanks!

PR #1885, opened 2023-10-17
Add options scope to junitreporter

### LINKED ISSUE #1660: Junit reporter does not use standard decimal separator
Hi,

I identified the following issue while trying to read back a junit file created for `ggplot2` test suite. In this package, one of the tests is executed with a different option for decimal separator (https://github.com/tidyverse/ggplot2/blob/main/tests/testthat/test-aes.r#L19-L25). Therefore when testthat collects junit data it uses an altered option and due to the `as.character()` time is saved using a comma as separator (https://github.com/r-lib/testthat/blob/426eb523f6772c76175d6e8d7ca04355b1ee7ad5/R/reporter-junit.R#L93). Because of that, I'm receiving coercing error while trying to read it back. 

In the case of the come, the issue is not that huge, but it sep option can be set to an arbitrary sign rendering the time field useless to read back for someone unaware. As the issue comes from the as.character(), I suggested the following change

```
xml2::xml_attr(self$suite, "time") <- withr::with_options(list(OutDec = "."), as.character(round(self$suite_time, 3)))
```

thanks to that we won't compromise any test environment but at the same time always get numbers in a standard format recognizable by any system - with point as decimal separator. I've checked and `withr` is already a dependency so it won't add any new.

Let me know what you think about this suggestion, I'm happy to prepare a PR if you agree with my evaluation and approach.

maksymiuks:
Resolves #1660 

According to @hadley's suggestions, I just enforced the decimal operator as an option scoped to the entire function. With that change, JUnit file should always be valid regardless of the decimal separator option used in the test.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -57,7 +57,7 @@
 
 * `test_file()` gains a `desc` argument which allows you to run a single 
   test from a file (#1776).
-  
+
 # testthat 3.1.10
 
 * Fix for upcoming R-devel release.
--- a/NEWS.md
+++ b/NEWS.md
@@ -54,6 +54,9 @@
 
 * `test_file()` gains a `desc` argument which allows you to run a single 
   test from a file (#1776).
+  
+* `JunitReporter()` now uses ensures numeric values are saved the xml file 
+  with `.` as decimal separator. (@maksymiuks, #1660)
 
 # testthat 3.1.10
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # testthat (development version)
 
+* `JunitReporter()` now uses ensures numeric values are saved the xml file 
+  with `.` as decimal separator. (@maksymiuks, #1660)
+
 # testthat 3.2.0
 
 ## Lifecycle changes
@@ -55,9 +58,6 @@
 * `test_file()` gains a `desc` argument which allows you to run a single 
   test from a file (#1776).
   
-* `JunitReporter()` now uses ensures numeric values are saved the xml file 
-  with `.` as decimal separator. (@maksymiuks, #1660)
-
 # testthat 3.1.10
 
 * Fix for upcoming R-devel release.
--- a/R/reporter-junit.R
+++ b/R/reporter-junit.R
@@ -85,22 +85,27 @@ JunitReporter <- R6::R6Class("JunitReporter",
     },
 
     end_context = function(context) {
+      # Make sure that any options posed by particular tests environments
+      # will not interfere with how values are saved to the file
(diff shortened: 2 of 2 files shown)


NEWS.md
```diff
@@ -54,6 +54,9 @@
 
 * `test_file()` gains a `desc` argument which allows you to run a single 
   test from a file (#1776).
+  
+* `JunitReporter()` now uses ensures numeric values are saved the xml file
```
hadley, line 58: Can you please put this under the development version heading?

R/reporter-junit.R
```diff
@@ -85,22 +85,27 @@ JunitReporter <- R6::R6Class("JunitReporter",
...
+      # will not interfere with how values are saved to the file
+      withr::local_options(list(OutDec = "."))
       xml2::xml_attr(self$suite, "tests") <- as.character(self$tests)
       xml2::xml_attr(self$suite, "skipped") <- as.character(self$skipped)
       xml2::xml_attr(self$suite, "failures") <- as.character(self$failures)
       xml2::xml_attr(self$suite, "errors") <- as.character(self$errors)
       #jenkins junit plugin requires time has at most 3 digits
       xml2::xml_attr(self$suite, "time") <- as.character(round(self$suite_time, 3))
-
+
```
hadley, line 97: Could you please get rid of these spurious whitespace changes?

R/reporter-junit.R
```diff
@@ -85,22 +85,27 @@ JunitReporter <- R6::R6Class("JunitReporter",
     },
 
     end_context = function(context) {
+      # Make sure that any options posed by particular tests environments
+      # will not interfere with how values are saved to the file
```
hadley, line 89:
```suggestion
      # Always uses . as decimal place in output regardless of options set in test
```

hadley (commented):
Thanks for working on this!

NEWS.md
maksymiuks: I'm sorry, my bad, didn't spot that there are 2 versions back to back. Want me to copy the same subsection structure (## Lifecycle changes, ## New features etc.) or just put it under the development version and that's it?

hadley: No, just copy the bullet. We add the subsections as part of the release process.

maksymiuks: Gotcha, done :)

R/reporter-junit.R
maksymiuks: Done! Thanks

R/reporter-junit.R
maksymiuks: Done! Thanks

hadley:
Thanks!

PR #1890, opened 2023-10-27
Make skip_if_offline() more robust by checking multiple hosts

jdblischak:
r-project.org has been down all morning, which is causing my package tests to be skipped. This PR updates `skip_if_offline()` to check multiple hosts, and returns `TRUE` on the first connection. It's backwards compatible because it still works when provided a single domain to `host`. And it will be just as fast once r-project.org is restored.

I added github.com since it's hard to get much done anyways when GitHub is down. And I added captive.apple.com because I learned from [sindresorhus/is-online](https://github.com/sindresorhus/is-online#how-it-works) that this is how iOS determines internet connectivity, so it should have a very reliable uptime.

If you are open to this update, I'll add an entry to `NEWS.md`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # testthat (development version)
 
+* `skip_if_offline()` now uses `captive.apple.com` by default. This is the 
+  host that Apple devices use to check that they're online so should be 
+  a higher reliability host than `r-project.org` (@jdblischak, #1890).
+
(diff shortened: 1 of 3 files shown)


jdblischak:
> Let's begin by just swapping `r-project.org` out for `captive.apple.com`

Works for me. And I see that you already made this change and added the NEWS bullet. Thanks!

PR #1914, opened 2023-12-02
Explain abbreviations and add links

salim-b:
*BDD* and *DSL* abbreviations are not self-explanatory for most people, I think.


hadley:
Thanks!

PR #1918, opened 2023-12-30
fix typos

JBGruber:
These look like typos to me.


hadley:
Why type two t's when you can write three 😆 Thanks for fixing these!

PR #1919, opened 2024-01-06
Fix typo in vignette

ilyaZar:
docs: fix typo -> tripple "t" in testtthat -> testthat


ilyaZar:
already fixed in #1918

PR #1933, opened 2024-02-15
Add expect_*_forwarded

### LINKED ISSUE #1927: Feature request: Expect the same error (or warning or message) message as a reference
I often want to have tests to ensure that the error/warning/message is the same as the expected one.  When it is an error internal to the package, it's no trouble.  When it is not an error internal to the package, best practice is not to check that the message is the same.  But, that inhibits checking that an error from an external package is the desired error.

One way around this is to capture the expected error message, but this is cumbersome (see example below).  Would a feature that could capture and verify the message from an external package is the same be of interest?  I think that it would apply to `expect_error()`, `expect_warning()`, and `expect_error()`.

I think that it could look something like the following:

```r
expect_error(
  myfun(x = c()),
  same_message =
    {
      x <- c()
      stopifnot(length(x) == 1)
    }
)
```

``` r
myfun <- function(x) {
  stopifnot(length(x) == 1)
  x
}

expect_error(
  myfun(x = c()),
  regexp =
    attr(
      try(
        {
          x <- c()
          stopifnot(length(x) == 1)
        },
        silent = TRUE
      ),
      "condition"
    )$message,
  fixed = TRUE
)
#> Error in expect_error(myfun(x = c()), regexp = attr(try({: could not find function "expect_error"
```

<sup>Created on 2024-02-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

billdenney:
Fix #1927


hadley:
Thanks for working on this, but looking at it with fresh eyes, it's not clear this is a good fit for testthat. It feels a bit specialised, and testing errors generated by other packages does feel like a bit of an anti-pattern to me.

PR #1954, opened 2024-04-21
Focus back on build pane after accepting snapshots

olivroy:
Currently, it stays on Viewer (which becomes empty) after accepting snapshots.

This PR makes it, so that accepting snapshots with the review app focuses back to Build Pane automatically!


olivroy:
Oops, doesn't work. Sorry!

PR #1959, opened 2024-05-14
Reinforce that skip is due to a package

MichaelChirico:
Have observed this with some packages that might not easily be detected as such, e.g.

```r
skip_if_not_installed("parameters")
skip_if_not_installed("stats")
```

Alternatively, we could quote-wrap the name to make it clear the package is not a normal noun.

Incidentally fixing #1967


hadley:
Maybe use the {packagename} convention?

MichaelChirico:
> Maybe use the {packagename} convention?

Sure, this is pretty identical to the reasoning now laid out in the style guide: https://style.tidyverse.org/documentation.html#package-names

> If the package name might be misinterpreted as an ordinary word, disambiguate by following it with “package” or by wrapping the package name in {} (but not both).

Using `{` makes it less tempting to use `paste()` 😉

MichaelChirico:
(happy to update the other tests if we're aligned on the core fix)

hadley:
Yeah, I think so. 

Although looking at again, I wonder if "cannot be loaded" should be "is not installed"? That's technically less correct, but I think almost every time a package fails to load is because it's not installed.

MichaelChirico:
I would keep the more technically correct wording, I have run into a few cases where `requireNamespace()` failed on an installed package and it can be very confusing to untangle if you're thinking "but it's installed!".

A more involved approach could be to surface the error message:

```r
res <- tryCatch(loadNamespace(pkg), error=identity)
if (inherits(res, "error")) {
  if (not_found(res)) { skip("Not installed") }
  else { skip(paste("Failed to load: ", conditionMessage(res))) }
}
```

hadley:
I like it!

MichaelChirico:
OK, great! I will tidy this up with passing tests later today or tomorrow.

MichaelChirico:
For #1967, I _think_ the current change is already enough. Just the non-ASCII leading character `{` should be enough to wind up grouping these messages together, and it's probably enough to just have `skip()` messages citing packages all grouped together (i.e., even if a non-`skip_if_not_installed()` skip message uses `{pkg}` to start, it's good to have that grouped with the `skip_if_not_installed()` messages; the possibility a `skip()` message starts with `{` and is not related to a package name is ignored as very unlikely.

hadley:
Fine with me.

MichaelChirico:
Thanks for wrapping up! Slipped my mind 😵‍💫

hadley:
No problems!

PR #1974, opened 2024-07-22
proposed praise messages

pachadotdev:
#### Changes Made:

- Replaced:
  - "\U0001f4a3 Your tests are da bomb \U0001f4a3"
  - "\U0001f525 Your tests are lit \U0001f525"

  with:
  - "\U0001f3b8 Your tests rock \U0001f3b8"
  - "\U0001f44f Your tests get an ovation \U0001f44f"

#### Reason for Changes:

While the original phrases are idiomatic expressions that fun and engaging, the new messages aim to maintain a positive and enthusiastic tone while being more sensitive with potential readers. In my own case, I recently received a lot of hate for being a member of a specific community and I could only think about that when I read that my tests are "the bomb". I know you never wanted to communicate anything negative, and my change points at keeping the inclusiveness that characterizes the R community.

Respectfully,
Pacha

### DIFF
--- a/R/praise.R
+++ b/R/praise.R
@@ -21,8 +21,8 @@ praise <- function() {
     "\U0001f9ff Your tests look perfect \U0001f9ff",
     "\U0001f3af Your tests hit the mark \U0001f3af",
     "\U0001f41d Your tests are the bee's knees \U0001f41d",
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1975, opened 2024-07-22
Improve stack trace by creating message in separate variable

olivroy:
Currently, we get this. This PR will simplify a little the message.

image


hadley:
I think the problem is actually that `stop_reporter()` uses the `error_call` argument instead of the `call` argument. Do you mind fixing that instead?

PR #1979, opened 2024-08-15
Better advice for mocked base functions.

jonthegeek:
If the function is used as an argument somewhere, binding NULL to it causes that call to fail. Instead bind the actual function, and then mock that.

I think it might also be necessary to do the same for the other default packages, but I'm not certain. Will likely know better at the end of TDD!


hadley:
That's not a good solution either because you're effectively including a copy of that function into your package.

PR #2012, opened 2024-11-05
Collect more reliable trace_env

hadley:
And create more helpers for testing various more complicated success and failure modes.

Fixes #1994

@DavisVaughan no need to go deep here on the actual implementation/fix, but I'd appreciate your eyes on the newly exported expectation test helpers.

PR #2014, opened 2024-11-05
Improve ... checking logic for 3e `expect_error()` (and friends)

### LINKED ISSUE #1522: Spurious `...` warning from expect_error/warning
It makes it harder to debug a failing test because it's misleading.

```
> testthat::expect_warning("no warning", "something", ignore.case = TRUE)
Error: "no warning" did not throw the expected warning.
Warning message:
1 components of `...` were not used.

We detected these problematic arguments:
* `ignore.case`

Did you misspecify an argument? 
```

You can even see it from a passing test:

```
> testthat::expect_warning("no warning", NA, ignore.case = TRUE)
Warning message:
1 components of `...` were not used.

We detected these problematic arguments:
* `ignore.case`

Did you misspecify an argument? 
```

### LINKED ISSUE #1932: expect_error and friends report unused arguments for `fixed` when used interactively but not when used non-interactively
When I ran the code below interactively, I got the following warning:

```
Error: 1 did not throw the expected warning.
Warning in expect_warning(1, regexp = "foo", fixed = TRUE) :
  Arguments in `...` must be used.
✖ Problematic argument:
• fixed = TRUE
ℹ Did you misspell an argument name?
```

But when run in reprex, I don't see the warning about `fixed`.  Why is there different behavior for interactive vs non-interactive use for argument matching?

``` r
library(testthat)

expect_message(
  message("foo"),
  regexp = "foo",
  fixed = TRUE
)
  
expect_message(
  1,
  regexp = "foo",
  fixed = TRUE
)
#> Error: 1 did not produce any messages.

expect_warning(
  warning("foo"),
  regexp = "foo",
  fixed = TRUE
)

expect_warning(
  1,
  regexp = "foo",
  fixed = TRUE
)
#> Error: 1 did not produce any warnings.

expect_error(
  stop("foo"),
  regexp = "foo",
  fixed = TRUE
)

expect_error(
  1,
  regexp = "foo",
  fixed = TRUE
)
#> Error: 1 did not throw an error.
```

<sup>Created on 2024-02-15 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

hadley:
Fixes #1932. Fixes #1522

I'm hoping #2012 will fix the incorrect call in the error.


R/expect-condition.R
```diff
@@ -301,6 +297,13 @@ cnd_matcher <- function(base_class,
   check_string(class, allow_null = TRUE, call = error_call)
   check_string(pattern, allow_null = TRUE, allow_na = TRUE, call = error_call)
 
+  if (is.null(pattern) && dots_n(...) > 0) {
+    cli::cli_abort(
+      "{.arg ...} ignored when {.arg pattern} is not set.",
+      call = error_call
```
DavisVaughan, line 303: Yea I see `error_call = trace_env` in the caller of `cnd_matcher()` above, but I see that https://github.com/r-lib/testthat/pull/2012/ fixes that, so 👍

R/expect-condition.R
```diff
@@ -301,6 +297,13 @@ cnd_matcher <- function(base_class,
   check_string(class, allow_null = TRUE, call = error_call)
   check_string(pattern, allow_null = TRUE, allow_na = TRUE, call = error_call)
 
+  if (is.null(pattern) && dots_n(...) > 0) {
+    cli::cli_abort(
```
DavisVaughan, line 301: It looks like we warned before, just confirming you want an error now?

Before:

```r
 warn(conditionMessage(cnd), call = error_call)
```

tests/testthat/_snaps/expect-condition.md
```diff
@@ -29,14 +29,18 @@
...
-      i Did you misspell an argument name?
+      Error:
+      ! `...` ignored when `pattern` is not set.
+    Code
+      expect_condition(stop("Hi!"), "x", foo = "bar")
+    Condition
+      Error:
+      ! Failed to compare message to `pattern`.
+      Caused by error in `grepl()`:
+      ! unused argument (foo = "bar")
```
DavisVaughan, line 45: That's nice

tests/testthat/_snaps/expect-condition.md
```diff
@@ -29,14 +29,18 @@
...
     Code
       expect_condition(stop("Hi!"), foo = "bar")
     Condition
-      Warning in `expect_condition()`:
-      Arguments in `...` must be used.
-      x Problematic argument:
-      * foo = "bar"
-      i Did you misspell an argument name?
+      Error:
+      ! `...` ignored when `pattern` is not set.
```
DavisVaughan, line 38: Yea the message makes me think you really wanted `cli_warn()` or something

DavisVaughan (approved):

R/expect-condition.R
hadley: Right, because I'm more convinced now that this is a real error.

hadley: I _think_ previously I thought that `grepl()` accepted `...`? I'm not sure.

DavisVaughan: 🤔 But you don't ever call `grepl()` when `pattern` is NULL right? And that's what this is saying, I think. Just that those `...` aren't going to be used

DavisVaughan: I think I agree it should probably be an error, I would just be more aggressive in the message

```
"{.arg ...} ignored when {.arg pattern} is not set."
```

sounds like a warning but maybe:

```
"Can't specify {.arg ...} when {.arg pattern} is not set."
```

is better for an error?

hadley: Oooh got it.

R/expect-condition.R
hadley: And it worked, yay!

PR #2017, opened 2024-11-06
Implement `expect_S7_class()`

### LINKED ISSUE #1580: Add expect_s7_class

### LINKED ISSUE #2016: FR: `expect_inherits()`
Now that `base::inherits()` can accept any object in the `what` argument, and S7 encourages using reified class objects instead of class names as character vectors, it might be useful to add an `expect_inherits()` function. This function would generally behave like `expect_s3_class()` but could dispatch `nameOfClass()` on the second argument when needed.

For example, it could replace tests like those seen here: https://github.com/cole-brokamp/fr/pull/15

Possibly related issue: https://github.com/r-lib/testthat/issues/1448

hadley:
Fixes #1580. Closes #2016.


R/expect-inheritance.R
```diff
@@ -92,6 +94,33 @@ expect_s3_class <- function(object, class, exact = FALSE) {
...
+  if (!S7::S7_inherits(object)) {
+    fail(sprintf("%s is not an S7 object", act$lab))
+  } else {
+    expect(
+      S7::S7_inherits(object, class),
+      sprintf(
+        "%s inherits from %s not <%s>.",
+        act$lab,
+        paste0("<", setdiff(base::class(object), "S7_object"), ">", collapse = "/"),
+        attr(class, "name")
```
t-kalinowski, line 116:
```suggestion
        attr(class, "name", TRUE)
```

t-kalinowski (approved):

Issue #2047, opened 2025-01-09
`std::uncaught_exception` error in vendored catch.h with clang++-20 and C++20 standard


tylermorganwall:
# Description

The vendored catch.h header in the testthat package references `std::uncaught_exception`, which has been deprecated in C++17 and removed in C++20. When compiling with clang++-20 and the -std=gnu++20 flag, this leads to a compilation error:

```
clang++-20 -stdlib=libc++ -std=gnu++20 -I"/opt/R/devel/lib/R/include" -DNDEBUG  -I'/github/home/R/x86_64-pc-linux-gnu-library/4.5/Rcpp/include' -I'/github/home/R/x86_64-pc-linux-gnu-library/4.5/RcppThread/include' -I'/github/home/R/x86_64-pc-linux-gnu-library/4.5/progress/include' -I'/github/home/R/x86_64-pc-linux-gnu-library/4.5/spacefillr/include' -I'/github/home/R/x86_64-pc-linux-gnu-library/4.5/testthat/include' -I/usr/local/include   -D RAY_REPRODUCE_PERLIN -DSTRICT_R_HEADERS -DRAY_HAS_X11 -DHAS_SSE -DHAS_SSE2 -DRAYSIMD -DRAYSIMDVECOFF -fpic  -O3 -Wall -pedantic -frtti -Wp,-D_FORTIFY_SOURCE=3 -Wall -pedantic  -c test-runner.cpp -o test-runner.o
  In file included from test-runner.cpp:7:
  In file included from /github/home/R/x86_64-pc-linux-gnu-library/4.5/testthat/include/testthat.h:1:
  In file included from /github/home/R/x86_64-pc-linux-gnu-library/4.5/testthat/include/testthat/testthat.h:72:
  /github/home/R/x86_64-pc-linux-gnu-library/4.5/testthat/include/testthat/vendor/catch.h:8382:20: error: no member named 'uncaught_exception' in namespace 'std'; did you mean 'uncaught_exceptions'?
   8382 |         if ( !std::uncaught_exception() ){
        |               ~~~~~^
  /usr/lib/llvm-20/bin/../include/c++/v1/__exception/operations.h:35:31: note: 'uncaught_exceptions' declared here
     35 | _LIBCPP_EXPORTED_FROM_ABI int uncaught_exceptions() _NOEXCEPT;
        |                               ^
  In file included from test-runner.cpp:7:
[...]

PR #2060, opened 2025-02-21
Update vendered catch.h file to remove C++20 deprecated function

tylermorganwall:
I have confirmed this update fixes issue #2047 on with rhub:

https://github.com/tylermorganwall/rayrender/actions/runs/13449546083/job/37581472753

Right now this is causing a compilation error on CRAN due to a deprecated function (https://cran.r-project.org/web/checks/check_results_rayrender.html) and this error will prevent any packages from using C++ catch.h unit tests.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# testthat 3.2.3.9000
+
+* Fixed use of the deprecated C++20 function `std::uncaught_exception()` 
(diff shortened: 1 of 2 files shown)


tylermorganwall:
The failures appear to be unrelated from the change.

tylermorganwall:
The test failures appear to be unrelated to the change.

hadley:
Sorry I missed this — Kevin did #2097 which is a little more general.

PR #2119, opened 2025-07-28
Improve self-testing capabilities

### LINKED ISSUE #2110: expect_success() and expect_failure() potentially misleading under multiple conditions
As observed in #1469:

```r
expect_success({ expect(TRUE, '1'); expect(FALSE, '2') })
expect_failure({ expect(TRUE, '1'); expect(FALSE, '2') })
```

Both of these pass, whereas I might expect both to fail.

This is strictly documented correctly:

> check that there's at least one success or failure respectively.

But I wonder if that's the natural default behavior, especially with Hadley's comment that producing multiple "success" conditions is usually wrong.

hadley:
`expect_success()` and `expect_failure()` are now stricter, testing the full set of assumptions (i.e. exactly one success/failure and zero failures/successes). This means that `expect_no_successes()` and `expect_no_failures()` can be deprecated.

Fixes #2110

@lionel- could you please take a look at my restarts and reassure yourself that I'm using them correctly?

@MichaelChirico I'd love any thoughts on the documentation. It's going to take a few PRs to dig myself out of this hole, so no guarantees I'll make any bigger changes here, but I'll definitely consider them for future work.


R/expect-self-test.R
```diff
@@ -34,56 +47,88 @@ new_capture("expectation_success")
 #' @param ... Other arguments passed on to [expect_match()].
 #' @export
 expect_success <- function(expr) {
```
MichaelChirico, line 49: This seems likely to be a breaking change. I guess you have a system for revdep checking before release, but it might be worth trying out on a sample of 50/100 packages now to see how big the impact might be.

tests/testthat/test-expect-known.R
```diff
@@ -12,12 +12,9 @@ test_that("uses specified width", {
 
 test_that("creates file on first run", {
   file <- withr::local_tempfile()
-  expect_success(
-    expect_warning(
-      expect_known_output(cat("ok!\n"), file),
-      "Creating reference"
-    )
-  )
+  expect_known_output(cat("ok!\n"), file) |>
```
MichaelChirico, line 15: the 4.1.0 requirement is so satisfying 🧘‍♂️

tests/testthat/test-expect-self-test.R
```diff
@@ -1,45 +1,78 @@
...
-  expect_failure(fail("abc"), "abc")
-})
+test_that("expect_failure() requires 1 failure and zero successes", {
+  expect_success(expect_failure(fail()))
 
-test_that("succeed always succeeds", {
-  expect_success(succeed())
+  expect_failure(expect_failure({}))
+  expect_failure(expect_failure(succeed()))
+  expect_failure(expect_failure({
```
MichaelChirico, line 6: a bit torn but worth considering adding the parallel test with two `fail()`s, even though the second can't be reached?

R/expect-self-test.R
hadley: Looking at https://github.com/search?q=org%3Acran+expect_success+path%3A%2F%5Etests%5C%2Ftestthat%5C%2F%2F&type=code suggests that there will be a few packages that this breaks, but most of the uses of `expect_success()` look incorrect and think PRs that just remove `expect_success()` shouldn't be too hard for me to do.

tests/testthat/test-expect-known.R
hadley: The simple joys of being about to use R features that were released 5 years ago 😆

tests/testthat/test-expect-self-test.R
```diff
@@ -1,45 +1,78 @@
-test_that("fail always fails", {
-  expect_failure(fail())
-  expect_failure(fail("abc"), "abc")
-})
+test_that("expect_failure() requires 1 failure and zero successes", {
+  expect_success(expect_failure(fail()))
 
-test_that("succeed always succeeds", {
-  expect_success(succeed())
+  expect_failure(expect_failure({}))
```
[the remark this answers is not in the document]
MichaelChirico: to clarify I would define the alias at the top of the file, or perhaps even inside each `test_that()`, so that it's perfectly clear it's just a copy.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # testthat (development version)
 
+* `expect_no_failures()` and `expect_no_succeses()` are now deprecated as `expect_success()` now test for no failures and `expect_failure()` tests for no successes (#)
```
lionel-, line 3:
```suggestion
* `expect_no_failures()` and `expect_no_successes()` are now deprecated as `expect_success()` now test for no failures and `expect_failure()` tests for no successes (#)
```

R/expect-self-test.R
```diff
@@ -1,30 +1,43 @@
...
+  last_failure <- NULL
+
+  withRestarts(
+    withCallingHandlers(
+      expr,
+      expectation_failure = function(cnd) {
+        last_failure <<- cnd
+        n_failure <<- n_failure + 1
+        # Finish the test without bubbling up
+        invokeRestart("failed")
```
lionel-, line 16: IIUC `n_failure` is either 0 or 1, because we're unwinding back to `failed` at the first failure?

In this case I think it would be a little clearer to remove `n_failure` and rename `last_failure` (whose "last" prefix imply there might be more than one assigned here) to just `failure`. Then you test for `!is.null(failure)` instead of `n_failure > 0`.

Also fine as is but then I'd add a comment saying we expect at most 1 failure.

R/expect-self-test.R
```diff
@@ -34,42 +47,50 @@ new_capture("expectation_success")
...
-  if (is.null(exp)) {
+  if (status$n_failure == 0) {
     fail("Expectation did not fail")
-  } else if (!is.null(message)) {
-    expect_match(exp$message, message, ...)
-  } else {
-    succeed()
+  } else if (status$n_failure > 1) {
+    # This should be impossible, but including for completeness
+    fail("Expectation failed more than once")
```
lionel-, line 80: oh yeah then you wouldn't need this branch

lionel- (approved):
LGTM although internal variables might be a little clearer

R/expect-self-test.R
hadley: I think this is a good suggestion but I'm going to leave it until after #2129 because there I am encountering a weird situation where `fail()` is not causing an expectation to exit and I think that this might help me debug it.

tests/testthat/test-expect-self-test.R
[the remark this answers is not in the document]
hadley: I'll continue to contemplate that as I iterate, but I'm going to leave it off this PR for now.

tests/testthat/test-expect-self-test.R
hadley: Agreed, I think that's a good place to reinforce that we should only ever see one fail

PR #2129, opened 2025-07-28
Use `pass()` and `fail()` in testthat expectations

hadley:
Wait to merge until after #2119

Claude was not as useful as I'd hoped here, possibly because I couldn't precisely frame the principle I cared about — i.e. every branch should `pass()`, `fail()`, or error, and then use standard early return style to make it easier to read.


hadley:
@MichaelChirico no need to look at this (large!) PR, but just a heads up that I've removed `length` from `expect_shape()`, at least for now, because I've realised I don't understand at all how we're supposed to compose expectations. (I'll hopefully figure that out in the a future PR).

Issue #2091, opened 2025-07-02
Give more informative message with `expect_named()`


EmilHvitfeldt:
Currently `expect_identical()` gives a much more informative error when some is wrong compared to `expect_named()`

``` r
library(testthat)

x <- list(a = 1, b = 2, d = 3)

expect_identical(names(x), c("a", "b", "c"))
#> Error: names(x) not identical to c("a", "b", "c").
#> 1/3 mismatches
#> x[3]: "d"
#> y[3]: "c"
expect_named(x, c("a", "b", "c"))
#> Error: Names of `x` ('a', 'b', 'd') don't match 'a', 'b', 'c'
```

<sup>Created on 2025-07-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hadley:
Hmmm, the problem is that you might be using `expect_named()` with `ignore.order = TRUE`, so we can't change it in general. But maybe we could do better in the default case?

EmilHvitfeldt:
hmm something similar could be done in the `ignore.order = TRUE`, but more custom. You only have problems, "in object, not in expected" and "in expected, not in object"

``` r
library(testthat)

x <- setNames(nm = 1:100)

expect_named(x, as.character(c(1:56, 58:100)), ignore.order = TRUE)
#> Error: Names of `x` ('1', '10', '100', '11', '12', '13', '14', '15', '16', '17', '18', '19', 
#> '2', '20', '21', '22', '23', '24', '25', '26', '27', '28', '29', '3', '30', '31', '32', '33', 
#> '34', '35', '36', '37', '38', '39', '4', '40', '41', '42', '43', '44', '45', '46', '47', '48', 
#> '49', '5', '50', '51', '52', '53', '54', '55', '56', '57', '58', '59', '6', '60', '61', '62',
#>  '63', '64', '65', '66', '67', '68', '69', '7', '70', '71', '72', '73', '74', '75', '76', '77', 
#> '78', '79', '8', '80', '81', '82', '83', '84', '85', '86', '87', '88', '89', '9', '90', '91', 
#> '92', '93', '94', '95', '96', '97', '98', '99') don't match '1', '10', '100', '11', '12', 
#> '13', '14', '15', '16', '17', '18', '19', '2', '20', '21', '22', '23', '24', '25', '26', '27', 
#> '28', '29', '3', '30', '31', '32', '33', '34', '35', '36', '37', '38', '39', '4', '40', '41', 
[...]

PR #2130, opened 2025-07-28
Add more informative feedback in `expect_named()`

EmilHvitfeldt:
Hope to close https://github.com/r-lib/testthat/issues/2091

### DIFF
--- a/R/expect-named.R
+++ b/R/expect-named.R
@@ -41,19 +41,42 @@ expect_named <- function(
       sprintf("%s does not have names.", act$lab)
     )
   } else {
-    exp_names <- normalise_names(expected, ignore.order, ignore.case)
+    exp <- quasi_label(enquo(expected), arg = "expected")
+    exp$val <- normalise_names(exp$val, ignore.order, ignore.case)
     act$names <- normalise_names(act$names, ignore.order, ignore.case)
 
-    expect(
-      identical(act$names, exp_names),
-      sprintf(
-        "Names of %s (%s) don't match %s",
-        act$lab,
-        paste0("'", act$names, "'", collapse = ", "),
-        paste0("'", exp_names, "'", collapse = ", ")
-      ),
-      info = info
-    )
+    if (ignore.order) {
+      act_miss <- unique(act$names[!act$names %in% exp$val])
+      exp_miss <- unique(exp$val[!exp$val %in% act$names])
+
+      expect(
+        length(exp_miss) == 0 && length(act_miss) == 0,
+        paste0(
+          "Names of ",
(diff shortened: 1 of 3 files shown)


R/expect-named.R
```diff
@@ -41,19 +41,42 @@ expect_named <- function(
...
-        paste0("'", act$names, "'", collapse = ", "),
-        paste0("'", exp_names, "'", collapse = ", ")
-      ),
-      info = info
-    )
+    if (ignore.order) {
+      act_miss <- unique(act$names[!act$names %in% exp$val])
+      exp_miss <- unique(exp$val[!exp$val %in% act$names])
+
+      expect(
```
EmilHvitfeldt, line 52: This `expect()` block is copied and modified slightly from `expect_setequal()` to include `"Names of "`.

EmilHvitfeldt:
@hadley for the `ignore.cases = FALSE` (default case) I was thinking that the `waldo::compare()` that powers `expect_identical()` would provide a high value reporting of the issues here. If you agree would I need to include `if (edition_get() >= 3) {`?

hadley:
Can you hold off a day or two until I've finished refactoring how we write expectations? I'll have a better set of recommendations then.

R/expect-named.R
hadley: Could you explore extracting out a single helper function that we could use here, in `expect_setequal()` and `expect_mapequal()`?  A expectation helper should end in `_` and have an argument `trace_env = caller_env()` that is passed to every use of `fail()`.

R/expect-named.R
```diff
@@ -39,21 +39,48 @@ expect_named <- function(
     if (identical(act$names, NULL)) {
       msg <- sprintf("%s does not have names.", act$lab)
       return(fail(msg))
+    } else {
+      return(pass(act$val))
```
EmilHvitfeldt, line 43: The vignette doesn't explicitly mention early return `pass()` but it felt natural here

hadley: Yeah, that looks reasonable to me

R/expect-equality.R
```diff
@@ -126,7 +126,8 @@ expect_waldo_equal_ <- function(
   exp,
   info = NULL,
   ...,
-  trace_env = caller_env()
+  trace_env = caller_env(),
+  error_prefix = NULL
```
EmilHvitfeldt, line 130: The addition of `error_prefix` here and in `expect_setequal_()` are done to avoid duplication of these function by allowing me to add `"Names of "`, otherwise I would need some gnarly `$lab` modifications

PR #2134, opened 2025-07-30
Document `expect_length()` and `expect_shape()` together

hadley:
@MichaelChirico thinking about it more, I think it makes sense to not introduce a duplicated way to test for length, but we should keep `expect_length()` and `expect_shape()` close together.


MichaelChirico:
I'm fine with that; WDYT about dropping the "shape" concept and just using expect_nrow(), expect_ncol(), and expect_dim()?

hadley:
Hmmm, that's a good question. My gut feeling is that I don't like it because it increases the number of functions by quite a bit, mostly for concepts that are tightly related. It also feels out of keeping with functions like `expect_error()` that can do a lot, and while I think somewhat regret the fact that testthat provides a small number of complex expectations instead of a larger number of simpler expectations, I don't want to walk back that decision now.

hadley:
Hmmm, I think I'd find `expect_shape_nrow()` etc more compelling since it does help more with autocomplete. But there's something I like about `expect_shape()` as it is.

PR #2141, opened 2025-07-30
Automatically determine the argument name in `quasi_label()`

hadley:
This removes one thing to explain about writing expectations. I also improved the error when the argument is missing by passing along the correct call.

@lionel- not sure why we didn't do this before but it seems like it should be safe and it makes life slightly simpler.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # testthat (development version)
 
+* `expect_snapshot_value()` can now handle expressions that generate `-` (#1678) or zero length atomic vectors (#2042).
 * `expect_matches()` failures should be a little easier to read (#2135).
 * New `local_on_cran(TRUE)` allows you to simulate how your tests will run on CRAN (#2112).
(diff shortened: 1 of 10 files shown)


R/quasi-label.R
```diff
@@ -34,10 +34,21 @@
...
 #' # variable name
 #' show_failure(expect_equal(f(!!i), !!(i * 10)))
-quasi_label <- function(quo, label = NULL, arg = "quo") {
+quasi_label <- function(quo, label = NULL, arg = NULL) {
+  if (is.null(arg)) {
+    arg <- substitute(quo)
+    if (is_call(arg, "enquo")) {
+      arg <- arg[[2]]
+    }
+    arg <- deparse1(arg)
```
lionel-, line 43: slightly icky but ok I guess?

lionel-: Use `as_label()` instead of `deparse1()`?

lionel- (approved):

PR #2142, opened 2025-07-30
Reframe expectation text

hadley:
First line always describes what was expected; second and subsequent line describes the actual and/or the differences between expected and actual.


R/quasi-label.R
```diff
@@ -66,6 +65,39 @@ quasi_capture <- function(.quo, .label, .capture, ...) {
...
+auto_label <- function(expr, value) {
+  if (is.call(expr) || is.name(expr)) {
+    label <- expr_label(expr)
+    if (can_inline(value)) {
+      paste0(label, " (", as_label(value), ")")
+    } else {
+      label
+    }
+  } else {
+    expr_label(expr)
```
lionel-, line 77: I'd avoid `expr_label()` as it's a remnant from lazyeval. Do you get good results if you use `as_label()` instead?

R/quasi-label.R
```diff
@@ -66,6 +65,39 @@ quasi_capture <- function(.quo, .label, .capture, ...) {
...
+  }
+}
+
+can_inline <- function(x) {
```
lionel-, line 81: `rlang::is_syntactic_literal()` might also be useful here? Would still need the nchar check.

tests/testthat/_snaps/expect-constant.md
```diff
@@ -1,34 +1,34 @@
 # logical tests act as expected
 
-    FALSE (`actual`) is not equal to TRUE (`expected`).
+    FALSE is not equal to TRUE.
```
lionel-, line 3: Should we quote the expressions?


```suggestion
    `FALSE` is not equal to `TRUE`.
```

I guess this becomes too heavy in the parenthetised/string-quoted cases

```
    x` (`1`) is not identical to `"a"`.
```

tests/testthat/_snaps/expect-setequal.md
```diff
@@ -1,67 +1,73 @@
...
-    1:2 (`actual`) and 1:50 (`expected`) don't have the same values.
+    `x` doesn't have the same values as `y`.
     * Only in `expected`: 3, 4, 5, 6, 7, 8, 9, 10, 11, ...
     
 
 # expect_contains() gives useful message on failure
 
-    `x1` (`actual`) doesn't fully contain all the values in `x2` (`expected`).
-    * Missing from `actual`: "d"
-    * Present in `actual`:   "a", "b", "c"
```
lionel-, line 44: That's actually a case where the parameter names are useful and add context (as opposed to e.g. `object`).

PR #2148, opened 2025-07-31
Handle interrupts in tests

### LINKED ISSUE #1464: Mention current test on interrupt
This would help to know what test is in an infloop or takes too long.

hadley:
I don't see any way to test this programmatically, so I've been testing by adding `Sys.sleep(10)` to a random test. 

Fixes #1464


R/reporter-stop.R
```diff
@@ -14,48 +14,53 @@ StopReporter <- R6::R6Class(
...
+    # Expectations that should cause the test to fail (error, failure)
     n_fail = 0L,
+    # Successful expectations
+    n_success = 0L,
     stop_reporter = TRUE,
     praise = TRUE,
 
     initialize = function(stop_reporter = TRUE, praise = TRUE) {
       super$initialize()
-      self$failures <- Stack$new()
```
hadley, line 24: This is a refactoring I included in this PR because it was confusing me — `failures` did not just track failing expectations, but all expectations that we need to inform the user about.

R/reporter-stop.R
```diff
@@ -14,48 +14,53 @@ StopReporter <- R6::R6Class(
...
-      failures <- self$failures$as_list()
-      if (length(failures) == 0 && self$praise) {
-        self$cat_line(colourise("Test passed", "success"), " ", praise_emoji())
-        return()
-      }
-
-      messages <- vapply(failures, issue_summary, rule = TRUE, character(1))
-      if (length(messages) > 0) {
+      if (self$issues$size() == 0) {
+        if (self$praise && self$n_success > 0) {
```
hadley, line 53: The key change is to only report success if there was one actual success — normally if there are zero successes you'll get a skip, but that doesn't help with interrupts. (I also tried emitting a skip, but I think that adds additional unneeded noise).

Issue #1501, opened 2021-11-30
More documentation for using Catch


astamm:
The documentation says:

### Symbol Registration
> If you've opted to disable dynamic symbol lookup in your package, then you'll need to explicitly export a symbol in your package that testthat can use to run your unit tests.

Could you be more explicit as to how to do this please?
Thanks a lot.

astamm:
Thanks a lot!

PR #2151, opened 2025-07-31
expand docs / example for routine registration

kevinushey:
Closes https://github.com/r-lib/testthat/issues/1501.


hadley:
Thank you!

PR #2153, opened 2025-08-01
Use restart in `test_code()`

hadley:


R/test-that.R
```diff
@@ -175,7 +167,7 @@ test_code <- function(test, code, env, reporter, skip_on_empty = TRUE) {
 
     debug_end <- if (inherits(e, "skip_empty")) -1 else 2
     register_expectation(e, debug_end)
-    signalCondition(e)
+    invokeRestart("end_test")
```
hadley, line 170: This now seems like a clear bug — resignalling a condition inside a calling handler would have just duplicated the condition.

R/test-that.R
```diff
@@ -124,19 +124,11 @@ test_code <- function(test, code, env, reporter, skip_on_empty = TRUE) {
     # to be able to debug with the DebugReporter
     register_expectation(e, 2)
 
-    e[["handled"]] <- TRUE
     test_error <<- e
+    invokeRestart("end_test")
   }
   handle_fatal <- function(e) {
     handled <<- TRUE
-    # Error caught in handle_error() has precedence
```
hadley, line 132: No longer needed because we jump out of both the tryCatch + withCallingHandlers stack with a restart.

PR #2160, opened 2025-08-01
✨ Respect `max_failures` in `ParallelProgressReporter` ✨

### LINKED ISSUE #1162: Support `max_failures` in `ParallelProgressReporter`

hadley:
This is so surprisingly simple that I wonder if I (or more corretly Claude) has missed something.

I also made the somewhat related tweak to not do the final report (which repeats all failures etc) in the case of an early exit.

Fixes #1162


R/reporter-progress.R
```diff
@@ -258,6 +262,10 @@ ProgressReporter <- R6::R6Class(
     },
 
     end_reporter = function() {
+      if (self$is_full()) {
+        return()
+      }
+
```
gaborcsardi, line 268: Why is this needed here? I think if we are at the end of the test suite, then no need to quit without the report.

PR #2170, opened 2025-08-04
Improve failure message for comparison expectations

### LINKED ISSUE #2006: expect_comparison could use waldo to determine digits in the failure case?
https://github.com/r-lib/testthat/blob/8c2bd6efa9d338b9f340b8a108adb22e0b3c432b/R/expect-comparison.R#L39

Motivation: https://github.com/imbs-hl/ranger/issues/747

I know {waldo} has recently improved "what's the right number of digits to show when `compare()` fails", so presumably we could re-use that logic here, e.g. `compare(act - exp, 0)`. The specifics may need to be more careful, but I think it's clearly not helpful to have `expect_gt()` fail with message "Difference: 0`.

hadley:
Using the digits calculation from waldo, displaying the actual calculation, comparison, and renaming "more" to "greater"

Fixes #2006

@MichaelChirico could you please take a look? This doesn't need to be perfect, but is hopefully better than the current display.


R/expect-comparison.R
```diff
@@ -31,21 +31,27 @@ expect_compare_ <- function(
   msg <- c(
     "<" = "not strictly less than",
     "<=" = "not less than",
-    ">" = "not strictly more than",
-    ">=" = "not more than"
+    ">" = "not strictly greater than",
+    ">=" = "not greater than"
   )[[operator]]
 
+  negated_op <- switch(operator, "<" = ">", "<=" = ">=", ">" = "<", ">=" = "<=")
```
MichaelChirico, line 38: Aren't these wrong? e.g. I think we'll get

```
expect_snapshot_failure(expect_gt(x, 1.0 * x))
```

```
    `x` is not strictly greater than 1.0 * x.
    0.0000100 - 0.0000110 = 0 < 0
```

(even if I'm reading it wrong we'll still want a test of the edge case)

PR #2173, opened 2025-08-05
Automatically skip missing packages on CRAN

### LINKED ISSUE #1598: Automatically skip in case of `packageNotFoundError`?
As in this change I've made for webR checks: https://github.com/lionel-/testthat/commit/a53e90a7c0075330302dbca0b9dfdc7b4aa5b2f0.

Perhaps only Suggested packages should be skipped.

hadley:
Practically, this should mean that you never need to add `skip_if_not_installed()` for suggested packages. It's already the tidyverse style to not do this, but now this works everywhere.

Fixes #1598

@MichaelChirico is there anyway we could make this more convenient for google infrastructure? Maybe consult another env var?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # testthat (development version)
 
-* On CRAN, `test_that()` will now automatically skip if a package is not installed (#1585). Practically, this means that you no longer need to check that suggested packages are installed. (We don't do this in the tidyverse because we think it has limited payoff, but other styles advise differently.)
+* On CRAN, `test_that()` now automatically skips if a package is not installed (#1585). Practically, this means that you no longer need to check that suggested packages are installed. (We don't do this in the tidyverse because we think it has limited payoff, but other styles advise differently.)
 * `expect_snapshot()` no longer skips on CRAN, as that skips the rest of the test. Instead it just returns, neither succeeding nor failing (#1585).
 * Interrupting a test now prints the test name. This makes it easier to tell where a very slow test might be hanging (#1464)
 * Parallel testthat now does not ignore test files with syntax errors (#1360).


R/test-that.R
```diff
@@ -167,6 +167,11 @@ test_code <- function(code, env, reporter = NULL, skip_on_empty = TRUE) {
           }
         },
         expectation = handle_expectation,
+        packageNotFoundError = function(e) {
+          if (on_cran()) {
+            skip(paste0(e$package, " is not installed."))
```
MichaelChirico, line 172: Maybe sync it exactly with the `skip_if_not_installed()` message?

https://github.com/r-lib/testthat/blob/60998023ddb0a1fd812bc66b26e47cb66f219498/R/skip.R#L107

R/test-that.R
```diff
@@ -167,6 +167,11 @@ test_code <- function(code, env, reporter = NULL, skip_on_empty = TRUE) {
           }
         },
         expectation = handle_expectation,
+        packageNotFoundError = function(e) {
```
MichaelChirico, line 170: this looks like a great & reasonable compromise -- minimal implementation & fits well with existing approach that always sets `!on_cran` in package GHA.

one possible concern is that it doesn't make any effort to distinguish "package missing in current environment" from "you forgot to add a package to your `Suggests`". I guess `R CMD check` is inspecting `tests/` files for `pkg::` usage so it should be caught anyway.

MichaelChirico:
> @MichaelChirico is there anyway we could make this more convenient for google infrastructure? Maybe consult another env var?

no I think it's fine. thanks for the consideration! AIUI it will now work exactly as `skip_if_not_installed()` does (since we test `on_cran()`) -- the test will pass and it's the importer's responsibility to add requirements as needed.

One advantage of the `skip_if_not_installed()` approach is we can easily do static analysis of testthat/ files and pull out which packages are needed. With this, we'll need to pull that information out from the test logs, but that's basically how it goes in the current status quo.

PR #2174, opened 2025-08-05
Disable skipping if only one snapshot left

### LINKED ISSUE #2025: Is "Skip" in the snapshot review app broken?
I'm working on making links like `testthat::snapshot_review()` work in Positron's package development tasks, which run in a terminal. In the course of this effort, I've launched that Shiny app _a lot_ and it sure feels like "Skip" just does nothing. Is it me? Or is it broken?

The toy package I made this recoding with is here: https://github.com/jennybc/clilinks

At the time of writing, it has exactly 1 snapshot test, which will always fail.

https://github.com/user-attachments/assets/2b764370-8d95-4074-8555-8b166038b703

hadley:
Fixes #2025

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # testthat (development version)
 
+* `snapshot_review()` disables the skip button if there's only one snapshot to review (#2025).
 * Interrupting a test now prints the test name. This makes it easier to tell where a very slow test might be hanging (#1464)
 * Parallel testthat now does not ignore test files with syntax errors (#1360).
 * `expect_lt()`, `expect_gt()`, and friends have a refined display that is more likely to display the correct number of digits and shows you the actual values compared.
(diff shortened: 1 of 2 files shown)


hadley:
Need variant for single file case: no file select drop down, no skip.

Now we have `snapshot_reject()` also need a reject button.

hadley:
@jennybc can you please take another look? Now that we both understand where the problem is, I think this solution should be much more compelling 😄

jennybc (approved):
This works well for me now.

Through the process of reviewing/discussing this, I've also refined my mental model around what's actually going on during snapshot review. It has helped me to realize that the app's buttons really are just a way to present choices around "what are we going to do about `_snaps/xyz.new.md`?".

hadley:
That's great to hear. It also helped my mental model, especially adding the reject button.

PR #2185, opened 2025-08-07
✨Use cli for all errors etc ✨

hadley:


R/parallel-taskq.R
```diff
@@ -220,21 +215,22 @@ task_q <- R6::R6Class(
...
-          paste0(
-            "testthat subprocess failed to start, stderr:\n",
-            msg$error$stderr
+        cli::cli_abort(
+          c(
+            "testthat subprocess failed to start.",
+            " " = "{no_wrap(msg$error$stderr)}"
           ),
           test_file = NULL,
-          parent = msg$error,
```
gaborcsardi, line 229: I think this looses the parent error here.

R/parallel.R
```diff
@@ -110,7 +107,7 @@ default_num_cpus <- function() {
   if (!is.null(ncpus)) {
     ncpus <- suppressWarnings(as.integer(ncpus))
     if (is.na(ncpus)) {
-      abort("`getOption(Ncpus)` must be an integer")
+      cli::cli_abort("{.fn getOption}({.arg Ncpus}) must be an integer.")
```
gaborcsardi, line 110: This is weird, maybe use `{.code ... ` instead? Or at least double quote the option name?

R/source.R
```diff
@@ -125,9 +128,17 @@ source_dir <- function(
   chdir = TRUE,
   wrap = TRUE
 ) {
-  files <- normalizePath(sort(dir(path, pattern, full.names = TRUE)))
```
gaborcsardi, line 128: `normalizePath()` is lost here. Intentionally?

gaborcsardi (approved):
Had some minor comments.

R/parallel-taskq.R
gaborcsardi: I guess now that we collect the stdout/stderr separately, it does not, in most cases.

PR #2191, opened 2025-08-08
Include test path in snaphot hints

### LINKED ISSUE #1577: Provide `path` for `snapshot_review()` hint
When running a test with a default path of `./tests/testthat`, the snapshot hint makes sense:
```
Run `testthat::snapshot_review('app-set-inputs/')` to review changes
```

However, it would be nice to the have the starting location in the hint if the starting location is not `./tests/testthat`.

```
testthat::test_dir("a/b/c/d/tests/testthat")
#> .....
#> Run `testthat::snapshot_review('app-set-inputs/', 'a/b/c/d/tests/testthat')` to review changes
#> .....
```

hadley:
If you're not testing the current directory/package.

Fixes #1577


hadley:
@schloerke could you please take a look and double check that this meets your needs?

schloerke:
And the base reporter hints as well, please: https://github.com/r-lib/testthat/blob/71a33cdde7704ad0e95d88160cb754a2f5866943/R/reporter-check.R#L147-L148. Thank you!

schloerke:
> @schloerke if the updated code doesn't help, can you please give me an example of how you're running the tests, and then I can iterate.

* Checkout `rstudio/shinycoreci`
* Install dev dependencies w/ `{pak}`
* Within R, run tests with `shinytest2::test_app("inst/apps/001-hello")`

hadley:
Where do I get shinyjster from?

hadley:
@schloerke I figured it out and made a bunch of fixes. Let me know how this version goes 😄

R/snapshot.R
```diff
@@ -387,28 +381,66 @@ expect_snapshot_helper <- function(
...
+    return("")
+  }
+
   if (reset_output) {
     local_reporter_output()
   }
 
-  if (is.null(variant) || variant == "_default") {
-    name <- file
+  id <- c(file, name)
```
schloerke, line 398: Q: The hint has changed from "Review the test file snapshots" to "Review this specific snapshot only", which causes work (delete the file name or many copy/paste's) to review more than one snapshot in a single test file. While the new code is accurate, it makes reviewing more time consuming.

I'd advocate for still returning the variant name for review. 

Expected:
* `testthat::snapshot_review("mytest/", "inst/apps/001-hello/tests/testthat")`
Not:
* `testthat::snapshot_review("mytest/002.png", "inst/apps/001-hello/tests/testthat")`

```suggestion
  id <- file
```

-------------------------

Or would it be possible to have the final comments suggest to review the test file's snaps.

hadley: Ooooooh, I did wonder about that, but I thought that was the existing behaviour too. I removed the variant because it didn't work, but I think that's because it doesn't work when both file and variant are supplied.

(There's something about the code paths here that I find extremely confusing.)

schloerke (approved):

PR #2194, opened 2025-08-08
Some notes on snapshot cleanup

### LINKED ISSUE #1615: Add information about snapshot cleaning workflow in vignette, especially regarding variant
This is a question I asked myself in the past: Does `_snaps` folder need some cleaning when you change the snapshot expectation ? 

This topic arises mainly when using `variant` argument and the value change because of refactoring for example. The folder for snapshot will change and so what is needed in the process as those files are followed on git  ? 

It seems **testthat** will take care of that and a local `devtools::test()` should remove what is unused. 

If that is all that would be interesting to mention. If there are some edge case, it could be good too.

hadley:
Fixes #1615

@cderv does this help clarify behviour?

### DIFF
--- a/vignettes/snapshotting.Rmd
+++ b/vignettes/snapshotting.Rmd
@@ -128,6 +128,8 @@ test_that("bullets", {
 If this is a deliberate change, you can follow the advice in the message and update the snapshots for that file by running `snapshot_accept("pizza")`; otherwise you can fix the bug and your tests will pass once more.
 (You can also accept snapshot for all files with `snapshot_accept()`).
 
+If you delete the test, the corresponding snapshot will be remove the next time you run the tests. If you delete all snapshots in the file, the entire snapshot file will be deleted the next time you run all the tests.
+
 ### Snapshot format
 
 Snapshots are recorded using a subset of markdown.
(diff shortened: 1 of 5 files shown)


vignettes/snapshotting.Rmd
```diff
@@ -128,6 +128,8 @@ test_that("bullets", {
 If this is a deliberate change, you can follow the advice in the message and update the snapshots for that file by running `snapshot_accept("pizza")`; otherwise you can fix the bug and your tests will pass once more.
 (You can also accept snapshot for all files with `snapshot_accept()`).
 
+If you delete the test, the corresponding snapshot will be remove the next time you run the tests. If you delete all snapshots in the file, the entire snapshot file will be deleted the next time you run all the tests.
```
cderv, line 131: Except when variants are used, right? 
This is what I understood from the clarification in `snapshot.R` help addition. 

In that case, the user is responsible for cleaning old, unused variants snapshot.

cderv (approved):
Thanks ! 

It adds useful clarification about variants. 

Just added a confirmation of my understand as the vignettes does not mention variant again.

PR #2195, opened 2025-08-08
More improvements to match output

hadley:
Particularly focussing on multiline outputs

### DIFF
--- a/tests/testthat/_snaps/expect-match.md
+++ b/tests/testthat/_snaps/expect-match.md
@@ -5,23 +5,39 @@
 ---
 
     `one` does not match regexp "asdf".
-    Text: "bcde"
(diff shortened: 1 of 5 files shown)


tests/testthat/_snaps/expect-match.md
```diff
@@ -5,23 +5,39 @@
 ---
 
     `one` does not match regexp "asdf".
-    Text: "bcde"
+    Text:
+    ✖ │ bcde
```
gaborcsardi, line 9: Will the UTF-8 snapshots cause issues on non-UTF-8 platforms? Maybe testthat does something about this, I forgot.

gaborcsardi (approved):
Fancy!

PR #2198, opened 2025-08-09
Vignette on challenging tests

### LINKED ISSUE #1265: Add a prominent link to mockery?
At the moment the only reference to mockery is via the deprecation message of former mocking functions.

Could there be a link to mockery maybe from the README?

hadley:
Fixes #1265

### DIFF
--- a/vignettes/challenging-tests.Rmd
+++ b/vignettes/challenging-tests.Rmd
@@ -1,8 +1,8 @@
 ---
-title: "Challenging Testing Problems"
+title: "Challenging testing problems"
 output: rmarkdown::html_vignette
 vignette: >
-  %\VignetteIndexEntry{Challenging Testing Problems}
+  %\VignetteIndexEntry{Challenging testing problems}
   %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
 ---
@@ -14,30 +14,37 @@ knitr::opts_chunk$set(
 )
 ```
 
+Testing is easy when your functions are pure: they take some inputs and return predictable outputs. But real-world code often involves randomness, external state, graphics, user interaction, and other challenging elements. This vignette provides practical solutions for testing these tricky scenarios.
+
+Other packages:
+
+* For testing graphical output, we recommend vdiffr.
+* For testing code that uses HTTP requests we recommend vcr or httptest2.
+
 ```{r setup}
 library(testthat)
 ```
 
-Testing is easy when your functions are pure: they take some inputs and return predictable outputs. But real-world code often involves randomness, external state, graphics, user interaction, and other challenging elements. This vignette provides practical solutions for testing these tricky scenarios.
+## External state
 
-## Output Affected by RNG
+Tests should be isolated from global options, environment variables, and other external state that might affect behavior.
+
+### Output affected by RNG
 
 Random number generation can make tests non-deterministic. Use `withr::local_seed()` to ensure reproducible results within your tests.
 
-### The Problem
-
 ```{r, eval = FALSE}
-# This test will randomly pass or fail
-test_that("random sample has expected properties", {
-  x <- sample(1:100, 10)
-  expect_length(x, 10)
-  expect_true(all(x %in% 1:100))
-  # This might fail randomly:
-  expect_equal(x[1], 42)
+simulate_data <- function(n) {
+  rnorm(n, mean = 0, sd = 1)
+}
+
+test_that("simulate_data returns correct structure", {
+  result <- simulate_data(5)
+  expect_length(result, 5)
+  expect_type(result, "double")
+  expect_equal(result[1], 1.048, tolerance = 0.001)
 })
 ```
-
-### The Solution
 
 ```{r}
 test_that("random sample has expected properties", {
@@ -50,28 +57,7 @@ test_that("random sample has expected properties", {
 })
 ```
 
-For functions that internally use random numbers:
-
-```{r}
-simulate_data <- function(n) {
-  rnorm(n, mean = 0, sd = 1)
-}
-
-test_that("simulate_data returns correct structure", {
-  withr::local_seed(456)
-  result <- simulate_data(5)
-  expect_length(result, 5)
-  expect_type(result, "double")
-  # Test specific values with fixed seed
-  expect_equal(result[1], 1.048, tolerance = 0.001)
-})
-```
-
-## Output Affected by External State
-
-Tests should be isolated from global options, environment variables, and other external state that might affect behavior.
-
-### Global Options
+### Global options
 
 ```{r}
 # Function that depends on global options
@@ -89,7 +75,7 @@ test_that("format_number respects digits option", {
 })
 ```
 
-### Environment Variables
+### Environment variables
 
 ```{r}
 # Function that depends on environment variables
@@ -108,11 +94,11 @@ test_that("get_api_url uses default when env var not set", {
 })
 ```
 
-### Working Directory
+### Reading and writing files
 
 ```{r}
 test_that("function works in different directories", {
-  withr::local_dir(tempdir())
+  withr::local_dir(withr::local_tempdir())
   # Test code that depends on working directory
   writeLines("test content", "temp_file.txt")
   expect_true(file.exists("temp_file.txt"))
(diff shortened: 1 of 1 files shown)


vignettes/challenging-tests.Rmd
```diff
@@ -0,0 +1,145 @@
...
+
+# Pretend we're snapshotting
+snapper <- local_snapshotter(fail_on_new = FALSE)
+snapper$start_file("snapshotting.Rmd", "test")
+
+# Pretend we're testing testthat so we can use mocking
+Sys.setenv(TESTTHAT_PKG = "testthat")
+```
+
+This vignette is a quick reference guide for testing challenging functions. It's organised by the problem, rather than technique used to solve it, so you can quickly skim the whole vignette, spot the problem your facing, then learn more about useful tools for solving it.
```
mcol, line 23:
```suggestion
This vignette is a quick reference guide for testing challenging functions. It's organised by the problem, rather than technique used to solve it, so you can quickly skim the whole vignette, spot the problem you're facing, then learn more about useful tools for solving it.
```

vignettes/challenging-tests.Rmd
```diff
@@ -0,0 +1,145 @@
...
+
+test_that("three dice adds values of individual calls", {
+  local_mocked_bindings(dice = mock_output_sequence(1, 2, 3))
+  expect_equal(roll_three(), 6)
+})
+```
+
+## HTTP requests
+
+If you're trying to test functions that rely on HTTP requests, we recommend using {vcr} or {httptest2}. These packages provide the ability to record and then later reply HTTP requests so that you can test without an active internet connection. If your package is going to CRAN, we highly recommend either using one of these packages or using `skip_on_cran()` for your internet facing tests. This ensures that your package won't break on CRAN just because the service you're using is temporarily down.
```
mcol, line 66:
```suggestion
If you're trying to test functions that rely on HTTP requests, we recommend using {vcr} or {httptest2}. These packages provide the ability to record and then later reply HTTP requests so that you can test without an active internet connection. If your package is going to CRAN, we highly recommend either using one of these packages or using `skip_on_cran()` for your internet-facing tests. This ensures that your package won't break on CRAN just because the service you're using is temporarily down.
```

vignettes/mocking.Rmd
```diff
@@ -0,0 +1,226 @@
...
+
+(If, like me, you're confused as to why you'd want to cruelly make fun of your tests, here mocking is used in the sense of making a fake or simulated version of something, i.e. a mock-up.)
+
+testthat supports mocking primarily through `local_mocked_bindings()` for mocking functions, and we'll focus on that function in this vignette. But it also provides other methods for specialised cases: you can use `local_mocked_s3_method()` to mock an S3 method, `local_mocked_s4_method()` to mock a S4 method, and `local_mocked_r6_class()` to mock an R6 class. Once you understand the basic idea of mocking, I think it should be straightforward to apply these other functions where needed.
+
+## Getting started with mocking
+
+Lets begin by motivating mocking with a simple example. Imagine you're writing a function like `rlang::check_installed()`. The goal of this function is to check if a package is installed, and if not, give a nice error message. It also takes an option `min_version` argument which you can use to enforce a version constraint. A simple base R implementation might look something like this:
```
mcol, line 31:
```suggestion
Let's begin by motivating mocking with a simple example. Imagine you're writing a function like `rlang::check_installed()`. The goal of this function is to check if a package is installed, and if not, give a nice error message. It also takes an option `min_version` argument which you can use to enforce a version constraint. A simple base R implementation might look something like this:
```

vignettes/mocking.Rmd
```diff
@@ -0,0 +1,226 @@
...
+test_that("check_installed() checks package is installed", {
+  local_mocked_bindings(requireNamespace = function(...) TRUE)
+  expect_no_error(check_installed("package-name"))
+
+  local_mocked_bindings(requireNamespace = function(...) FALSE)
+  expect_snapshot(check_installed("package-name"), error = TRUE)
+})
+```
+
+For the second test, we mock `requireNamepace()` to return `TRUE`, and then `packageVersion()` to return a fixed version number. Together this simulated that version 2.0.0 of any package is installed and makes our snapshot rely only on the state set within the test.
```
mcol, line 105:
```suggestion
For the second test, we mock `requireNamepace()` to return `TRUE`, and then `packageVersion()` to return a fixed version number. Together this simulates that version 2.0.0 of any package is installed and makes our snapshot rely only on the state set within the test.
```

vignettes/mocking.Rmd
```diff
@@ -0,0 +1,226 @@
...
+  )
+  
+  expect_no_error(check_installed("package-name", "1.0.0"))
+  expect_snapshot(check_installed("package-name", "3.4.5"), error = TRUE)
+})
+```
+
+## Case studies
+
+To give you a bit more experience with mocking, this section looks at a few places where we use mocking throughout the tidyverse. Note that these are all a little complex; this is the nature of mocking: if you can use a simpler technique, you should. So mocking on comes up for otherwise intractable problems.
```
mcol, line 121:
```suggestion
To give you a bit more experience with mocking, this section looks at a few places where we use mocking throughout the tidyverse. Note that these are all a little complex; this is the nature of mocking: if you can use a simpler technique, you should. So mocking only comes up for otherwise intractable problems.
```

vignettes/mocking.Rmd
```diff
@@ -0,0 +1,226 @@
...
+  time <- 2
+  expect_equal(timer(), 1)
+})
+```
+
+## How does mocking work?
+
+To finish up, it's worth discussing how mocking works. It took us some iteration (`testthat::with_mock()`, as well as {mockery}, {mockr}, and {mockthat} packages) to get to current state and understanding how it works will help you to understand some of the tradeoffs. 
+
+The fundamental tension of mocking is that you want it to actually work (i.e. temporarily modify the implementation of a function), but also be "hygenic", i.e. it should only allow you to modify functions that you own, not arbitarily affect any code. To achieve this goal `local_mocked_bindings()` works by modifying an environment that you own: your package's namespace environment. (If you've forgotten exactly what these are, you can refresh your memory at <https://adv-r.hadley.nz/environments.html#special-environments>.) So you can imagine mocking works something like this if you're temporarily replacing the value of `my_function` with a `new` implementation:
```
mcol, line 206:
```suggestion
The fundamental tension of mocking is that you want it to actually work (i.e. temporarily modify the implementation of a function), but also be "hygienic", i.e. it should only allow you to modify functions that you own, not arbitrarily affect any code. To achieve this goal `local_mocked_bindings()` works by modifying an environment that you own: your package's namespace environment. (If you've forgotten exactly what these are, you can refresh your memory at <https://adv-r.hadley.nz/environments.html#special-environments>.) So you can imagine mocking works something like this if you're temporarily replacing the value of `my_function` with a `new` implementation:
```

vignettes/mocking.Rmd
```diff
@@ -0,0 +1,226 @@
...
+
+1. The package namespace is locked, which means that you can't add new bindings to it. That means if you want to mock base functions, you have to provide some binding that can be overriden. The easiest way to do this is (e.g.) `mean <- NULL`. This creates a binding that `local_mocked_bindings()` can modify, but because of R's [lexical scoping rules](https://adv-r.hadley.nz/functions.html#functions-versus-variables) this doesn't affect ordinary calls to `mean()`.
+
+1. `::` doesn't use the package namespace, so if you want to mock a function from another package that you call with `::`, you either have to switch from `pkg::fun()` to `fun()` by importing `fun` into your `NAMESPACE` (e.g. with `@importFrom pkg fun`), or create your own wrapper function that you can mock. Typically, one of these options will feel fairly natural.
+
+Overall these limitations feel correct to me: `local_mocked_bindings()` makes it easy to temporarily change the implementation of functions that you have written, while offering work arounds to override the implementations of functions that you others have written in the scope of your package.
```
mcol, line 226:
```suggestion
Overall these limitations feel correct to me: `local_mocked_bindings()` makes it easy to temporarily change the implementation of functions that you have written, while offering workarounds to override the implementations of functions that you others have written in the scope of your package.
```

vignettes/snapshotting.Rmd
```diff
@@ -215,7 +219,48 @@ test_that("you can't add weird things", {
 })
 ```
 
-## Snapshotting values
+
+Snapshot tetss are particularly important when testing complex error messages.
```
mcol, line 223:
```suggestion
Snapshot tests are particularly important when testing complex error messages.
```

vignettes/test-fixtures.Rmd
```diff
@@ -31,21 +31,62 @@ Ideally, a test should leave the world exactly as it found it. But you often nee
...
+  invisible() 
+}
+```
+
+This is particularly useful when you need to chain together multiple `local_` functions to achieve the desired state.
+
 ## Foundations
 
-Before we can talk about test fixtures, we need to lay some foundations to help you understand how they work. We'll motivate the discussion with a `sloppy()` function that prints a number with a specific number of significant digits by adjusting an R option:
+Before we go further, lets lay some foundations to help you understand how `local_` functions work. We'll motivate the discussion with a `sloppy()` function that prints a number with a specific number of significant digits by adjusting an R option:
```
mcol, line 89:
```suggestion
Before we go further, let's lay some foundations to help you understand how `local_` functions work. We'll motivate the discussion with a `sloppy()` function that prints a number with a specific number of significant digits by adjusting an R option:
```

mcol (commented):
I've marked a few typos. These vignettes are very helpful additions to the documentation!

hadley:
Thanks @mcol !

vignettes/test-fixtures.Rmd
```diff
@@ -31,21 +31,62 @@ Ideally, a test should leave the world exactly as it found it. But you often nee
...
+```{r}
+local_digits <- function(sig_digits, env = parent.frame()) {
+  withr::local_options(digits = sig_digits, .local_envir = env)
+
+  # mark that this function is called for its side-effects not its return value
+  invisible() 
+}
+```
+
+This is particularly useful when you need to chain together multiple `local_` functions to achieve the desired state.
```
jennybc, line 85: This feels like some text got lifted from elsewhere or moved and it doesn't quite make sense anymore? It doesn't feel like the first example, using `local_options()` segues naturally into the 2nd paragraph ("If you write a lot of code like this ..."). What "duplication" are we talking about? It feels like this must have unfolded differently in the original prose? Sorry this is a vague impression.

hadley: Removed, thanks!

Issue #2066, opened 2025-03-05
Running a single snapshot test deletes snapshots from all other snapshot tests in that file


jennybc:
Moving this conversation here: https://github.com/posit-dev/positron/issues/6580#issuecomment-2699248525

You can grab this toy package where I've been exercising various things when I work on Positron's test explorer: <https://github.com/jennybc/testfun>

(For the record, I'm writing this when testfun's `main` is at 8d84efc, just in case it evolves before running this reprex.)

Run an individual test file with 3 snapshot tests:

```
> test_file("tests/testthat/test-snapshots.R")

══ Testing test-snapshots.R ═══════════════════════════════════════════════════════
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 3 ] Done!
```

All is well. All tests are passing. All snapshots are there.

Now run just one of those snapshot tests (which is ultimately what Positron's test explorer lets users do with various UI gestures):

```
> test_file("tests/testthat/test-snapshots.R", desc = "snapshot test 2")

══ Testing test-snapshots.R ═══════════════════════════════════════════════════════
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1 ] Done!
```

All seems well. But the git diff shows that the snapshot file has changed and snapshots 1 and 3 have gone missing. You can see this also by re-running the whole file:

```
> test_file("tests/testthat/test-snapshots.R")

══ Testing test-snapshots.R ═══════════════════════════════════════════════════════
[ FAIL 0 | WARN 2 | SKIP 0 | PASS 3 ]

── Warning (test-snapshots.R:2:3): snapshot test 1 ────────────────────────────────
Adding new snapshot:
Code
  # first snapshot

── Warning (test-snapshots.R:10:3): snapshot test 3 ───────────────────────────────
Adding new snapshot:
Code
  # third snapshot
[ FAIL 0 | WARN 2 | SKIP 0 | PASS 3 ]
```

DavisVaughan:
This is probably the same issue, but here's proof that it doesn't have to be a snapshot test that you run. Running any test will clear out the snapshot file for that test file.

https://github.com/user-attachments/assets/0b115be7-a53a-4ef9-9367-541208eab160

d-morrison:
This issue also comes up when snapshot tests are conditionally skipped (e.g., on certain OSes); `devtools::test()` deletes skipped snapshots as unused (can reprex if helpful)

jennybc:
[...]

PR #2199, opened 2025-08-09
Snapshot refactoring

hadley:
* Generalise `test_file_reporter()` so we no longer need `test_files_reporter_parallel()`
* Remove code needed for old R
* Drop deprecated argument

`local_snapshotter()` is [only used in devtools](https://github.com/search?q=org%3Acran+local_snapshotter+&type=code).

In preparation for #2066


R/test-files.R
```diff
@@ -306,15 +306,32 @@ test_files_setup_state <- function(
   withr::defer(source_test_teardown(".", env), frame) # old school
 }
 
-test_files_reporter <- function(reporter, .env = parent.frame()) {
+test_files_reporter <- function(
+  reporter,
+  mode = c("serial", "parallel"),
+  .env = parent.frame()
+) {
+  # User selected reporter
```
gaborcsardi, line 314: Maybe put in a 
```r
mode <- match.arg(mode)
```
here, because otherwise the default is not very useful?

gaborcsardi (approved):
All good, but see one comment below.

Issue #2112, opened 2025-07-25
Make it easier to simulate running tests on CRAN


hadley:
* `on_cran()` needs to prioritise env var over `interactive()`, if its set. 
* Provide some helper to set the env var correctly so you can write `set_on_cran(TRUE)` rather than `Sys.setenv(NOT_CRAN = "false")`

This is particularly important for snapshot tests. If you accidentally put clean up code inside a snapshot, you can get to CRAN test failures that are hard to debug.

PR #2200, opened 2025-08-09
Don't override user-specified NOT_CRAN env var

hadley:
Continutation of #2112

### DIFF
--- a/R/local.R
+++ b/R/local.R
@@ -190,7 +190,7 @@ local_test_directory <- function(path, package = NULL, .env = parent.frame()) {
 }
 
 local_interactive_reporter <- function(.env = parent.frame()) {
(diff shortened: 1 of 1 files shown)


R/local.R
```diff
@@ -190,8 +190,7 @@ local_test_directory <- function(path, package = NULL, .env = parent.frame()) {
 }
 
 local_interactive_reporter <- function(.env = parent.frame()) {
-  # Definitely not on CRAN
-  withr::local_envvar(NOT_CRAN = "true", .local_envir = .env)
+  local_assume_not_on_cran()
```
gaborcsardi, line 193: Lost the `.env` argument here?

gaborcsardi (approved):
Makes sense!

PR #2077, opened 2025-04-07
Improve desc_filter so you can run single BDD-style tests (aka run just a single `it` test from `describe`)

kubajal:
Currently, `desc_filter` let's you run either a single `test_that` or a single `describe` suite of tests. This PR adds support for selecting just a single `it(...)` test from a `describe` suite by adding the `&&&` separator to the `desc` argument.
So e.g. running `testthat::test_file("tests.R", desc="level 0&&&level 1 A")`
```
# tests.R
describe("level 0", {
  it("level 1 A", {}),
  it("level 1 B", {})
})
```
effectively triggers just
```
describe("level 0", {
  it("level 1 A", {}),
})
```

This PR is a continuation of https://github.com/r-lib/testthat/pull/1904 and https://github.com/r-lib/testthat/pull/1840.

PR #2203, opened 2025-08-09
More tests of running a nested subtest

jennybc:
Consider this late-arriving review comments on #2077

I'll make a few comments inline.


hadley:
Thanks!

Issue #2184, opened 2025-08-07
Feature request: test discovery


kubajal:
It would be great if there was some API for test discovery. This would enable easier IDE integrations and improve overall user experience. The current alternative is to parse test content and extract `test_that` / `describe` / `it` calls ( e.g. in https://github.com/meakbiyik/vscode-r-test-adapter/blob/master/src/testthat/parser.ts#L84).

Writing parsers of test files is painful and raises issues such as https://github.com/meakbiyik/vscode-r-test-adapter/issues/32, ie. edge cases when the tests are generated using paste0 or a _for_ loop (see ).

I propose to expose a function such as `testthat::discover_tests(path)` that would return or print the list of tests in a CSV format.

PR #2220, opened 2025-08-15
Feature: expose srcref of tests in reporters

kubajal:
Currently, there is no way to report the line number of tests in reporters. This PR enables that.

This PR follows the logic of e.g. [LocationReporter](https://github.com/r-lib/testthat/blob/main/R/reporter-location.R) which prints the srcref of all `expect_*` assertions. It extends this behavior by also reporting locations of `test_that`, `describe` and `it` expressions.

With minimal effort, this PR enables https://github.com/r-lib/testthat/issues/2184.


hadley:
I've thought it over a bit more, and it doesn't feel like a good fit for testthat to me. I've done most of the work to make it maintainable for me over the long term, but it still feels a bit fragile and I'm not fully sold on the motivation.

PR #2221, opened 2025-08-17
Replaced filter_desc with dynamic label checking

kubajal:
This PR:
 1. replaces `filter_desc` functionality with checking matching test labels in the runtime by comparing the selectors with current stack maintained by `local_description_push`
 2. Also, instead of checking for string-string equality, it checks for string-regex matching; this gives more flexibility in selecting subsets of tests

Fixes https://github.com/r-lib/testthat/issues/2222

To do: increase test coverage and fix snapshot failures (I don't know how to do that).


hadley:
As I mentioned in the issue, I don't think this is a good idea, sorry.

PR #2231, opened 2025-09-08
Incorporate doc feedback from @maelle

hadley:


vignettes/challenging-tests.Rmd
```diff
@@ -134,6 +134,25 @@ test_that("user must respond y or n", {
 })
 ```
 
+If you don't care about reproducing the output of `continue()` and just want to recreate its return value, you can use `mock_output_sequence()`. This creates a function that each time it's called returns the next input supplied to `mock_output_sequence()`. The following code shows how it works and how you might use it to test `readline()`:
```
maelle, line 137: I find the second sentence unclear.
```suggestion
If you don't care about reproducing the output of `continue()` and just want to recreate its return value, you can use `mock_output_sequence()`. This creates a function that returns the input supplied to `mock_output_sequence()` in sequence: the first input at the first call, the second input at the second call, etc. The following code shows how it works and how you might use it to test `readline()`:
```

vignettes/custom-expectation.Rmd
```diff
@@ -21,6 +21,35 @@ This vignette shows you how to write your own expectations. Custom expectations
 
 In this vignette, you'll learn about the three-part structure of expectations, how to test your custom expectations, see a few examples, and, if you're writing a lot of expectations, learn how to reduce repeated code.
 
+## Do you need it?
+
+But before you read the rest of the vignette and dive into the full details of creating a fully 100% correct expectation, consider if you can get away with a simpler wrapper. For example, take this expectation from tidytext:
```
maelle, line 26: why not start with an example of a wrapper that is fine? For instance it could be 

```r
expect_df <- function(tbl) {
  expect_s3_class(tbl, "data.frame")
}
```

Or here is a real example: https://github.com/ropensci/aorsf/blob/2b7fa72bff5cd4e8b2984358e35d0938e50540b0/tests/testthat/helper-orsf.R#L259

maelle (commented):
Thank you!

vignettes/custom-expectation.Rmd
hadley: Good idea, thanks!

PR #2233, opened 2025-09-09
Fix the lister when running in parallel

gaborcsardi:
It was not prepared for receiving updates from multiple files concurrently and some results were lost.

### DIFF
--- a/R/reporter-list.R
+++ b/R/reporter-list.R
@@ -11,50 +11,52 @@ ListReporter <- R6::R6Class(
   "ListReporter",
   inherit = Reporter,
   public = list(
-    current_start_time = NA,
-    current_expectations = NULL,
-    current_file = NULL,
-    current_context = NULL,
-    current_test = NULL,
+    running = NULL,
+    current_file = "", # so we can still subset with this
     results = NULL,
 
     initialize = function() {
       super$initialize()
       self$capabilities$parallel_support <- TRUE
+      self$capabilities$parallel_updates <- TRUE
       self$results <- Stack$new()
+      self$running <- new.env(parent = emptyenv())
     },
 
     start_test = function(context, test) {
+      # is this a new test block?
       if (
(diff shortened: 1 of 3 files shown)


R/reporter-list.R
```diff
@@ -11,50 +11,52 @@ ListReporter <- R6::R6Class(
...
 
     initialize = function() {
       super$initialize()
       self$capabilities$parallel_support <- TRUE
+      self$capabilities$parallel_updates <- TRUE
       self$results <- Stack$new()
+      self$running <- new.env(parent = emptyenv())
     },
 
     start_test = function(context, test) {
```
hadley, line 26: I don't know if it's important, but to fully handle nested tests (which always existed via `describe()` and `it()` but are now fleshed out for `test_that()` too), `start_test()` might be called multiple times before `end_test(); i.e. tests are a stack now.

gaborcsardi: Yes, that is actually how the parallel reporting works. We call `start_file()` and `start_test()` every time we get a result from a subprocess. 

But do you also call `end_test()` multiple times, or no?

hadley: Right — each test always generates one `start_test()` and one `end_test()` but there might be multiple starts before you get to the match ends. There's a `tests/testthat/reporter/nested.R` you can use to test if needed.

hadley: Yeah, seems fine. I mostly wanted to make sure that you were aware of this change to the reporter API because I keep forgetting about it.

hadley (approved):

R/reporter-list.R
gaborcsardi: TBH, I never really looked at how `describe()` tests are different, so I indeed wasn't aware of this. 

But the logic should be the same as before for non-parallel runs, and parallel runs should now produce the same results as non-parallel runs. This was my reasoning for this PR being correct.

PR #2238, opened 2025-09-17
Allow to set a tolerance for set_state_inspector()

### LINKED ISSUE #2237: add a tolerance argument to set_state_inspector()
It would be nice to be able to control the tolerance when using `set_state_inspector()` to avoid diffs like this:
```R
Global state has changed:
`before$par$mai`: 1.01999999999999980 0.81999999999999984 0.81999999999999973 0.41999999999999998
 `after$par$mai`: 1.02000000000000002 0.81999999999999984 0.82000000000000006 0.41999999999999998
```

mcol:
I'm not sure how to write a test for this, but locally it works as desired.

Fixes #2237.


hadley:
Thank you!

Issue #1851, opened 2023-08-31
expect_not_contains


aronatkins:
It would be helpful to have a collection of negated checkers, or utilities to help negate existing checks. For example, we have `expect_contains`, but not a straightforward way to check when values should _no_ be in the haystack.

```r
expect_not_contains(c("a","b"), c("a","c"))
Error: c("a", "b") (`actual`) contains some of the values in c("a", "c") (`unexpected`).
* Missing from `actual`: "c"
* Present in `actual`:   "a", "b"
```

I could write this as the following, but would not receive any assistance in reporting the unexpected values:

```r
expect_false(any(c("a","c") %in% c("a","b")))
```

Getting feedback is cumbersome:

```r
haystack <- c("a","b")
values <- c("a","c")
testthat::expect_false(any(values %in% haystack), paste("unexpected: ", paste0(values[values %in% haystack], collapse = ", ")))
```

aronatkins:
The "utilities to help negate checks" could be really powerful, but it's a departure from the current approach. I didn't really detail that idea earlier, so here's what I was trying to suggest:

Consider a generic `expect_that` helper that allowed composable checks.

```r
expect_that(equals, A, B)
expect_that(not(equals), A, B)
expect_that(contains, A, B)
expect_that(not(contains), A, B)
```

I'm hand-waving the implementation details and how the "checks" are defined, but something like this would reduce the number of unique `expect_not_some_specific_thing` functions and build a grammar of composable test operators.

The operators don't all need to take two arguments.

```r
expect_that(file_exists, A)
expect_that(not(file_exists), B)
expect_that(nearly_equal, A, B, 0.05)
expect_that(not(nearly_equal), A, B, 0.05)
```

PR #2239, opened 2025-09-19
`expect_not_contains()` and `expect_not_in()`

stibu81:
This introduces two negated expectations as suggested in #1851 with the following functionality:

* `expect_not_contains(x, y)` tests that `x` contains none of the elements of `y` (i.e. `y` is disjoint from `x`).
* `expect_not_in(x, y)` tests that no element of `x` is in `y` (i.e. `x` is disjoint from `y`).

While the not negated expectations actually do something different, these two are actually equivalent. It might still make sense to have them both.

During implementation I realised that one might have different expectations from these names. For example, one might expect that `expect_not_in(x, y)` checks that:
* none of the elements of `x` are in `y` (which is what I implemented)
* `x` is not a subset of `y`
Both of them could also meaningfully be understood as inversions of the other two expectations. Would the second variant also be of interest?


lionel-:
How about a single `expect_dijoint()` function?

PR #2240, opened 2025-09-19
Add `required` argument to expect_true

### LINKED ISSUE #1836: FR: all= argument for expect_true()/expect_false()
`expect_true(all(.))` is a very common construction. Here's >10,000 hits on GitHub:

https://github.com/search?q=%22expect_true%28all%28%22&type=code

I propose adding `all=` as an argument to `expect_true()`. The advantage is that `testthat` can offer a better test failure interface than is currently available.

Compare:

```r
test <- c(rep(TRUE, 10), FALSE)
expect_true(all(test))
# Error: all(test) is not TRUE

# `actual`:   FALSE
# `expected`: TRUE 
```

Not very helpful (NB: this is the same in 2e and 3e as of now). Proposed interface:

```r
expect_true(test, all = TRUE)
# Error: test are not all TRUE

# `actual` is FALSE at index 11
```

I don't see this as often for `expect_false(all(.))`, but for consistency we could add the argument there. We could also add an `any=` argument to handle `expect_true(any(.))`, [which is not uncommon](https://github.com/search?q=%22expect_true%28any%28%22&type=code), though maybe it's better served with something similar to `expect_match`: `all = TRUE` --> require `all()`, `all = FALSE` --> require `any()`, `all = NULL` --> current behavior, i.e. require `identical(., TRUE)`.

snystrom:
Adds `required` argument to `expect_true()` that accepts `c("one", "any", "all")`. I've held off implementing for `expect_false` until we make sure the implementation is sound for easier review. This also means the docstring is slightly out of sync between `expect_true` and `expect_false` for now (and indeed the language could be a little nicer--I expect to polish this up too once we're aligned on the approach).

When `required = "any"` we throw an error to avoid edges with numeric vectors that contain `1` which may be coerced to `TRUE`.

Fixes #1836.

### DIFF
--- a/R/expect-constant.R
+++ b/R/expect-constant.R
@@ -7,6 +7,10 @@
 #'
 #' Attributes are ignored.
 #'
+#' @param required "one" (default) to require an identical match to `TRUE`.
+#'   "any" to replicate `expect_true(any(TRUE))`. "all" for
+#'   `expect_true(all(TRUE))`.
+#'
 #' @inheritParams expect_that
 #' @family expectations
 #' @examples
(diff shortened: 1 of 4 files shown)


hadley:
Thanks for working on this! I'll take it from here 😄

PR #2258, opened 2025-10-06
Make labels more flexible

hadley:
@DavisVaughan could you please have a look at the specific labels, but also contemplate what you think of this design.

### DIFF
--- a/R/expect-setequal.R
+++ b/R/expect-setequal.R
@@ -108,7 +108,7 @@ expect_mapequal <- function(object, expected) {
       if (edition_get() >= 3) {
         act <- labelled_value(act$val[exp_nms], act$lab)
 
-        msg <- "Expected %s to contain same values as %s."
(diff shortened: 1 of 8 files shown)


R/expect-setequal.R
```diff
@@ -103,13 +100,16 @@ expect_mapequal <- function(object, expected) {
...
+      msg <- "Expected %s to have the same names as %s."
+      act_names <- labelled_value(names(act$val), act$lab)
+      exp_names <- labelled_value(names(exp$val), exp$lab)
+      expect_setequal_(msg, act_names, exp_names)
     } else {
       if (edition_get() >= 3) {
         act <- labelled_value(act$val[exp_nms], act$lab)
-        expect_waldo_equal_("equal", act, exp)
+
+        msg <- "Expected %s to contain same values as %s."
```
DavisVaughan, line 111:
```suggestion
        msg <- "Expected %s to contain the same values as %s."
```

PR #2259, opened 2025-10-06
Implement `expect_all_equal()` and friends

### LINKED ISSUE #1836: FR: all= argument for expect_true()/expect_false()

### LINKED ISSUE #2235: Suggest a new expectation around "Vector with a constant value"
Inspiration: https://github.com/r-lib/lintr/issues/2927

I recently found a test that was silently passing with the format:

```r
expect_true(all(x == y))
```

It should have failed because the "actual" value, `x`, wound up being `NULL` (think `x <- df$COL` where a refactor removed the `COL` vector from the data.frame).

With that in mind I think this style of test should be avoided at all costs (and there's nothing `expect_true()` should do because it's `all()` causing the issue).

That said, my immediate thought was to turn to `expect_identical(x, y)`, with `tolerance=` added as needed (or `expect_equal`, pick your favorite).

But a very common style for this test is:

```r
expect_true(all(x == "FOO"))
```

I.e., testing that a vector has a single value, `"FOO"`.

I don't think there's a neat way to express that in {testthat} today. There are two stricter equivalents I've hit on, neither of which feel particularly satisfying:

```r
# ugly, and all the uglier if 'x' is an expression!
expect_identical(x, rep("FOO", length(x))

# somewhat "tricky", and requires coercion for non-character 'x'
expect_named(table(x), "FOO")
## c.f.
expect_named(table(int_vector), "1")
## even worse
expect_named(table(dbl_vector), "3.14159265358979")

## still not fully satisfying
expect_identical(unique(x), "FOO")
expect_identical(unique(int_vector), 1L)
expect_identical(unique(dbl_vector), pi, tolerance=...)
```

Is this a good enough use case to generate its own expectation? Possibly #1836 could be related here -- if we remove the `all()` call, {testthat} itself could catch the silent error case under the hood and offer a useful failure:

```r
expect_true(x == "FOO", all = TRUE)
# internally, testthat finds 'x' to be empty and fails
```

hadley:
Fixes #1836. Fixes #2235. Closes #2240.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # testthat (development version)
 
+* `expect_all_equal()`, `expect_all_true()`, and `expect_all_false()` are a new family of expectations that checks that every element of a vector has the same value. Compared to using `expect_true(all(...))` they give better failure messages (#1836, #2235).
 * Expectations now consistently return the value of the first argument, regardless of whether the expectation succeeds or fails. The primary exception are `expect_message()` and friends which will return the condition. This shouldn't affect existing tests, but will make failures clearer when you chain together multiple expectations (#2246).
 * `set_state_inspector()` gains `tolerance` argument and ignores minor FP differences by default (@mcol, #2237).
 * `expect_vector()` fails, instead of erroring, if `object` is not a vector (@plietar, #2224).


R/expect-all.R
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  exp$val <- rep(exp$val, length(act$val))
+  names(exp$val) <- names(act$val)
+  expect_waldo_equal_(
+    "Expected every element of %s to equal %s.",
+    act,
+    exp,
+    trace_env = trace_env
+  )
```
DavisVaughan, line 66: Need `tolerance`?

R/expect-all.R
```diff
@@ -0,0 +1,67 @@
+#' Do you expect every value in a vector to have this value?
```
DavisVaughan, line 1:
```suggestion
#' Does a vector have a constant value?
```

Just thinking out loud

R/expect-all.R
```diff
@@ -0,0 +1,67 @@
...
+#' @export
+#' @examples
+#' x1 <- c(1, 1, 1, 1, 1, 1)
+#' expect_all_equal(x1, 1)
+#'
+#' x2 <- c(1, 1, 1, 2, 1, 1, 1, 1, 1, 1, 1, 2)
+#' show_failure(expect_all_equal(x2, 1))
+#'
+#' # expect_all_true() and expect_all_false() are helpers for common cases
+#' show_failure(expect_all_true(rpois(100, 10) < 20))
```
DavisVaughan, line 17: Maybe set a seed otherwise the pkgdown site changes constantly

tests/testthat/_snaps/expect-all.md
```diff
@@ -0,0 +1,56 @@
...
+      Error in `expect_all_equal()`:
+      ! `expected` must be length 1.
+
+# can compare atomic vectors
+
+    Code
+      expect_all_equal(x, TRUE)
+    Condition
+      Error:
+      ! Expected every element of `x` to equal TRUE.
```
DavisVaughan, line 30:
```suggestion
      ! Expected every element of `x` to equal `TRUE`.
```

??

hadley: FWIW that would be inconsistent with `expect_equal()`.

PR #2269, opened 2025-10-19
make `expect_lt()` etc. work properly for non-numeric data

### LINKED ISSUE #2268: `expect_lt()` and friends crash on failing tests with non-numeric arguments
`expect_lt()` and friends are implemented such that they accept any object that implements the required comparison operators (see d43ed65). And indeed this works, all of these remain silent:

```
expect_gt(9, 7)
expect_gt(as.Date("2025-01-09"), as.Date("2025-01-07"))
expect_gt(as.POSIXct("2025-01-09 00:00:00"), as.POSIXct("2025-01-07 00:00:00"))
expect_gt("9", "7")
```

However, when the tests fail, these expectations crash because during the preparation of the message the inputs are processed with functions that are not defined for some types (`log10()`, `-`):

```
expect_gt(9, 7)
expect_gt(as.Date("2025-01-07"), as.Date("2025-01-09"))
## Error in Math.Date(min(x)) : log10 not defined for "Date" objects

expect_gt(as.POSIXct("2025-01-07 00:00:00"), as.POSIXct("2025-01-09 00:00:00"))
## Error in Math.POSIXt(min(x)) : 'log10' not defined for "POSIXt" objects

expect_gt("7", "9")
## Error in act$val - exp$val : non-numeric argument to binary operator
```

This is not strictly a bug since the documentation reads "Do you expect a *number* bigger or smaller than this?" (emphasis mine), but at least the behaviour is inconsistent. As long as one only writes successful tests, one does not notice that the functions do not work reliably for some types.

The reason that this was not caught in the tests is that the existing tests are only for successful expectations.

Going back to strictly allowing only numeric values would likely break many existing tests, so I would to adapt the code that creates the messages to be more robust. I think I know how to do this and would create a PR if there are no objections.

stibu81:
This is an attempt to make `expect_lt()`, `expect_lte()`, `expect_gt()`, and `expect_gte()` work properly for non-numeric data by ensuring that the failure messages are created correctly also for those. It fixes #2268.

With these changes, the examples mentioned in #2268 lead to the following output:

```
expect_gt(7, 9)
## Error: Expected 7 > 9.
## Actual comparison: 7.0 <= 9.0
## Difference: -2.0 <= 0

expect_gt(as.Date("2025-01-07"), as.Date("2025-01-09"))
## Error: Expected `as.Date("2025-01-07")` > `as.Date("2025-01-09")`.
## Actual comparison: "2025-01-07" <= "2025-01-09"

expect_gt(as.POSIXct("2025-01-07 01:00:00"), as.POSIXct("2025-01-09 01:00:00"))
## Error: Expected `as.POSIXct("2025-01-07 01:00:00")` > `as.POSIXct("2025-01-09 01:00:00")`.
## Actual comparison: "2025-01-07 01:00:00" <= "2025-01-09 01:00:00"

expect_gt("7", "9")
## Error: Expected "7" > "9".
## Actual comparison: "7" <= "9"
```

I decided to differentiate between numeric inputs and all others. For numeric inputs, nothing changes, but for anything else, the messages are created differently. I made two decisions that others might disagree with and that I could change if requested:

* I added quotation marks around the values in the line "Actual comparison" to make sure that the output never looks like numbers.
* I don't output the line "Difference". This line could be output for dates and times, but it would need a little more preparation to make the output meaningful. I'm not sure that this is helpful, but of course I could make the change if you think it makes sense. I would format te output for dates and times similarly to as it is done in the `print`-method for `difftime`.

### DIFF
--- a/tests/testthat/test-expect-comparison.R
+++ b/tests/testthat/test-expect-comparison.R
@@ -84,6 +84,14 @@ test_that("comparisons with Date objects work", {
   expect_snapshot_failure(expect_gt(date, date2))
 })
 
+test_that("comparisons of date/time with NA work", {
+  time <- as.POSIXct("2020-01-01 01:00:00")
+  date <- as.Date("2020-01-01")
+
+  expect_failure(expect_lt(time, NA))
+  expect_failure(expect_gt(date, NA))
+})
+
 test_that("comparisons with character objects work", {
   expect_success(expect_lte("a", "b"))
 


hadley:
Thanks for working on this! Much appreciated 😄

PR #2282, opened 2025-11-06
Add OpenTelemetry instrumentation

shikokuchuo:
This PR adds a basic layer of OpenTelemetry instrumentation to testthat.

Sample traces from testthat's own tests:
Screenshot 2025-11-06 at 13 35 44

To do:

- [x] Record test results as span attributes and set error status if failing

Sample span attributes:
[
Screenshot 2025-11-28 at 10 28 10
](url)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # testthat (development version)
 
+* testthat now emits OpenTelemetry traces for tests when tracing is enabled. Requires the otel and otelsdk packages (#2282).
+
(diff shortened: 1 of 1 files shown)


hadley (approved):
Thanks for this! 

Did you check it works with nested tests? (i.e. `test_that()` inside another `test_that()`, or more commonly an `it()` inside of `describe()`) (I assume it should work, but worth checking.)

PR #2298, opened 2025-11-24
Always report success/failure count

### LINKED ISSUE #2297: expect_failure() should report that the correct number of failures was indeed observed
```r
expect_failure({
  pass()
  fail('x')
})
# Error: Expected exactly one failure and no successes.
# Actually succeeded 1 times
```

It's not clear from the message itself that the failure side of the equation is balanced -- I was reading through logs and puzzling over why the test didn't fail as I was finding it to do interactively.

MichaelChirico:
Closes #2297 

Initial proposal

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # testthat 3.3.1
 
+* `expect_success()` and `expect_failure()` are more clear about what the expectation actually did (#2297).
 * The hint to use `snapshot_download_gh()` is now only emitted when running in a job named "R-CMD-check" (#2300).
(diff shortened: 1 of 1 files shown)


hadley:
Hmmm, the intent was to report exactly what the problem was, which your change loses. I wonder if we could do something with the bullets?

```
Expected 1 success and 0 failures
✅ Observed 1 success
❌ Observed 1 failure
```

See `expect_match()` for existing art in this direction.

MichaelChirico (commented):
Excellent! LGTM

PR #2303, opened 2025-12-31
Enhance OO mocking

### LINKED ISSUE #2302: `local_mocked_s4_method()` doesn't work for generics from other packages

hadley:
Fixes #2302

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # testthat 3.3.1
 
 * `local_mocked_s3_method()` and `local_mocked_s4_method()` can now mock methods that don't already exist, and can use `definition = NULL` to temporarily remove a method. `local_mocked_s4_method()` now also works when the generic is defined in another package (#2302).
+* `expect_snapshot()` now reports the original error class for base errors, rather than `rlang_error` (#2286).
 * `expect_success()` and `expect_failure()` are more clear about what the expectation actually did (#2297).
 * The hint to use `snapshot_download_gh()` is now only emitted when running in a job named "R-CMD-check" (#2300).
 * `expect_snapshot_file()` correctly reports file name if duplicated (@MichaelChirico, #2296).
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -185,11 +185,21 @@ snapshot_replay.condition <- function(
   }
 
   if (cnd_class) {
-    type <- paste0(type, " <", class(x)[[1]], ">")
+    type <- paste0(type, " <", error_class(x), ">")
   }
 
   c(snap_header(state, type), snapshot_lines(msg, transform))
 }
+
+error_class <- function(x) {
+  # If error was entraced from base R error, use original error class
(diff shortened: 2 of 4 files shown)


R/mock-oo.R
```diff
@@ -29,16 +30,43 @@ local_mocked_s3_method <- function(
 ) {
   check_string(generic)
   check_string(signature)
-  check_function(definition)
+  check_function(definition, allow_null = TRUE)
+
+  generic_def <- get(generic, envir = frame)
```
DavisVaughan: Unused. Remove?

R/mock-oo.R
```diff
@@ -51,20 +79,28 @@ local_mocked_s4_method <- function(
...
+    if (is.null(def)) {
+      methods::removeMethod(generic_def, signature, env)
+    } else {
+      suppressMessages(methods::setMethod(generic_def, signature, def, env))
+    }
+    old
+  }
+
+  old <- set_method(definition)
+  withr::defer(set_method(old), frame)
```
DavisVaughan: Was it purposeful that `set_method()` locally captures `generic_def` and `signature` from the surrounding env rather than making them arguments? That kind of local capture always feels like a code smell to me (or when I _need_ it I drop a comment calling it out). I would have rather seen `set_method(generic_def, signature, def)` at each call site, where you see the full set of info required to set the method.

At the very least, I had to think about it for a minute when I saw `generic_def` in there but didn't see it defined anywhere, and had to backtrack up the outer function to find it.

tests/testthat/test-mock-oo.R
```diff
@@ -16,10 +16,37 @@ test_that("validates its inputs", {
...
+    expect_length(x, 42)
+
+    # Now remove it
+    local_mocked_s3_method("length", "test_mock_class2", NULL)
+    expect_length(x, 0)
+  })
+
+  # Method should be removed after scope ends
+  expect_length(x, 0)
+})
```
DavisVaughan: What about a test where `length.test_mock_class2` _does_ already exists before entering the `local()`, then you temporarily remove it for the `local()` scope, and then after exiting the local scope you ensure that the `old` definition was restored?

R/mock-oo.R
```diff
@@ -29,16 +30,43 @@ local_mocked_s3_method <- function(
 ) {
   check_string(generic)
   check_string(signature)
-  check_function(definition)
+  check_function(definition, allow_null = TRUE)
+
+  generic_def <- get(generic, envir = frame)
```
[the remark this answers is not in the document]
hadley: I don't think it's worth it, as there's no simpler equivalent of `getGeneric()` for S3 generics and the existing error message is _ok_.

tests/testthat/test-mock-oo.R
hadley: Ah yeah, this test doesn't really focus on the removing. I'll improve.

PR #2305, opened 2026-01-06
Add an LLM focussed reporter, and use it when appropriate

hadley:
This is a little messy because I needed to refactor `default_reporter()`.

I skipped deprecation on `default_progress_reporter()` because I can't imagine why any one would call it, and indeed, there's [no usage on CRAN](https://github.com/search?q=org%3Acran%20default_parallel_reporter&type=code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,11 @@
 * `test_dir()` will no longer run tests in parallel if only a single file is being tested (#2305). 
 * `default_parallel_reporter()` is no longer exported; use `default_reporter(parallel = TRUE)` instead  (#2305).
 * New `LlmReporter()` designed for LLMs to read. It's currently used automatically inside Claude code, cursor, and gemini cli, and you can set `AGENT=1` to use with any coding agent (#2287).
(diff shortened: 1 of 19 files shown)


R/reporter.R
```diff
@@ -117,35 +117,54 @@ Reporter <- R6::R6Class(
...
-#' * [ProgressReporter] for interactive, non-parallel; override with
-#'   `testthat.default_reporter`
-#' * [ParallelProgressReporter] for interactive, parallel packages;
-#'   override with `testthat.default_parallel_reporter`
-#' * [CompactProgressReporter] for single-file interactive; override with
-#'   `testthat.default_compact_reporter`
-#' * [CheckReporter] for R CMD check; override with `testthat.default_check_reporter`
+#' @description
+#' `default_reporter()` returns the default reporter for [test_dir()].
+#' If the `CLAUDECODE` env var is set, it uses [LlmReporter]; otherwise if
```
DavisVaughan: Hmm, I actually think this is harder to mentally parse now

I like the original way where you get told about the 90% case first, `ProgressReporter`, then you get told about increasingly more niche things after that

I also like the bulleted list from before

PR #2318, opened 2026-01-22
Use better wording in expect_setequal()

### LINKED ISSUE #2310: `expect_setequal()` Absent/Needs wording
https://github.com/r-lib/testthat/blob/2a0c29b20cfb4dfdf67dfc8a2b36a7517a8d529c/R/expect-setequal.R#L53-L54

The language for errors in `expect_setequal()` seems a little off.  To me, both "Absent" and "Needs" convey the same thing.

Would it make more sense for "Needs" to be something like "Extra", to convey that there's an unexpected value?

``` r
library(testthat)
expect_setequal(object = 1:3, expected = 1:2)
#> Error: Expected `1:3` to have the same values as `1:2`.
#> Actual: 1, 2, 3
#> Expected: 1, 2
#> Needs: 3
expect_setequal(object = 1:2, expected = 1:3)
#> Error: Expected `1:2` to have the same values as `1:3`.
#> Actual: 1, 2
#> Expected: 1, 2, 3
#> Absent: 3
```

<sup>Created on 2026-01-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

mcol:
Fixes #2310.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # testthat (development version)
 
 * `run_cpp_tests()` no longer accidentally reports that a test has been skipped (#2315).
+* `expect_setequal()` uses better wording in the results (@mcol, #2310).
 
 # testthat 3.3.2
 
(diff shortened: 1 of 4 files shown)


hadley:
Thank you!

PR #2354, opened 2026-09-18
Fix snapshot cleanup deleting .new files for multi-dot filenames

### LINKED ISSUE #2326: Snapshot `.new` files deleted for filenames with multiple dots
## Summary

When using `expect_snapshot_file()` with snapshot filenames containing multiple dots (e.g., `tc12.2_image_annotate.html`), the generated `.new` files are incorrectly identified as "unused" and deleted during snapshot cleanup.

This prevents users from reviewing snapshot changes via `snapshot_review()` because the `.new` files no longer exist.

## Reproducible Example

```r
# Create package with testthat edition 3
usethis::create_package(file.path(tempdir(), "testpkg"))
usethis::use_testthat(3)

# Create test file
cat(file = testthat::test_path("test-snapshots.R"), '
test_that("multi_dot_filename", {

  path <- tempfile(fileext = ".html")
  writeLines("<html><body>New content</body></html>", path)
  expect_snapshot_file(path, "tc12.2_image_annotate.html")
})

test_that("single_dot_control", {
  path <- tempfile(fileext = ".html")
  writeLines("<html><body>New content</body></html>", path)
  expect_snapshot_file(path, "single_dot.html")
})
')

# Create existing snapshots with different content to trigger .new file creation
dir.create(testthat::test_path("_snaps/snapshots"), recursive = TRUE)
writeLines("<html><body>Old content</body></html>",
           testthat::test_path("_snaps/snapshots/tc12.2_image_annotate.html"))
writeLines("<html><body>Old content</body></html>",
           testthat::test_path("_snaps/snapshots/single_dot.html"))

# Run tests
devtools::test()
```

**Output:**
```
Deleting unused snapshots: 'snapshots/tc12.new.2_image_annotate.html'
```

**Result:**
- `single_dot.new.html` - **preserved** ✓
- `tc12.new.2_image_annotate.html` - **deleted** ✗

## Root Cause

There is an inconsistency between how `.new` filenames are **created** vs how they are **expected** during cleanup:

### File creation (`R/snapshot-file.R`, `split_path()`)

```r
split_path <- function(path) {
  ...
  ext_loc <- regexpr(".", name, fixed = TRUE)  # First dot
  ...
}
```

Uses `regexpr(".", ..., fixed = TRUE)` which finds the **first** dot.

### Cleanup expectation (`R/snapshot-cleanup.R`, `snapshot_expected()`)

```r
snap_files_seen_new <- paste0(
  tools::file_path_sans_ext(snap_files_seen),  # Last dot
  ".new.",
  tools::file_ext(snap_files_seen)             # Last dot
)
```

Uses `tools::file_path_sans_ext()` and `tools::file_ext()` which split on the **last** dot.

### For filename `tc12.2_image_annotate.html`:

| Function | Splits on | Result |
|----------|-----------|--------|
| `split_path()` → `new_name()` | First dot | Creates: `tc12.new.2_image_annotate.html` |
| `tools::file_path_sans_ext()` | Last dot | Expects: `tc12.2_image_annotate.new.html` |

The mismatch causes the actual `.new` file to be treated as orphaned and deleted.

## Suggested Fix

Update `snapshot_expected()` in `R/snapshot-cleanup.R` to use `split_path()` instead of `tools::file_path_sans_ext()`/`tools::file_ext()`:

```r
# Current (inconsistent):
snap_files_seen_new <- paste0(
  tools::file_path_sans_ext(snap_files_seen),
  ".new.",
  tools::file_ext(snap_files_seen)
)

# Fixed (consistent with new_name()):
pieces <- split_path(snap_files_seen)
snap_files_seen_new <- paste0(
  ifelse(pieces$dir == "", "", paste0(pieces$dir, "/")),
  pieces$name, ".new.", pieces$ext
)
```

## Environment

- testthat version: 3.2.3 (also reproduced on latest dev)
- R version: 4.4.0+

## Impact

This bug affects any package using `expect_snapshot_file()` with filenames containing multiple dots, which is common for versioned test cases (e.g., `test1.2_feature.svg`, `plot.v2.png`).

digibrander:
## Problem

`expect_snapshot_file()` with a snapshot name containing more than one dot (for example `tc12.2_image_annotate.html`) has its `.new` comparison file deleted by the next snapshot cleanup, so `snapshot_review()` has nothing to show.

## Root cause

`new_name()` in `R/snapshot-file.R` (used when the `.new` file is created) splits a filename via `split_path()`, which finds the *first* dot with `regexpr(".", name, fixed = TRUE)`. For `tc12.2_image_annotate.html` this produces `tc12.new.2_image_annotate.html`.

`snapshot_expected()` in `R/snapshot-cleanup.R` (used during cleanup to decide which files are still expected) instead built the `.new` name with `tools::file_path_sans_ext()` / `tools::file_ext()`, which split on the *last* dot. For the same filename that predicts `tc12.2_image_annotate.new.html`.

Since the predicted name never matches the file that was actually created, cleanup treats the real `.new` file as unused and deletes it.

## Fix

`snapshot_expected()` now calls `new_name()` directly instead of re-deriving the `.new` name with a different, inconsistent splitting convention. Both sides of the comparison are now guaranteed to agree because they're the same code.

## Testing

Added a regression test in `tests/testthat/test-snapshot-cleanup.R` using a multi-dot snapshot name. Verified it fails on the unmodified source (flags the real `.new` file as outdated) and passes with the fix. Ran the full existing test suite (`pkgload::load_all()` + `testthat::test_file()` per file, since the parallel `callr`-based runner doesn't start in this sandbox); 1041 pre-existing tests pass unmodified, 3 pre-existing failures reproduce identically with and without this change (they depend on running inside the package's own R CMD check/`devtools::test()` process context, e.g. `testing_package()` and `topenv()` checks, not on anything touched here).

Fixes #2326


jadhavgaurav:
Closing and reopening under the correct account (this was opened under the wrong GitHub identity due to a local auth mix-up on my end). The fix itself stands; resubmitting shortly as a fresh PR from the right account.

digibrander:
Closing to reopen under the correct GitHub account (this went out under a local auth mix-up). The fix itself stands and I'll resubmit shortly from the right account.

